PR #4, opened 2014-07-11
6.1 language

bwrsandman:
Port of https://code.launchpad.net/~savoirfairelinux-openerp/openerp-hr/language/+merge/194925

### DIFF
--- a/hr_language/__init__.py
+++ b/hr_language/__init__.py
@@ -1,0 +1,22 @@
+# -*- encoding: utf-8 -*-
+###############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2013 Savoir-faire Linux (<http://www.savoirfairelinux.com>).
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#    GNU Affero General Public License for more details.
+#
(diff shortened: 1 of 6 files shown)


lepistone:
Hi @joaoalf @yvaucher @pedrobaeza Are the points you raised on launchpad still open? If not, we can probably merge that. Thanks!

pedrobaeza:
Well, I still think we can find a way to not translate again all languages in this module, so I can't give my approval yet.

yvaucher:
I'm ok with this first version. My remark was answered. :+1:

ghost:
Is it really necessary to have 2 boolean fields for read and write. It seems redundant?

bwrsandman:
@dufresnedavid You can have someone who can read and not write.

Do you mean it would be better as a selection?

ghost:
I think it should be only one boolean named read/write and one boolean speak.

max3903:
On my resume, I could say I read spanish, but I do not write it.

yvaucher:
@dufresnedavid I disagree with you, reading is different than writing. As said @bwrsandman and @max3903 you can read a language and understand most of it but you couldn't write as you don't know enough rules of grammar or vocabulary. When reading you can make guess of meaning of some sentence or words in their context. You cannot guess when you are writing.

Moreover, if you take languages tests, you will be evalutated on listening, speaking, reading and writing skills. With a score on each skill.
http://en.wikipedia.org/wiki/International_English_Language_Testing_System#IELTS_test_structure

Still booleans seems to me a be too restrictive. However, as there are plenty of scales for languages, I wouldn't force to use one. In future we should replace it by a Selection list. Maybe with different modules for each type of scale.

yvaucher:
@dufresnedavid @max3903 @pedrobaeza @bwrsandman @joaoalf If one of you also aprove I merge.

About @dufresnedavid remark, correct me if I'm wrong, but we have more opinion in favor of multiple booleans. So I consider it's ok that way.

max3903:
:+1:

PR #6, opened 2014-07-11
Add meta files

bwrsandman:


rvalyi:
the build fails because no smtpclient python module is found, but I guess this is not the scope of this merge proposal right? So I would approve if you confirm. Thanks for that work.

bwrsandman:
@rvalyi actually it fails because of PEP8

pedrobaeza:
:+1:

PR #14, opened 2014-07-22
[7.0] Fix issue when no lastname is supplied

bwrsandman:


dreispt:
Why not this?

```
names = (vals.get('firstname'), vals.get('lastname'))
vals['name'] = " ".join(s for s in names if s)
```

bwrsandman:
Because the error it fixes is in cases where `vals['name']` is supplied and first and last aren't.
I want to avoid overwriting vals['name']
see: https://travis-ci.org/OCA/hr/builds/30551570#L2047

dreispt:
OK. :thumbsup:

pedrobaeza:
:+1:

PR #31, opened 2014-10-20
Update travis.

### LINKED ISSUE #11: [6.1] Fix `hr_interview` dependency on `smtpclient`
6.1 tests are failing because:

```
except_orm: (u'Error', u"You try to install module 'hr_interview' that depends
on module 'smtpclient'.\nBut the latter module is not available in your system.")
```

bwrsandman:
Closes #11


dreispt:
I believe there was a discussion on non OCA module dependencies, and the general opinion was that OCA modules should not depend on non-OCA modules.

The way to go should be to remove the dependency from the OCA module.

bwrsandman:
Removed the deps

pedrobaeza:
:+1:

yvaucher:
:+1:  errors not related to this one

PR #35, opened 2014-11-11
Added module hr_salary_rule_variable

ghost:
This module allows to separate from the salary rules, the data that changes over the years. Example, an employer must pay a contribution of 0.0495 x GROSS on every pay slip. The rate may change over the years, but the logic always stays the same (rate \* GROSS).


max3903:
:+1:

hr_salary_rule_variable/hr_salary_rule.py
```diff
@@ -0,0 +1,45 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import fields, orm
+
+
+class hr_salary_rule(orm.Model):
+    _name = 'hr.salary.rule'
```
pedrobaeza: Not needed

hr_salary_rule_variable/hr_payslip.py
```diff
@@ -0,0 +1,71 @@
...
+#
+##############################################################################
+
+from openerp.osv import orm
+from openerp.tools.safe_eval import safe_eval as eval
+from openerp.tools.translate import _
+
+
+class hr_payslip(orm.Model):
+    _name = 'hr.payslip'
```
pedrobaeza: Not needed

pedrobaeza:
Good contribution! What I don't see is where do you call `get_rule_variable` for applying variables on rules. Can you please explain me?

ghost:
In the python script of a salary rule, you may call it via the payslip browsable object:
payslip.get_rule_variable(rule_id, payslip.date_from)
rule_id always refer to the current rule.

pedrobaeza:
You should mention then how to use it on the moduel description.

The rest :+1:

pedrobaeza:
Thanks for the changes :+1:

PR #37, opened 2014-11-12
added module hr_worked_days_hourly_rate

ghost:
This module adds hourly rate and date fields on payslip worked days. Both fields are mandatory only when a number of hours is entered.


hr_worked_days_hourly_rate/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#
+##############################################################################
+
+{
+    'name': 'Worked Days Hourly Rates',
+    'version': '1.0',
+    'license': 'AGPL-3',
+    'category': 'Generic Modules/Human Resources',
+    'description': """
+Wroked Days Hourly Rate
```
max3903: Worked

max3903:
:+1:

hr_worked_days_hourly_rate/hr_payslip_worked_days.py
```diff
@@ -0,0 +1,41 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import fields, orm
+from datetime import datetime
+strptime = datetime.strptime
```
pedrobaeza: This is not needed.

hr_worked_days_hourly_rate/hr_payslip.py
```diff
@@ -0,0 +1,51 @@
...
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from openerp.osv import orm
+
+
+class hr_payslip(orm.Model):
+    _name = 'hr.payslip'
```
pedrobaeza: This is not needed.

hr_worked_days_hourly_rate/hr_payslip_worked_days.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class hr_payslip_worked_days(orm.Model):
+    _name = 'hr.payslip.worked_days'
```
pedrobaeza: This is not needed when you put _inherit and the result model is the same.

hr_worked_days_hourly_rate/hr_payslip_worked_days.py
```diff
@@ -0,0 +1,41 @@
...
+    _columns = {
+        'hourly_rate': fields.float('Hourly Rate'),
+
+        # When a worked day has a number of hours and an hourly rate,
+        # it is necessary to have a date, because hourly rates are likely to
+        # change over the time.
+        'date': fields.date('Date'),
+    }
+    _defaults = {
+        'hourly_rate': 0.0,
```
pedrobaeza: This should be 1.0 for normal computation.

hr_worked_days_hourly_rate/hr_payslip.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def sum_hourly_rates(
+        self, cr, uid, ids,
+        payslip,
+        context=None,
+    ):
+        """
+        This function is intended to be called by salary rules.
+
+        payslip: a Payslip Browsable Object which is different from a
```
pedrobaeza: Use standard PEP docstring `:param payslip:`

hr_worked_days_hourly_rate/hr_payslip.py
```diff
@@ -0,0 +1,51 @@
...
+        browse record
+
+        Returns the gross salary based on worked hours
```
pedrobaeza: Use standard dosctring `:return`. And I think this more as a sum of hours proportionally sized, that a gross salary.

hr_worked_days_hourly_rate/hr_payslip_worked_days.py
```diff
@@ -0,0 +1,41 @@
...
+class hr_payslip_worked_days(orm.Model):
+    _name = 'hr.payslip.worked_days'
+    _inherit = 'hr.payslip.worked_days'
+    _columns = {
+        'hourly_rate': fields.float('Hourly Rate'),
+
+        # When a worked day has a number of hours and an hourly rate,
+        # it is necessary to have a date, because hourly rates are likely to
+        # change over the time.
+        'date': fields.date('Date'),
```
pedrobaeza: I should put `date_start` and `date_stop` fields for covering the corresponding interval of the worked_days record.

hr_worked_days_hourly_rate/hr_payslip_worked_days.py
ghost: 1.0 would mean 1$/Euro. I prefer 0 because it is the amount used for unpaid leaves.

pedrobaeza: Ah, I interpreted that as a rate: this hour computes double that normal hour and so on. Field name doesn't help either... I think it's more logicall to set it as a rate than a direct price. You set only one price, and different rates (like discounts or increments based on percentage). There are a lot of companies that works this way: an extra hour is 1.6 price of a normal hour. If you don't decide to change it, please call it price_hour or similar. A help field also helps (as its name says :wink:)

Another thing: what about other currencies in this field. Are taken into account?

ghost: I don't agree on every aspect. Hourly rate is standard term in most anglo-saxon countries. What you mean is a factor by which to multiply the rate. Price per hour is not an accurate term and could lead to even more confusion, since you dont buy your employee's services. You pay a salary.

For the ratio, I will add a field to apply a percentage of the regular salary as you suggested.
I will also add a currency field and add help parameters.

ghost: Changed my mind in regard to the currency field. It should be added in another module if needed. This module will assume that the currency used is always the company's default currency.

ghost:
@eLBati You may remove the needs fixing label, my code passes the tests. Thanks!

hr_worked_days_hourly_rate/hr_payslip_worked_days.py
```diff
@@ -0,0 +1,77 @@
...
+        'hourly_rate': 0,
+        'rate': 100,
+        'date_from': lambda *a: datetime.strftime(
+            datetime.now(),
+            DEFAULT_SERVER_DATE_FORMAT
+        ),
+        'date_to': lambda *a: datetime.strftime(
+            datetime.now(),
+            DEFAULT_SERVER_DATE_FORMAT
+        ),
```
bwrsandman: This can be simplified to:

``` python
lambda *a: datetime.now().strftime(DEFAULT_SERVER_DATE_FORMAT)
```

hr_worked_days_hourly_rate/tests/test_hr_worked_days.py
```diff
@@ -0,0 +1,92 @@
...
+                'name': 'test',
+                'contract_id': self.contract_id,
+            },
+            context=self.context,
+        )
+
+        worked_days = self.worked_days_model.browse(
+            self.cr, self.uid, worked_days_id, context=self.context)
+
+        self.assertTrue(worked_days.total == 40 * 25 * 1.5)
```
bwrsandman: Prefer the use of `self.assertEqual(worked_days.total, 40 * 25 * 1.5)` which gives much more meaninful error messages

ghost: I prefer it as well!

bwrsandman:
:+1:

PR #38, opened 2014-11-12
added module hr_contract_multi_jobs

ghost:
This module removes the job_id field on contract view and adds a new field contract_job_ids. contract_job is a relationnal object between contracts and job positions.

A many2many field is not adequate because, it does not allow to specify information about each job for a contract. Example, an other module implements hourly rates. Many contracts could have the same job position but with different rates. In this case, we can add a field "hourly_rate" in the contract_job model.


dreispt:
Can you give the use case example for that need?

And shouldn't a "main" job_id be kept? I can see it important to get official statistics, such as legal reporting of a "profession code" for each contract. Localization modules depending on that would get wrong data.

ghost:
@dreispt I agree with you on the main "job_id". I will push a commit to add this feature.

For the use case example, I already put two in the module description. If you mean a use case for a real enterprise that Savoir-faire Linux has as a client, I do not have one yet. Though, many companies in Quebec have this issue. Employees are often paid with an hourly rate based salary, and some times, employees occupy 2 positions at the same time. This feature will help us sell the payroll package.

max3903:
:+1:

hr_contract_multi_jobs/hr_contract.py
```diff
@@ -0,0 +1,86 @@
...
+                    if contract_job.is_main_job
+                ]
+                if len(main_jobs) != 1:
+                    return False
+        return True
+
+    _constraints = [
+        (
+            _check_one_main_job,
+            """You must assign one and only one job position as main \
```
pedrobaeza: Remove `\` on text, and please put without triple quotas, because it introduces unwanted characters on strings. You can split line this way:

```
"You must assign one and only one job position as main "
"job position."
```

hr_contract_multi_jobs/hr_contract_job.py
```diff
@@ -0,0 +1,57 @@
...
+            required=True,
+            ondelete='cascade',
+        ),
+        'contract_id': fields.many2one(
+            'hr.contract',
+            'Contract',
+            required=True,
+            ondelete='cascade',
+        ),
+        'is_main_job': fields.boolean(
```
pedrobaeza, line 54: Don't you check that this is only marked for one job for each employee?

ghost, line 54: @pedrobaeza Yes, I check that with the constraint _check_one_main_job.

hr_contract_multi_jobs/hr_contract.py
ghost: @pedrobaeza Thanks, I changed the quotas.

dreispt:
:+1: LGTM

pedrobaeza:
Thanks for the changing. :+1:

PR #40, opened 2014-11-14
7.0 payroll employer contribution

ghost:
This module allows to compute employer contributions based on the wage bill of a company instead of every single payslip separated. It is similar to the payslip model, but when a salary rule calls payslip.sum(...), it sums over every payslips related to a company instead of every payslips related to an employee.

This module is required when a contribution from an employer is based on a rate that depends on the wage bill. For exemple, a rate based on the portion of the wages that exceeds 1 Million $/Euro.


max3903:
:+1:

PR #42, opened 2014-11-18
7.0 : Add module hr_expense_sequence

alexis-via:
This module puts a sequence in the 'name' field of hr.expense.expense and adds a 'description' field to replace it. That way, the expense number is used in the ref of the account.move.

### DIFF
--- a/hr_expense_sequence/hr_expense_sequence.py
+++ b/hr_expense_sequence/hr_expense_sequence.py
@@ -1,0 +1,61 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    HR Expense Sequence module for OpenERP
+#    Copyright (C) 2014 Akretion (http://www.akretion.com)
+#    @author Alexis de Lattre <alexis.delattre@akretion.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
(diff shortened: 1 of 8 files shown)


pedrobaeza:
:+1: 

Travis errors are unrelated, although you should rebased when #39 is merged.

hr_expense_sequence/hr_expense_sequence.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class hr_expense_expense(orm.Model):
+    _inherit = 'hr.expense.expense'
+    _order = 'name desc'
+
+    _columns = {
+        # Move the description from the 'name' field to 'description' field
+        # In the 'name' field, we now store the number/sequence
+        'name': fields.char('Number', size=32, readonly=True),
```
ghost, line 34: If you have a serious reason to put size=32, please specify it.

ghost:
You take the field name and make rename it description. This seems problematic to me. If someone wants to push a module that does not inherit yours and has a function that refers to the field 'name', it will instead refer to your field (which is different) on travis when testing. Travis installs all modules of the repository before testing.
So this is my point. You should keep
'name': fields.char('Description'...)
and add your own field
'my_new_field': fields.char('Number'...)

alexis-via:
@dufresnedavid I implemented it that way because it's the "name" field of the expense note that is used as reference in the account move, and the code that generate the account move is not easily inheritable. And I really want to have the number of the expense in the account move, and not it's description.

ghost:
You can make the field 'name' not required and invisible and then create a new field for the number. Then you stay compatible with other modules.

alexis-via:
@dufresnedavid Do you have an example of a module that would be "incompatible" with this module because it uses the "name" field in a particular way ? I would be very interested to have an example ; that way, we would be able to discuss with facts and not just theory.

With the module hr_expense_sequence, the "name" field is kept and it still a Char field ; we just put a sequence inside it, that's all.

ghost:
I guess you're right. This is not a big deal. Could you add a default value for the field description? Then :+1:

dreispt:
Two OKs, merging...

PR #43, opened 2014-11-18
8.0 : Add module hr_expense_sequence

alexis-via:
This PR is the same as the following PR for v7 https://github.com/OCA/hr/pull/42
Please comment on the PR for v7.

### DIFF
--- a/hr_expense_sequence/__openerp__.py
+++ b/hr_expense_sequence/__openerp__.py
@@ -1,0 +1,48 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    HR Expense Sequence module for OpenERP
+#    Copyright (C) 2014 Akretion (http://www.akretion.com)
+#    @author Alexis de Lattre <alexis.delattre@akretion.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Sequence on Expenses',
+    'version': '0.1',
+    'category': 'Human Resources',
+    'license': 'AGPL-3',
+    'summary': "Adds a sequence on expenses",
+    'description': """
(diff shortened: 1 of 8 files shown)


hr_expense_sequence/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+##############################################################################
+
+{
+    'name': 'Sequence on Expenses',
+    'version': '0.1',
+    'category': 'Human Resources',
+    'license': 'AGPL-3',
+    'summary': "Adds a sequence on expenses",
+    'description': """
+Sequence on Expenses
```
pedrobaeza, line 30: You can put this as a separate file README.rst

hr_expense_sequence/hr_expense_sequence.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class hr_expense_expense(orm.Model):
+    _inherit = 'hr.expense.expense'
+    _order = 'name desc'
+
+    _columns = {
+        # Move the description from the 'name' field to 'description' field
+        # In the 'name' field, we now store the number/sequence
+        'name': fields.char('Number', size=32, readonly=True),
```
pedrobaeza: Please use new API to only overwrite changed attribute (string).

pedrobaeza:
Two minor issues, but the rest LGTM.

alexis-via:
OK pedro, let's go for the new API on this small module, I know you like it :)

pedrobaeza:
Cool! :smile:

pedrobaeza:
:+1:

max3903:
Module has already been merged with #70

pedrobaeza:
Sorry for overlapping, but I even didn't remember this PR when I had again the need. Do you want credits in my module, @alexis-via?

alexis-via:
Arf, incredible !

@pedrobaeza Don't worry about that. It's your code, not mine, so you don't need to give me any credits.

alexis-via:
@pedrobaeza 
I had a look at your code in "your" hr_expense_sequence module : I see that you defined a new field "number" to store the sequence. In "my" hr_expense_sequence, I defined a new field "description" to store the description and I used the existing field "name" to store the sequence. Why ? Because the "name" field is copied on the account.move of the expense in the "ref" field, so you can see on the account.move to number of the related expense note, which is quite important for me.

But we could enhance "your" module to do that, via an inherit of account_move_get() (addons/hr_expense.py line 148). What you do think ? If you agree, I'll try to make a PR for that.

pedrobaeza:
@alexis-via, I always prefer not to change data model swapping fields content, specially if no proper installation/uninstallation scripts are providden. Even that, it's always trickier to do that. That's why I added a new column for the sequence.

I'm OK with adding that overriding for change the name of the account move line.

Thanks.

alexis-via:
@pedrobaeza Here is my PR to add the "number" of the expense on the "ref" field of the account.move (that only field I can inherit easily) : https://github.com/OCA/hr/pull/87

PR #44, opened 2014-11-18
7.0 Add module hr_expense_analytic_default

alexis-via:
Extract from the module description:

This module adds 2 fields:

1) Default Analytic Account on the employee form,

2) Default Analytic Account on the expense form.

If you set a Default Analytic Account on the employee, it will be copied
to the Default Analytic Account of the expenses of this employee. The
Default Analytic Account of the expense is used by default on each
expense line.

It is very useful when you use analytic per "sector" of the company (R&D, Sales, Marketing, etc...) ; in this case, it is easy to assign a default analytic account for each employee and it will be easier for this employee to have the analytic account of his sector set by default on his expense and expense lines.

### DIFF
--- a/hr_expense_analytic_default/__init__.py
+++ b/hr_expense_analytic_default/__init__.py
@@ -1,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    HR Expense Analytic Default module for OpenERP
+#    Copyright (C) 2014 Akretion (http://www.akretion.com)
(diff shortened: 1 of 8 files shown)


pedrobaeza:
Some of the comments of the 8.0 PR can be applied here. The rest LGTM.

mdietrichc2c:
LGTM, as in 8.0 :+1:

charbeljc:
:+1:

max3903:
:+1:

PR #46, opened 2014-11-18
7.0 : Add modules hr_expense_onchange and hr_expense_show_taxes

alexis-via:
The module hr_expense_onchange is just a technical module that adds on_change on 2 fields of expense
lines : unit_amount and unit_quantity. It is designed to be used as a dependency for other modules, including hr_expense_show_taxes and l10n_fr_hr_expense_private_car.

The module hr_expense_show_taxes will allow you to see the untaxed amount and tax amount on
each expense line. You will also see the total untaxed and total taxes of the expense.

### DIFF
--- a/hr_expense_onchange/__init__.py
+++ b/hr_expense_onchange/__init__.py
@@ -1,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    HR Expense Onchange module for OpenERP
+#    Copyright (C) 2014 Akretion (http://www.akretion.com)
+#    @author Alexis de Lattre <alexis.delattre@akretion.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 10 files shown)


eLBati:
Hello @alexis-via travis reports some `E265 block comment should start with '# '`

alexis-via:
After some more thoughts about this module, I think the implementation about the taxes in the expense lines is not the best approach. I think it would be better to have a true M2M field "taxes_ids" in the expense line, and use it to compute the amount of taxe. That would avoid the creation of several similar expense products with different level of taxes, for example:
- Meals (NO VAT)
- Meals (VAT 5.5%)
- Meals (VAT 10%)
- Meals (VAT 20%)

I seems that it's the direction that will be taken by future versions of the expense module in odoo, cf
https://docs.google.com/a/akretion.com.br/document/d/10XFQk7z088Z-MLEH3IaKZY28-eeb4tTqd_gzurkA-_Y/edit#heading=h.tuomw8y39650
https://pad.odoo.com/p/r.bd4JuNZUZclDd9oL

So it's better if we provide module in OCA for expense that move in the same direction.

feketemihai:
@alexis-via Can we close this PR since it's more than one year and a half since not touch?

alexis-via:
@feketemihai I don't care about v7, so you can trash this PR.

PR #47, opened 2014-11-18
Add modules hr_expense_onchange and hr_expense_show_taxes

alexis-via:
This PR is the equivalent for v8 of the following PR for v7: https://github.com/OCA/hr/pull/46
 Please comment on the PR for v7 (unless your remark is specific for v8)

### DIFF
--- a/hr_expense_onchange/__init__.py
+++ b/hr_expense_onchange/__init__.py
@@ -1,0 +1,23 @@
+# -*- encoding: utf-8 -*-
+##############################################################################
+#
+#    HR Expense Onchange module for OpenERP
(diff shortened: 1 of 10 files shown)


fclementic2c:
Alexis, I just did a test and I can see the untaxed amount ... the view is Ok.

However, in the terp you talk about constraints that do not work on my instance. I can use no supplier VAT or I can set a no 'taxe included' VAT on a product that I can expense... 

Did I miss anything ?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #50, opened 2014-11-27
hr_worked_days_activity

ghost:
.name_get(context=context) returns a list containning a tuple (id, name). What we want is to display only the name.

### DIFF
--- a/hr_worked_days_activity/hr_activity.py
+++ b/hr_worked_days_activity/hr_activity.py
@@ -46,12 +46,7 @@ class hr_activity(orm.Model):
                 res[activity.id] = activity.job_id.name_get()[0][1]
 
             elif activity.type == 'leave' and activity.leave_id:
-                # There is a missing context check in
-                # addons/hr_holidays/hr_holidays.py, so we need to pass the
-                # context here. This is fixed by a patch in v8.
-                res[activity.id] = activity.leave_id.name_get(
-                    context=context
-                )[0][1]
+                res[activity.id] = activity.leave_id.name_get()[0][1]
 
             else:
                 res[activity.id] = ''
--- a/hr_worked_days_activity/hr_activity.py
+++ b/hr_worked_days_activity/hr_activity.py
@@ -43,11 +43,12 @@ class hr_activity(orm.Model):
 
         for activity in self.browse(cr, uid, ids, context=context):
             if activity.type == 'job' and activity.job_id:
-                res[activity.id] = activity.job_id.name_get(
-                    context=context
-                )[0][1]
+                res[activity.id] = activity.job_id.name_get()[0][1]
 
             elif activity.type == 'leave' and activity.leave_id:
+                # There is a missing context check in
+                # addons/hr_holidays/hr_holidays.py, so we need to pass the
+                # context here. This is fixed by a patch in v8.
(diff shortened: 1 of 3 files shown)


pedrobaeza:
:+1:

hr_worked_days_activity/hr_activity.py
```diff
@@ -43,10 +43,14 @@ def _get_activity_name(
 
         for activity in self.browse(cr, uid, ids, context=context):
             if activity.type == 'job' and activity.job_id:
-                res[activity.id] = activity.job_id.name_get(context=context)
+                res[activity.id] = activity.job_id.name_get(
+                    context=context
```
joaoalf: You don't need to pass the context again.

hr_worked_days_activity/hr_activity.py
```diff
@@ -43,10 +43,14 @@ def _get_activity_name(
...
+                )[0][1]
 
             elif activity.type == 'leave' and activity.leave_id:
-                res[activity.id] = activity.leave_id.name_get(context=context)
+                res[activity.id] = activity.leave_id.name_get(
+                    context=context
```
joaoalf: You don't need to pass the context again.

joaoalf:
Hi @dufresnedavid 

Thanks for your contribution.

I've made some comments about your code.

ghost:
@joaoalf Thanks

hr_worked_days_activity/hr_activity.py
ghost: @joaoalf En enlevant le context en argument, j'obtien l'erreur suivante dans mon test:
  File "/home/ddufresne/odoo/7.0/git/odoo/addons/hr_holidays/hr_holidays.py", line 99, in name_get
    if not context.get('employee_id',False):
AttributeError: 'NoneType' object has no attribute 'get'
Ça ne cause pas ce problème pour activity.job_id.name_get mais pour être constant, je laisserais le context aux deux endroits.

yvaucher: @dufresnedavid Please add a comment there to explain it. This is an error in core addons.

I created a patch here https://github.com/odoo/odoo/pull/4341

ghost:
Thanks @yvaucher and @joaoalf. I removed the unecessary context in hr_job.name_get() and explained why it is required for hr_holidays_status.name_get().

joaoalf:
@dufresnedavid 
I noticed that in the _get_activity_name() method you don't have a context initialization. Probably if you put a context initialization at the begining of the _get_activity_name() method you may avoid the problem.    

Ex:

<pre>
context = context or {}
</pre>

<pre>
if context is None:
    context = {}
</pre>

ghost:
@joaoalf Of course, I will do it. Thanks

yvaucher:
@dufresnedavid you can remove those contexts https://github.com/odoo/odoo/pull/4341 has been merged

max3903:
@yvaucher merged in version 8, not 7...

yvaucher:
@max3903 uh right, my bad. So I guess it won't be changed any soon in 7.0 I'm doing 8.0 PR anyway. But I won't block this PR for this :+1: 

PR on 7.0: https://github.com/odoo/odoo/pull/4529

PR #51, opened 2014-11-28
Removed useless action on workflow node

ghost:
Removed a useless action that writes {'status': 'new'} when employee is created. The default value is enough but also, this node makes tests on other modules crash on travis when creating an employee.


pedrobaeza:
:+1:

veloutin:
:+1:

foutoucour:
:+1:

PR #53, opened 2014-12-07
Add module hr_contract_hourly_rate

ghost:
This module adds hourly rate classes on contract jobs. When computing the employee's payslip, this will allow to fetch the employee's hourly rate for the payslip duration. The method get_job_hourly_rate on contract model serves this purpose.


ghost:
Error is unrelated to this module. It is fixed in PR https://github.com/OCA/hr/pull/54.

hr_contract_hourly_rate/hr_hourly_rate_class.py
```diff
@@ -0,0 +1,65 @@
...
+        for hourly_rate_class in self.browse(cr, uid, ids, context):
+            rates = hourly_rate_class.line_ids
+            for r1 in rates:
+                for r2 in rates:
+                    if r1.id != r2.id:
+                        if r1.date_end:
+                            if r1.date_start <= r2.date_start <= r1.date_end:
+                                return False
+                        elif r1.date_start <= r2.date_start:
+                            return False
```
bwrsandman: Won't this always be triggered if there is `date_end is False` for `r1 and r2`?

hr_contract_hourly_rate/hr_contract.py
```diff
@@ -0,0 +1,109 @@
...
+                    ):
+                        return rate.rate
+                break
+        return False
+
+    def _check_has_hourly_rate_class(
+        self, cr, uid, ids, context=None,
+    ):
+        """
+        Checks if every contract job on the contract has an hourly rate
```
bwrsandman: Nitpicking, but `Check` would be better.
Imperative, present tense: "check", not "checked" or "checks".
Above you use "Get the hourly rate..." which is good. So for consistency sake.

hr_contract_hourly_rate/hr_contract.py
```diff
@@ -0,0 +1,109 @@
...
+class hr_contract(orm.Model):
+    _inherit = 'hr.contract'
+    _columns = {
+        'salary_computation_method': fields.selection(
+            [
+                ('wage', 'Annual Wage'),
+                ('hourly_rate', 'Hourly Rate')
+            ],
+            'Salary Computation Method',
+            help="Wheather to use the annual wage or an hourly rate "
```
bwrsandman: Wheather -> Whether

hr_contract_hourly_rate/hr_hourly_rate_class.py
ghost: @bwrsandman Yes it will, but that's the point. If the user needs two rates, he must put a end date to one of these. Example, between 2014-01-01 and 2014-12-31, you earn 25$/hour and from 2015-01-01 to infinite, you earn 30$/hour.

bwrsandman: So you can have at most one rate with no end date?

ghost: @bwrsandman Exactly!

hr_contract_hourly_rate/hr_contract.py
```diff
@@ -0,0 +1,109 @@
...
+                for contract_job in contract.contract_job_ids:
+                    if not contract_job.hourly_rate_class_id:
+                        return False
+
+            return True
+
+    _constraints = [(
+        _check_has_hourly_rate_class,
+        "Error! At least one job on contract has no hourly rate "
+        "class assigned.",
```
bwrsandman, line 107: Please make sure this string actually is translated in odoo, there was a bug and it might still be there which makes it so these constraints errors don't get translated.

There's a way around that bug, by replacing the string in the constaint declaration with a function reference and returning `_("Message")`

hr_contract_hourly_rate/hr_hourly_rate_class.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def _check_overlapping_rates(self, cr, uid, ids, context=None):
+        """
+        Checks if a class has two rates that overlap in time.
+        """
+        for hourly_rate_class in self.browse(cr, uid, ids, context):
+            rates = hourly_rate_class.line_ids
+            for r1 in rates:
+                for r2 in rates:
+                    if r1.id != r2.id:
```
bwrsandman: This double loop and if makes the code far too indented and pretty ugly.

Consider using [itertools.permutations](https://docs.python.org/2/library/itertools.html#itertools.permutations)

Something like:

``` python
from itertools import permutations

for r1, r2 in permutations(hourly_rate_class.line_ids, 2):
    if r1.date_end and (r1.date_start <= r2.date_start <= r1.date_end):
        return False
    elif not r1.date_end and (r1.date_start <= r2.date_start):
        return False
```

**Edit:** Added `2` as a parameter to have only size 2 permutations.

hr_contract_hourly_rate/hr_contract.py
ghost, line 107: @bwrsandman I just checked. It gets translated properly.

hr_contract_hourly_rate/hr_hourly_rate_class.py
ghost: @bwrsandman Nice, I agree its much better.

hr_contract_hourly_rate/hr_contract.py
bwrsandman, line 107: Good news!

bwrsandman:
:+1: 
Looks good to me!

foutoucour:
:+1:

max3903:
:+1:

PR #54, opened 2014-12-08
Fix programming errors in hr modules

ghost:
Fixed 2 programming errors in hr modules that make tests crash on other modules.

The first one is in hr_simplify. The file refers to the field 'state' on contract (hr.py line 159). But the field does not exist in this module or modules which it depends on.

The second error is in hr_job_category. It inherits the method write on contract and makes computations without validating the data passed to the method.

These modules obviously need unit tests.

### DIFF
--- a/hr_job_categories/hr.py
+++ b/hr_job_categories/hr.py
@@ -107,31 +107,29 @@ class hr_contract(orm.Model):
         #
         for contract in self.browse(cr, uid, ids, context=context):
             for data in prev_data:
-                if(
-                    'job_id' in vals
-                    and 'job_id' in data
-                    and data['job_id']
-                    and 'id' in data
-                    and data['id']
+                if (
+                        data.get('id') == contract.id
+                        and not vals.get('job_id', False)
+                        or (
+                            data.get('job_id')
+                            and data['job_id'][0] != vals.get('job_id')
+                        )
                 ):
-                    if (data['id'] == contract.id
-                            and not vals.get('job_id', False)
-                            or data['job_id'][0] != vals['job_id']):
-                        if data.get('job_id'):
-                            prev_job_id = data['job_id'][0]
-                        else:
-                            prev_job_id = False
-                        _l.warning(
-                            'prev Job, new job: %s, %s',
-                            prev_job_id, vals.get('job_id', False)
+                    if data.get('job_id'):
+                        prev_job_id = data['job_id'][0]
+                    else:
+                        prev_job_id = False
+                    _l.warning(
+                        'prev Job, new job: %s, %s',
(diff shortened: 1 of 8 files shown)


hr_job_categories/hr.py
```diff
@@ -107,23 +107,31 @@ def write(self, cr, uid, ids, vals, context=None):
...
-                    )
-                    self._remove_tags(
-                        cr, uid, contract.employee_id.id, prev_job_id,
-                        context=context
-                    )
-                    if vals.get('job_id'):
-                        self._tag_employees(
-                            cr, uid, contract.employee_id.id,
-                            contract.job_id.id, context=context
+                if(
```
pedrobaeza: Why don't you simplify it doing `if vals.get('job_id') and data.get('id'):`?

ghost: @pedrobaeza I agree. I pushed the change. Thanks!

bwrsandman:
> These modules obviously need unit tests.

Okay, could you add two unittests?
- [x] `write()` of `hr.contract` in `hr_job_categories` when the condition you edited is true
- [x] for `_no_of_contracts()` in  `hr_simplify` which refer to state of `hr.contract`

ghost:
@bwrsandman For now, I would prefer not to. I don't personally use nor inherit these modules but they interfere with my unittests in other modules in travis. If we begin to write unit tests in every untested modules from Michael Telahun Makonnen that generate bugs, I think it is going to take a lot of time.

bwrsandman:
As a general rule, if you fix a bug, it is good practice to add a unittest to:
1. prove that the code works now
2. prevent regressions

No one is going to take it upon themselves to add unittests to every module in hr, but we can add them where needed.

I am only asking for two untests. One for each programming bug you found. Seems reasonable.

ghost:
@bwrsandman ok, I'll put unittests there. Thanks

hr_simplify/tests/test_hr_simplify.py
```diff
@@ -0,0 +1,74 @@
...
+        and returns the number of employee
+        """
+        cr, uid, context = self.cr, self.uid, self.context
+
+        res = self.job_model._no_of_contracts(
+            cr, uid, [self.job_id],
+            name='no_of_employee', args=None, context=context
+        )
+
+        self.assertTrue(res[self.job_id]['no_of_employee'] == 1)
```
bwrsandman: You can do `self.assertEqual(res[self.job_id]['no_of_employee'], 1)`, you'll get a more informative error message when it goes wrong.

bwrsandman:
:+1: Very good.

ghost:
@bwrsandman Thanks!

max3903:
:+1:

PR #58, opened 2014-12-26
Payslip ytd amount

ghost:
This module adds a year-to-date field on payslip lines. This is used to display the year-to-date amount when generating the pdf payslip.


max3903:
:+1:

PR #59, opened 2015-01-02
german translation

rschnapka:
initial german translation, hr_language not fully done (yet)

### DIFF
--- a/hr_language/i18n/de.po
+++ b/hr_language/i18n/de.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: OpenERP Server 7.0\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2013-11-14 22:04+0000\n"
-"PO-Revision-Date: 2015-01-01 18:14+0100\n"
+"PO-Revision-Date: 2015-01-02 10:36+0100\n"
 "Last-Translator: Rudolf Schnapka <schnapkar@golive-saar.de>\n"
 "Language-Team: German <kde-i18n-de@kde.org>\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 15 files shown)


hr_language/i18n/de.po
```diff
@@ -0,0 +1,466 @@
...
+
+#. module: hr_language
+#: selection:hr.language,name:0
+msgid "Japanese / 日本語"
+msgstr ""
+
+#. module: hr_language
+#: selection:hr.language,name:0
+msgid "Russian / русский язык"
+msgstr ""
```
pedrobaeza: You cannot let this empty. At least put the same as in English.

pedrobaeza:
Please put also the updated pot files if they are missing.

rschnapka:
ok... thought an empty translation would autom. take untranslated term. But fine - will do :-)

rschnapka:
properly added full translation terms now...

pedrobaeza:
Can you please address the other comment?

rschnapka:
You probably mean the completely missing i18n folders? I'll look into creating those as needed - are they generally needed, even if no terms are translatable in the module?

pedrobaeza:
I mean the pot files - translation templates - (if any is missing).

Thanks for your contributions.

rschnapka:
each and every de.po file was derived from a pot file - so in fact there should be none missing. The only possible open issue in this context might have been missing translation (i18n-folder) of a module as such ... that's why i was asking...
In this context every module that had a *.pot file now also has a de.po file. 
AND If there is no pot-file then there is also no i18n folder that is missing a pot.

pedrobaeza:
Ah, OK. Then it's :+1: for me

yvaucher:
:+1:

PR #68, opened 2015-01-24
7.0 hr_contract_hourly_rate add constraint on wage

ghost:
Remove the required attribute on wage. Instead, add a constraint that checks the salary computation method is 'wage' and the wage is not filled.

### DIFF
--- a/hr_contract_hourly_rate/hr_contract.py
+++ b/hr_contract_hourly_rate/hr_contract.py
@@ -35,6 +35,10 @@ class hr_contract(orm.Model):
             "for computation of payslip.",
             required=True,
         ),
+        # Make the contract not required
+        'wage': fields.float(
+            'Wage', digits=(16, 2),
+            help="Basic Salary of the employee"),
     }
     _defaults = {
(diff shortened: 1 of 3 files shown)


hr_contract_hourly_rate/hr_contract.py
```diff
@@ -99,11 +103,35 @@ class assigned to it.
...
+    ):
+        """
+        Check if the wage on the contract is not False
+        if the salary computation method is wage
+        """
+        for contract in self.browse(cr, uid, ids, context=context):
+
+            # This does not apply when employee is paid by hourly rate
+            if contract.salary_computation_method == 'wage':
+                if contract.wage is False:
```
guewen: `if not contract.wage:` would be better IMO

guewen:
I added one remark, apart that LGTM
Thanks

hr_contract_hourly_rate/hr_contract.py
ghost: I agree, Thanks!

max3903:
:+1:

PR #81, opened 2015-03-09
[ADD] hr_expense_invoice

pedrobaeza:
# Set supplier invoices on HR expenses

This module allows to set a supplier invoice for each expense line, adding
the corresponding journal items to transfer the debt to the employee.
# Installation

Install the module the regular way.
# Configuration

You don't need to configure anything more to use this module.
# Usage

Instead of coding a full expense line, select an existing supplier invoice,
and then the rest of the fields will be auto-filled and grayed.

When you generate the expenses account entries, lines with invoices filled
will be generated as opposite of the payable move line of the invoice, and
both will be reconciled, letting the employee payable account as the only
open balance.
# Known issues / Roadmap
- Multiple payment terms for a supplier invoice are not handled correctly.
- Partial reconcile supplier invoices are also not correctly handled.


hr_expense_invoice/models/hr_expense_expense.py
```diff
@@ -0,0 +1,103 @@
...
+
+class HrExpenseExpense(models.Model):
+    _inherit = 'hr.expense.expense'
+
+    @api.model
+    def line_get_convert(self, x, part, date):
+        res = super(HrExpenseExpense, self).line_get_convert(x, part, date)
+        if x.get('invoice'):
+            res['partner_id'] = (
+                x['invoice'].partner_id.commercial_partner_id.id)
```
ghost, line 17: Can you rename your args to something more readable. I know that in hr_expense, the editor uses x and part as parameter, but we the editor's programming standards are quite low. I would suggest line_get_convert(self, line, partner, date).

hr_expense_invoice/models/hr_expense_expense.py
```diff
@@ -0,0 +1,103 @@
...
+                'account_id': line.invoice.account_id.id,
+                'name': line.name,
+                'price': line.total_amount,
+                'invoice': line.invoice,
+            }
+        else:
+            return super(HrExpenseExpense, self).move_line_get_item(line)
+
+    @api.multi
+    def action_move_create(self):
```
ghost, line 33: Can you add a unit test on this method?

hr_expense_invoice/models/hr_expense_expense.py
pedrobaeza, line 17: I'm afraid not, because sometimes I have faced that a positional argument is called from other overwritings with keyword arguments, and this renaming will break these calls, so we have to stick to these bad names. I found it renaming `partner` to `partner_id` argument on an onchange method (because semantically, it's an id, not the record itself).

hr_expense_invoice/models/hr_expense_expense.py
pedrobaeza, line 33: Sorry, but now I don't have more time to spend on this. Can you do it for me, please?

hr_expense_invoice/models/hr_expense_expense.py
ghost, line 17: I understand.

hr_expense_invoice/models/hr_expense_expense.py
ghost, line 33: Sure

ghost:
:+1:

max3903:
@pedrobaeza I am not sure to understand the scenario behind this module. It should be used when a supplier invoice is paid by an employee instead of the company. Am I correct ?

pedrobaeza:
Yes, that was exactly the reason to do it: you have to move the debt from supplier to employee. This module make it in an automatic way, allowing to double expense moves.

hr_expense_invoice/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Set supplier invoices on HR expenses
+====================================
+
+This module allows to set a supplier invoice for each expense line, adding
+the corresponding journal items to transfer the debt to the employee.
```
max3903: This module should be used when a supplier invoice is paid by an employee. It allows to set ...

:+1: if you add that to the description.

pedrobaeza:
@max3903, sorry for the delay.

I've changed the README according to your comments.

max3903:
:+1:

pedrobaeza:
Thanks

PR #89, opened 2015-04-07
8.0 - hr_language migration from 7.0

sebalix:
There was some errors related to fields `read` and `write` on `hr.language` conflicting with method names (as methods and fields share the same _namespace_ since v8), I rename them with a `can_*` prefix and set the `oldname` attribute on the field definitions.


pedrobaeza:
:+1: 

Thanks!

max3903:
:+1:

yvaucher:
:+1:

PR #91, opened 2015-04-20
[MIG] Migrate hr_contract_multi_jobs

feketemihai:


ghost:
@feketemihai Thanks very much for porting the 2 modules. I agree on removing yaml tests, but only to replace them with python unittests.

feketemihai:
This is what i am doing now...

hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,72 @@
...
+    _inherit = 'hr.contract'
+
+    @api.multi
+    @api.depends('contract_job_ids', 'contract_job_ids.contract_id')
+    def _get_main_job_position(self):
+        """
+        Get the main job position from the field contract_job_ids which
+        contains one and only one record with field is_main_job == True
+        """
+        for contract in self:
```
hbrunn: consider using `api.one` instead of looping yourself

hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,72 @@
...
+    # Modify the job_id field so that it points to the main job
+    job_id = fields.Many2one(
+        'hr.job',
+        string="Job Title",
+        compute="_get_main_job_position",
+        store=True)
+
+    @api.model
+    @api.depends('contract_job_ids', 'contract_job_ids.is_main_job')
+    @api.constrains('contract_job_ids')
```
hbrunn, line 52: I don't think `@depends` does something useful here, and then the constraint should be about `contract_job_ids.is_main_job`

hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,72 @@
...
+
+            # if the contract has no job assigned, a main job
+            # is not required. Otherwise, one main job assigned is
+            # required.
+            if contract.contract_job_ids:
+                main_jobs = [
+                    contract_job
+                    for contract_job in contract.contract_job_ids
+                    if contract_job.is_main_job
+                ]
```
hbrunn: you can do the same with
`contract.contract_job_ids.filtered('is_main_job')`

hr_contract_multi_jobs/models/hr_contract.py
gurneyalex, line 52: I think using `api.model` is not consistent with `for contract in self`. You want `api.multi` instead.

ghost:
:+1:

max3903:
:+1:

PR #92, opened 2015-04-20
[MIG] Migrate hr_contract_hourly_rate

feketemihai:
The PR depends on #91 since the module depends on the hr_contract_multi_jobs module, ported in the pr specified.


feketemihai:
I will rebase after #91 will be approved.

hr_contract_hourly_rate/models/hr_contract.py
```diff
@@ -19,35 +18,24 @@
...
-        contract_id,
-        job_id=False,
-        main_job=False,
-        context=None,
-    ):
+    salary_computation_method = fields.Selection(
+        [('yearly', 'Annual Wage'),
+         ('monthly', 'Annual Wage'),
+         ('hourly', 'Hourly Wage')],
+        string='Salary Computation Method',
```
mdietrichc2c, line 31: I'm very much against changing the possible values of the fields.selection from 7.0 to 8.0; this will cause a lot of migration errors.

feketemihai, line 31: I know but it's a selection fields, the options from 7.0 are still available, and from my point of view there should not be a problem...

max3903:
:+1:

hr_contract_hourly_rate/models/hr_contract.py
mdietrichc2c, line 31: Yes, but the conditions below, like `if contract.salary_computation_method != 'hourly':` , will fail on V7 items for every case. This means that without properly migrating `wage` to `yearly` or `hourly_rate` to `hourly`, the behaviour will be different before and after migrating for every case.

Is it possible to, at least, add a comment in the README in order to describe this change in the selection? That would be enough for me to know it's documented somewhere.

mdietrichc2c:
@feketemihai perfect, :+1:

charbeljc:
:+1:

PR #93, opened 2015-04-20
[ADD] hr_employee_data_from_work_address

hbrunn:
# User data from employee

When using HR, there are a couple of partners involved per employee: The employee's work address, the employee's private address and the partner assigned to the employee's user (if any). The latter is used to store some data about the user, like the email address. In many cases, it would be more convenient to edit this data from the employee record. This module assumes you assign a (different) partner for the work address to every employee, and synchronizes its values with the employee record. Then you should use this partner record also for the user record.
# Usage

After installation, updating an employee's `work_email` or `work_phone` record transparently changes the linked partner's appropriate field. For obvious reasons, the default for the work address being the company's address has been lifted. Partners created through the work address field will have the `employee`-flag set, and the `partner_id` field on the user record filters for this flag.

When creating a user, you can select the existing partner record for your employee, so it will be updated transparently too. Creating a user via the user field in the employee form preselects this partner record.


hr_employee_data_from_work_address/README.rst
```diff
@@ -0,0 +1,37 @@
+User data from employee
+=======================
+
+When using HR, there are a couple of partners involved per employee: The employee's work address, the employee's private address and the partner assigned to the employee's user (if any). The latter is used to store some data about the user, like the email address. In many cases, it would be more convenient to edit this data from the employee record. This module assumes you assign a (different) partner for the work address to every employee, and synchronizes its values with the employee record. Then you should use this partner record also for the user record.
+
+Usage
+=====
+
+After installation, updating an employee's `work_email` or `work_phone` record transparently changes the linked partner's appropriate field. For obvious reasons, the default for the work address being the company's address has been lifted. Partners created through the work address field will have the `employee`-flag set, and the `partner_id` field on the user record filters for this flag.
```
StefanRijnhart: I agree with the functionality provided, and I noticed this assumption. However, as setting the company address is the default in Odoo, you might want to make this into a big warning saying that things get ugly quickly if your employees actually do have the company address (this will lead to random employees' email addresses and phone numbers appear on your sale orders, and users being created from such employees forms get the company address)

hr_employee_data_from_work_address/models/hr_employee.py
```diff
@@ -0,0 +1,44 @@
...
+from openerp import models, fields, api
+
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    address_id = fields.Many2one(required=True)
+    work_phone = fields.Char(related=['address_id', 'phone'])
+    work_email = fields.Char(related=['address_id', 'email'])
+    mobile_phone = fields.Char(related=['address_id', 'mobile'])
```
StefanRijnhart, line 30: Can we have the employee's image here as well, linking to the partner image?

StefanRijnhart:
You're not making runbot happy because

```
Table 'hr_employee': unable to set a NOT NULL constraint on column 'address_id'
```

You can either enhance the hr module's demo data with addresses for the demo employees, or you can migrate the existing data for users without an address_id (or even include employees that have the company address) in the following way:
- take the users' partner_id if there is a user
- otherwise, create a new partner based on the existing data from the employee.

hbrunn:
@StefanRijnhart thanks for your input, I addressed all your points

mdietrichc2c:
LGTM, :+1:

StefanRijnhart:
Thanks! :+1:

PR #96, opened 2015-04-22
[MIG] Migrate hr_security to V8.

feketemihai:


feketemihai:
@max3903 Can you check this, the module adds some rules to hr.contract for employee and officer, the ideea is that from hr.contract the contracts button on employee and contracts menu are restricted to HR Manager group, should i modify this restrictions to allow viewing contract based on the rules specified?

ghost:
@feketemihai Yes, it would make sens.

mdietrichc2c:
LGTM, :+1:

max3903:
:+1:

PR #97, opened 2015-04-22
[MIG] Migrate hr_simplify to V8 and new API

feketemihai:
I made a few comments, waiting for your oppinions about it.


hr_simplify/models/hr_employee.py
```diff
@@ -0,0 +1,57 @@
...
+        return res
+
+    contract_id = fields.Many2one("hr.contract", string="Contract",
+                                  compute="_get_latest_contract",
+                                  store=True,
+                                  help="Latest contract of the employee")
+    job_id = fields.Many2one("hr.job", string="Job",
+                             related="contract_id.job_id")
+
+    _sql_constraints = [
```
feketemihai, line 54: Do we need this constrain, basically if we put the constrain on this, we should put on ssnid and sinid fields which are similar.

hr_simplify/models/hr_job.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding:utf-8 -*-
```
feketemihai, line 1: File not imported yet...help requested.

hr_simplify/models/hr_job.py
```diff
@@ -0,0 +1,75 @@
...
+        'no_of_employee': fields.function(
+            _no_of_contracts,
+            string="Current Number of Employees",
+            help='Number of employees currently occupying this job position.',
+            store={
+                'hr.contract': (_get_job_position, ['job_id'], 10),
+            },
+            multi='no_of_employee',
+        ),
+        'expected_employees': fields.function(
```
feketemihai, line 62: Basically this is increasing the number of forecacsted employees, isn't this supposed to modify no_of_recruitment field, by making a difference between expected_employees and no_of_employee fields.

ghost:
@feketemihai My opinion is that this module should not be ported to v8. It does many things that are not related one to another. Globally, this module is incoherent. Fields in models are rewritten (e.g. state in contract model). The name of the module is also incoherent, because this module does not simplify things, it complicates them. For this reason, I would suggest to split the module in more modules and give them proper names.

feketemihai:
@dufresnedavid I started to port the hr_contract_init to V8 and i saw that the module depends on the 'hr_job_categories', 'hr_security', 'hr_simplify', that's why i started to port the dependencies first and i didn't saw any issue regarding this modules...
Waiting for others oppinion about this...

feketemihai:
@OCA/human-resources-maintainers After thinking more we should remove this module from the repo, it's just a mix of usability, not adding a precise feature

Thoughts?

dreispt:
I agree on removal from 8.0 (and 9.0?). Maybe we should also add some info on the 7.0 description stating that it is deprecated.

andhit-r:
@feketemihai @dreispt what is the final decision regarding this module? deprecated from 8.0?

feketemihai:
From my point of view, yes...we will have to add this in the name of the module...

andhit-r:
:+1:

PR #99, opened 2015-04-24
[MIG] Migrate hr_employee_firstname to V8

feketemihai:


hr_employee_firstname/models/hr.py
```diff
@@ -0,0 +1,81 @@
...
+from openerp import models, fields, api
+
+
+class hr_employee(models.Model):
+    _inherit = 'hr.employee'
+
+    @api.cr_context
+    def _auto_init(self, cr, context=None):
+        res = super(hr_employee, self)._auto_init(cr, context=context)
+        cr.execute("""
```
ghost: Please explain what this query does in docstring and explain why it cannot be done in a "create" method. Also, add a couple unit tests that check exactly what it does.

feketemihai: I will add a test...the ideea is that if you already have employee's in database, and install after the module, it will split the name_related fields to firstname and lastname fields for the ones that are recorded...i've put the calculation about "-2" is to check if by mistake the employee's name doesn't start with a " ", and if it start's to pass it.

feketemihai:
@dufresnedavid Can you help me here...i modify the auto_init but on runbot with the demo data, it raises error. How can i init the class and update the fields firstname and lastname to be ok?
Thanks.

ghost:
@feketemihai let me a couple of days and i'll get back to you with a solution.

feketemihai:
Thanks David for your work...i just merged it to see if it's ok on runbot.

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,89 @@
...
+
+        elif vals.get('name'):
+            vals['firstname'], vals['lastname'] = self.split_name(vals['name'])
+
+        return super(HrEmployee, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        if vals.get('firstname') or vals.get('lastname'):
+            self.ensure_one()
```
dreispt, line 79: Write should work on multiple records, this prevents it.

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,89 @@
...
+            })
+
+    @api.one
+    @api.onchange('firstname', 'lastname')
+    def get_name(self):
+        if self.firstname and self.lastname:
+            self.name = ' '.join([self.firstname, self.lastname])
+
+    def _firstname_default(self):
+        return ' ' if self.env.context.get('module') else False
```
dreispt, line 59: I don't understand the purpose of this.
Can it be removed?

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,89 @@
...
+            ee.write({
+                'firstname': firstname,
+                'lastname': lastname,
+            })
+
+    @api.one
+    @api.onchange('firstname', 'lastname')
+    def get_name(self):
+        if self.firstname and self.lastname:
+            self.name = ' '.join([self.firstname, self.lastname])
```
dreispt, line 56: Shouldn't we also have a onchange on `name`?

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,89 @@
...
+#
+##############################################################################
+
+from openerp import models, fields, api, SUPERUSER_ID
+
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    def split_name(self, name):
```
dreispt, line 29: @api.model ?

hr_employee_firstname/models/hr_employee.py
feketemihai, line 56: Yes i think we should have, but it's quite strange, since we are hiding the name field in the view, i think we should add an onchange on 'resource_id.name' field and to recompute firstname and lastname of the linked employee...

dreispt, line 56: OK - the onchange is for better UI interaction; it is only needed for the editable fields on the form.
if necessary, the actual work is done in the `write()` method.
So, you can forget this comment.

hr_employee_firstname/models/hr_employee.py
ghost, line 59: I added this. @feketemihai can you add a comment to explain it. This ensures that the when the module is installing, the demo records already present in the database won't generate an error because they have null values for fields firstname and lastname. When the module is installing, the context contains {'module': 'hr_employee_firstname'}.

hr_employee_firstname/models/hr_employee.py
ghost, line 79: @dreispt You should never write the same firstname and lastname for multiple employees at the same time.

hr_employee_firstname/models/hr_employee.py
dreispt, line 59: Thanks! I didn't know about that.

hr_employee_firstname/models/hr_employee.py
StefanRijnhart, line 79: @dufresnedavid I find this constraint a bit artificial

ghost, line 79: @StefanRijnhart What would you suggest instead ?

StefanRijnhart, line 79: Don't check if first or last name are written to more than one record. It's a responsibility of the user. They might even come up with a valid use case for that.

ghost, line 79: @StefanRijnhart I agree.

hbrunn, line 79: and what the code actually does is to forbid to write either first or last name to multiple records. This two lines definitely should go away

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,89 @@
...
+        new_name = [w for w in name.split(' ') if w]
```
hbrunn, line 30: you can tell python how many parts you want: https://docs.python.org/2/library/stdtypes.html#str.split
so `name.split(' ', 1)` gives you at most two tokens

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,89 @@
...
+            vals['firstname'], vals['lastname'] = self.split_name(vals['name'])
+
+        return super(HrEmployee, self).create(vals)
+
+    @api.multi
+    def write(self, vals):
+        if vals.get('firstname') or vals.get('lastname'):
+            self.ensure_one()
+
+            vals['name'] = ' '.join([
```
hbrunn, line 81: `vals.setdefault()` in case we already write `name` for whatever reason?

hr_employee_firstname/__openerp__.py
```diff
@@ -21,34 +21,23 @@
...
-This module allows you to add firstname and lastname in employee form,
-and concatenate both in name field.
-
-Contributors
-------------
-* El Hadji Dem (elhadji.dem@savoirfairelinux.com)
-* Sandy Carter (sandy.carter@savoirfairelinux.com)
-""",
     'depends': [
         'hr',
```
pedrobaeza, line 35: You should depend of partner_firstname and fill correctly partner data.

ghost, line 35: I agree with @pedrobaeza

yvaucher:
Superseeded by #127

PR #103, opened 2015-05-10
[ADD] Module hr_period

ghost:


hr_period/README.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Install the payroll of your localisation, then install this module.
+
+
+Configuration
+=============
+
+Create a fiscal year
+--------------------
+Go to: Human Resources -> Configuration -> Payroll -> Payoll Fiscal Year
```
pedrobaeza: s/Payoll/Payroll

hr_period/README.rst
```diff
@@ -0,0 +1,95 @@
+HR Payroll Period
```
pedrobaeza, line 1: Why not call the module _hr_payroll_period_ then?

hr_period/README.rst
```diff
@@ -0,0 +1,95 @@
...
+Install the payroll of your localisation, then install this module.
+
+
+Configuration
+=============
+
+Create a fiscal year
+--------------------
+Go to: Human Resources -> Configuration -> Payroll -> Payoll Fiscal Year
+
```
pedrobaeza, line 28: I wouldn't call it "Payroll fiscal year", but "Payroll periods" or "Payroll intervals"

hr_period/README.rst
```diff
@@ -0,0 +1,95 @@
...
+None
+
+
+Credits
+=======
+
+Contributors
+------------
+
+.. image:: http://sflx.ca/logo
```
pedrobaeza: We are not putting up to now company logos on OCA modules.

hr_period/README.rst
ghost, line 1: @pedrobaeza There is already a module named hr_payroll_period. That other module has like 15 dependencies and instead of only implementing payroll periods, it is a kind of wrapper around its dependencies.

hr_period/README.rst
ghost, line 28: @pedrobaeza There's 2 models. The first is the payroll period (e.g. a month or a week or ..). The second is the Fiscal year, which is an object that helps create and manage periods over a year. Though, I wouldn't mind renaming Payroll Fiscal Year to Payroll Year.

ghost:
@feketemihai I have a few active pull requests. I would appreciate your reviews very much. Thanks

hr_period/models/hr_fiscal_year.py
```diff
@@ -0,0 +1,288 @@
...
+        ]
+
+        self.pool['hr.period'].button_set_to_draft(
+            cr, uid, period_ids, context=context)
+
+        # Set the fiscal year to draft
+        self.write(cr, uid, ids, {'state': 'draft'}, context=context)
+
+    def search_period(self, cr, uid, ids, number, context=None):
+        fy = self.browse(cr, uid, ids[0], context=context)
```
bwrsandman, line 309: Make sure ids is a list and that it has only one item

bwrsandman:
:+1:
Looks good technically

hr_period/models/hr_fiscal_year.py
```diff
@@ -0,0 +1,288 @@
...
+
+INTERVALS = {
+    'annually': (relativedelta(months=12), 1),
+    'semi-annually': (relativedelta(months=6), 2),
+    'quaterly': (relativedelta(months=3), 4),
+    'bi-monthly': (relativedelta(months=2), 6),
+    'monthly': (relativedelta(months=1), 12),
+    'bi-weekly': (relativedelta(weeks=2), 26),
+    'weekly': (relativedelta(weeks=1), 52),
+}
```
feketemihai, line 42: Is it ok to include 'daily' as well.
, there are countries that use daily workers, so at the end of the day to pay for it.

feketemihai:
Can you add a multicompany rule on fiscalyear, it will simplify the views and ease the input, otherwise the user will get some warnings if he select a fiscalyear from other company..

ghost:
@feketemihai In the multi-company setup, the hr_manager should be able to access fiscalyears for any company. If someone needs a more restrictive setup, in my opinion, it should be done in another module.

Also, in the payslip and payslip batch views, the user is not able to select a period and a company that dont match. This is ensured by the domain on hr_period_id and the onchange_company_id method.

ghost:
@feketemihai Added daily schedule and fixed the methods as you suggested.

hr_period/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'HR Payroll Period',
+    'version': '1.0',
+    'license': 'AGPL-3',
+    'category': 'Generic Modules/Human Resources',
+    'description': """
```
dreispt: Remove description, since it will be filled with the README.
You could consider having a one line `summary` though.

hr_period/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+HR Payroll Period
+=================
+Add payroll periods
+""",
+    'author': "Savoir-faire Linux,Odoo Community Association (OCA)",
+    'website': 'https://odoo-community.org',
```
dreispt: I like to use here the repo's home address (https://github.com/OCA/hr), that that's just a personal opinion.

max3903: or http://www.savoirfairelinux.com

ghost:
@dreispt I only put a description in **openerp**.py because otherwise I get a warning when the module is installing.

dreispt:
Forget about my description comment - I didn't notice it targets 7.0.
AFAICS looks spotless :-)
:+1:

ghost:
@max3903 I guess we can merge here.

PR #108, opened 2015-05-17
7.0: make runbot build green

gurneyalex:
- remove type in ir.ui.view xml definitions
- fix a Python SyntaxWarning caused by wrong use of `assert`
- fix the addition of required columns by ensuring they get a value before the NOT NULL sql constraint is added 
- changed WARNING logging to DEBUG where the original logging level was clearly overrated

### DIFF
--- a/hr_contract_reference/hr_contract_view.xml
+++ b/hr_contract_reference/hr_contract_view.xml
@@ -1,11 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
     <data>
-        
+
         <record id="hr_contract_view" model="ir.ui.view">
             <field name="name">hr.contract.form.view.inherit.ref</field>
             <field name="model">hr.contract</field>
-            <field name="type">form</field>
             <field name="inherit_id" ref="hr_contract.hr_contract_view_form"/>
             <field name="arch" type="xml">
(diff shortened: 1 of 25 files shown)


pedrobaeza:
Still two PEP8 issues:

```
./hr_employee_firstname/hr.py:36:70: E502 the backslash is redundant between brackets
./hr_employee_firstname/hr.py:37:38: F821 undefined name 'pg_varchar'
```

gurneyalex:
Hi @pedrobaeza, thanks for the review, but this is still WIP. There are lots of warnings to fix...

gurneyalex:
Should be fine now. 

I'm not a user of OCA/hr, so I'm not sure what the best option is regarding the two last commits of this PR which create a new default record for a required field of the contracts so that all the contracts have their constraints satisfied when the modules are installed on a pre-existing database.

pedrobaeza:
Good job! :+1: 

I also doesn't know exactly that module, but I think the solution you propose is good enough for a start.

gurneyalex:
@dufresnedavid having this merged should help getting green builds on runbot for PR targeting the 7.0 branch. Thanks for your review.

hbrunn:
:+1:

guewen:
:+1:

PR #109, opened 2015-05-28
Do not copy the expense sequence

guewen:


pedrobaeza:
:+1: 

I fast track it as obvious.

PR #110, opened 2015-06-02
[FIX] Prefer view constraint over database constraint for compatibili…

StefanRijnhart:
…ty with other modules' tests. Fix courtesy of @gurneyalex


mdietrichc2c:
:+1: while waiting for Travis.

pedrobaeza:
:+1: Fast-tracking as this is a needed patch.

PR #111, opened 2015-06-03
[FIX] recalculate image_{small,medium} for partners

hbrunn:
without this, it looks like nothing happened for images because in the ui, we only see `image_medium`

### DIFF
--- a/hr_employee_data_from_work_address/README.rst
+++ b/hr_employee_data_from_work_address/README.rst
@@ -1,7 +1,7 @@
 User data from employee
 =======================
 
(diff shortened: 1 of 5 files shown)


ghost:
@hbrunn Can you add unit tests for this module ?

hbrunn:
@dufresnedavid usually, I'd gladly say 'yes', but this module doesn't really do anything interesting code wise. What would you suggest to test?

ghost:
adjust_employee_partners_pre and adjust_employee_partners_post

I rather think that what they are very interesting codewise, since you use the cursor directly.

hbrunn:
Those are init hooks. They will run anyways, adding a test won't increase code coverage I think?

ghost:
Obviously, it will, since your PR decreases the coverage. Also, the coverage is one thing, but verifying that your code does what is expected from it is another.

hbrunn:
@dufresnedavid now coveralls is happy

ghost:
:+1: Thanks

bwrsandman:
:+1:

PR #114, opened 2015-06-12
[IMP] Fix ytd amount on refund payslips

ghost:


yvaucher:
Could you add a test?

PR #125, opened 2015-08-11
8.0 hr emergency contact

Zwygart:


hr_emergency_contact/emergency_contact.py
```diff
@@ -0,0 +1,39 @@
...
+    ec_name = fields.Char('Name', size=256, required=True)
+    ec_relationship = fields.Char('Relationship', size=128, required=True)
+    ec_tel1 = fields.Char('Primary Phone No.', size=32)
+    ec_tel2 = fields.Char('Secondary Phone No.', size=32)
+    ec_woreda = fields.Char('Subcity/Woreda', size=32)
+    ec_kebele = fields.Char('Kebele', size=8)
+    ec_houseno = fields.Char('House No.', size=8)
+    ec_address = fields.Char('Address 2', size=256)
+    ec_country_id = fields.Many2one('res.country', 'Country')
+    ec_state_id = fields.Many2one('res.country.state', 'State', domain="[('country_id','=',country_id)]")
```
dreispt, line 39: Wouldn't it be better to use an `emergency_partner_id` instead?

hr_emergency_contact/hr.py
```diff
@@ -0,0 +1,48 @@
...
+    ec_name = fields.Char('Name', size=256)
+    ec_relationship = fields.Char('Relationship', size=64)
+    ec_tel1 = fields.Char('Primary Phone No.', size=32)
+    ec_tel2 = fields.Char('Secondary Phone No.', size=32)
+    ec_woreda = fields.Char('Subcity/Woreda', size=32)
+    ec_kebele = fields.Char('Kebele', size=8)
+    ec_houseno = fields.Char('House No.', size=8)
+    ec_address = fields.Char('Address 2', size=256)
+    ec_country_id = fields.Many2one('res.country', 'Country')
+    ec_state_id = fields.Many2one('res.country.state', 'State', domain="[('country_id','=',country_id)]")
```
dreispt, line 39: This is duplicated

hr_emergency_contact/emergency_contact.py
pedrobaeza, line 39: I agree with @dreispt

Zwygart:
This is one of my first requests. I just took a module from the list and wonder if I did the process the right way. (Pull, migrate, push)

If we (or I) should modify the module (not only migrate) it's up to you. 

Yes, it makes sense to just use a emergency_partner_id of a employee to put as emergency info.

pedrobaeza:
Yes, please work on these changes, but first, you should move the module from `__unported__` instead of extracting the commit history and added again. You should start from the current 8.0 branch, make `git mv __unported__/hr_emergency_contact ./`, and then make your changes.

Zwygart:
I followed the instructions from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-8#howto hoping it was the right way. I don't see instructions about **unported** .

dreispt:
@Zwygart To give you some perspective:
In most cases you can migrate a module doing just some technical adjustments to it.
But the `hr` modules were added to OCA in bulk, and didn't have a thorough peer reviews.
So there might be some inadequate functional implementations that can mean refactoring the module.
This is such a case.

Zwygart:
Ok, I see.

pedrobaeza:
@Zwygart. Sorry, that instructions were in case the module is not present already in the branch. But that's not the case. I will ask for an update on this documentation to avoid this situation again.

Zwygart:
Now, when I try to push, it says 

error: src refspec 8.0-hr_emergency_contact does not match any.
error: impossible de pousser des références vers 'git@github.com:Zwygart/hr.git'

What do I do wrong again?

max3903:
@Zwygart git does not find the local branch 8.0-hr_emergency_contact. Try with `local_branch_name:distant_branch_name`, like `8.0:8.0-hr_emergency_contact`

Zwygart:
I tried, and it worked with a normal push. And with a 

"git push -f 8.0:8.0-hr_emergency_contact HEAD" it was hanging. 

But, nothing appears here.

Zwygart:
gideon@gideon:~/OCA/hr/hr_emergency_contact$ git push -f myrepo HEAD
Counting objects: 256, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (71/71), done.
Writing objects: 100% (162/162), 50.24 KiB | 0 bytes/s, done.
Total 162 (delta 101), reused 147 (delta 89)
To git@github.com:zwygart/hr.git
- 6ff661a...354e57b HEAD -> 8.0 (forced update)

Why doesn't it appear here?

feketemihai:
Superseeded in #204 with direct link to contacts. I will close this one.

PR #132, opened 2015-08-18
Expense By Month

rudyvalentin:
The module add the month for your expense


pedrobaeza:
Hi, Rudy, can you explain the rationale behind this module?

rudyvalentin:
hi pedrobaeza, 
Each month the employee must make an expence . This module allows you to associate a monthly expense

pedrobaeza:
OK, thanks for the explanation. I let other experts like @dufresnedavid to tell about the convenience of this one.

hr_expense_months/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied,
+# including but not limited to the implied warranties
+# of merchantability and/or fitness for a particular purpose
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>
+{
+    'name': 'HR Expenses Upgrade',
```
ghost, line 20: The name of the module does not tell us anything about the module. Please change it to something like 'HR Expense Month'.

hr_expense_months/hr_expense_months.py
```diff
@@ -0,0 +1,94 @@
...
+            ('Mai', 'May'),
+            ('Juin', 'June'),
+	     ('Juillet', 'July'),
+	     ('Aout', 'August'),
+	     ('Septembre', 'September'),
+	     ('Octobre', 'October'),
+	     ('Novembre', 'November'),
+	     ('Decembre', 'December'),],
+            'Months',readonly=True,required=True,states={'draft':[('readonly',False)]}),
+    }
```
ghost, line 81: You redefine columns that already exist. This is extremely bad. Also, you change the type of a column (name -> from char to selection). This makes your module incompatible with other modules. This will make unit tests very hard to maintain.

The column 'months' should be a function field based on the field 'date'. Also, it should be month instead of months since you select only one month.

hr_expense_months/hr_expense_months.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def onchange_months(self, cr, uid, ids, mois):
```
ghost, line 83: mois -> french parameter ?

Also, please use the v8 api.

ghost:
:-1: Please use a linther, follow the OCA programming standards, use the v8 api. I'm not sure I get the use case of this module.

max3903:
@rudyvalentin Please split this PR. One for each of the 2 modules. Thanks!

hr_expense_months/hr_expense_months.py
```diff
@@ -0,0 +1,94 @@
...
+from openerp.osv import fields, osv, expression
+from openerp.tools.translate import _
+from openerp.tools.float_utils import float_round as round
+from openerp.tools.misc import DEFAULT_SERVER_DATE_FORMAT as DT_FORMAT
+
+import openerp.addons.decimal_precision as dp
+
+_logger = logging.getLogger(__name__)
+
+class hr_expense_expense(osv.osv):
```
ehdem, line 42: @rudyvalentin : Please use new API

feketemihai:
:-1:

PR #143, opened 2015-10-01
[ADD] hr_payslip_move_date

JordiBForgeFlow:
# HR Payroll Account Date

Adds a move date to the payslip.
Allows the user to propose/force a specific move date.
If none is proposed, the field will be completed with the actual move date.
# Usage
## Create a payslip

Go to: Human Resources -> Payroll -> Employee Payslip
- Complete the payslip details
- Go to 'Other information', 'Accounting' section and complete the field
  'Force move date'.
# Known issues / Roadmap

By default the Salary Journal should be set up with the field 'Check Date in
 Period' unflagged. Otherwise the application will raise an error message if
  the user enters a move date corresponding to a past period.

### DIFF
--- a/hr_payslip_move_date/models/hr_payslip.py
+++ b/hr_payslip_move_date/models/hr_payslip.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    Copyright (C) 2014 Eficent (<http://www.eficent.com/>)
+#               <jordi.ballester@eficent.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from openerp.osv import fields, osv, orm
+
+
+class HrPayslip(orm.Model):
+
+    _inherit = 'hr.payslip'
+
+    _columns = {
+        'move_date': fields.date(string='Force move date',
+                                 required=False),
+    }
+
+    def onchange_move_date(self, cr, uid, ids, move_date, context=None):
+        res = {'value': {}}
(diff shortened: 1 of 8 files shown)


ghost:
No support for payslip batches ?

JordiBForgeFlow:
Ok, will add it.
El dia 01/10/2015 20:32, "David Dufresne" notifications@github.com va
escriure:

JordiBForgeFlow:
@dufresnedavid Done. Thanks for the feedback!

hr_payslip_move_date/models/hr_payslip.py
```diff
@@ -0,0 +1,73 @@
...
+        res = super(HrPayslip, self).write(cr, uid, ids, vals, context=context)
+        if 'move_id' in vals and vals['move_id']:
+            for slip in self.browse(cr, uid, ids, context=context):
+                if slip.move_date:
+                    move_obj.write(cr, uid, [slip.move_id.id],
+                                   {'date': slip.move_date}, context=context)
+                    for move_line in slip.move_id.line_id:
+                        move_line_obj.write(cr, uid, [move_line.id],
+                                            {'date': slip.move_date},
+                                            context=context)
```
ghost, line 58: @jbeficent you should write in every move lines in one time. Otherwise, this will take a lot of time. The orm overhead in account moves in high in Odoo. It triggers many checks every time.

move_line_obj.write(cr, uid, all_move_line_ids, {...}, context=context)

hr_payslip_move_date/models/hr_payslip.py
```diff
@@ -0,0 +1,73 @@
...
+        return res
+
+
+class HrPayslipRun(orm.Model):
+
+    _inherit = 'hr.payslip.run'
+
+    _columns = {
+        'move_date': fields.date(string='Force move date',
+                                 required=False),
```
ghost, line 72: dont mention required if it is False

ghost:
Looks good. Could you add unit tests ?

hr_payslip_move_date/models/hr_payslip.py
JordiBForgeFlow, line 58: Well, the problem is at the time when the write is called, each payslip could potentially have a different date. Also, the date is not passed to the 'write' vals. Only the move_id is passed. I'm afraid that there's no other option, AFAIK.

ghost, line 58: yes you have at least one option. Just write the date in the account.move, then every account.move.line will be updated according to the account.move related.

ehdem:
@jbeficent : Fix travis errors https://travis-ci.org/OCA/hr/jobs/83287651

andhit-r:
Hi @OCA/human-resources-maintainers i think this PR can be closed. The module already merged on https://github.com/OCA/hr/pull/220

PR #148, opened 2015-10-05
[ADD] hr_payroll_allowance

saltonmassally:
Adds the concept of payroll allowances
The aim of this addon is to take away the complexity associated with setting
up salary rules and editing service structures to reflect allowances being paid
to employees.
Also from experience a lot of companies in West Africa (may also apply to other
regions) do not have standardized pay structures with employees in the same job
having different allowances.

### DIFF
--- a/hr_payroll_allowance/README.rst
+++ b/hr_payroll_allowance/README.rst
@@ -1,0 +1,58 @@
+HR Payroll Allowances
+=====================
+
+Adds the concept of payroll allowances
+The aim of this addon is to take away the complexity associated with setting
+up salary rules and editing service structures to reflect allowances being paid
+to employees.
+Also from experience a lot of companies in West Africa (may also apply to other
+regions) do not have standardized pay structures with employees in the same job
+having different allowances.
+
+
+Installation
+============
+
+Install as usual from the module installation screen
+
+
+Configuration
+=============
+
+* Go to Human Resources > Configuration > Salary Allowance Category and create
+the allowance categories applicable to you company
+
+* Go to the employee's contract and insert figures for the allowances
+
+
+Usage
+=====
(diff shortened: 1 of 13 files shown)


ghost:
@tarzan0820 ok, I get that this module adds allowances, but how does it get to the payslip ? Also, have you checked my module hr_employee_benefit ? I did not make the distinction on what is an employee benefit and what is an allowance. Both are manageable the same way as one model.

saltonmassally:
@dufresnedavid 

When allowances are created the following are done - salary rule created of the category ALW, contribution register created, salary rule is automatically added to the salary structure specified and fields are added to the hr.contract module to store data. These fields are dynamically presented on the hr.contract form. So as long as the salary rule on the contract matches that specified on the allowance it should show up on the payslip if the allowance conditions are met by the employee/contract... I hope I explained well

Yes I did see hr_employee_benefit and followed the porting process closely even downloaded and tried it as I think this reimplements some of what it does however I found it to be a little too much for what I needed... with this the user creates the list of allowances being offered by his/her company and that's all.... not tinkering with salary rules/salary structures

ghost:
@tarzan0820 ok, I think we have 2 distinct ways of thinking regarding allowances, but I think both are acceptable. You should however add unit tests to your module. It will make this much more easy to understand your patern.

saltonmassally:
@dufresnedavid any chance that you can review this?

feketemihai:
For me, it's the same thing as the hr_employee_benefit module, so i don't think we should duplicate features.

feketemihai:
@tarzan0820 I stick to my opinion that the 2 concepts are the same, and we should not implement it twice, instead if you want you can improve the hr_employee_benefit module.
@OCA/human-resources-maintainers Any other reviews?

feketemihai:
:-1:

PR #149, opened 2015-10-05
[ADD] hr_payroll_usability

saltonmassally:
This module makes several small usability improvements to odoo's handling of
payroll.
- Ability to define a default payroll structure for contracts
- Ability to filter which payroll structure can be selected on contracts

### DIFF
--- a/hr_payroll_usability/README.rst
+++ b/hr_payroll_usability/README.rst
@@ -1,0 +1,63 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+=================
+Payroll Usability
+=================
+
+This module makes several small usability improvements to odoo's handling of 
+payroll.
+
+* Ability to define a default payroll structure for contracts
+* Ability to filter which payroll structure can be selected on contracts
+
+Installation
+============
+
+To install this module, you need to:
+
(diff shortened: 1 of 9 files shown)


nhomar:
@tarzan0820  can you change the name for something more specific than 'usability' it is better if the module is named as the thing it is doing?.

i.e. hr_payroll_det_defaults < but propose your own.

saltonmassally:
@nhomar I see your point; however one of the major reason for this addon is to house very small improvements for which an entire addon might be seen as an overkill... I looked at other OCA repos and they have such x_usability modules also...

Let's maybe get some feedback from others and if the community feels strongly about it I can break up up the features into specific addons and rename appropriately

nhomar:
@tarzan0820 yes frequently we think that, but what is an usability improvement for you can be wrong in a different conext, that's why as specific is your approach it can be installed/unistalled or implemented n a more secure way.

Maybe in other repos they were legacy modules from other places if you point me out one of them we can make the fix... some times we need to do what we say and not what we do ;-)

saltonmassally:
@nhomar point well taken I will make the change

saltonmassally:
moved to :
https://github.com/OCA/hr/pull/155
https://github.com/OCA/hr/pull/156

PR #160, opened 2015-10-14
[ADD] hr_contract_default_trial_length

saltonmassally:
This addon allows you to define default trail length for contracts


hr_contract_default_trial_length/__init__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+###############################################################################
+#
+#    Copyright (C) 2016 Salton Massally (<smassally@idtlabs.sl>).
```
dreispt: 2016?

hr_contract_default_trial_length/__openerp__.py
```diff
@@ -0,0 +1,36 @@
...
+###############################################################################
+{
+    "name": "Default Contract Trail Length",
+    'version': '8.0.1.0.0',
+    'author': "Salton Massally <smassally@idtlabs.sl>, "
+               "Odoo Community Association (OCA)",
+    "website": "http://idtlabs.sl",
+    "category": "Human Resources",
+    "summary": "Define default trail length for contracts",
+    "depends": [
```
dreispt, line 29: Add license key

hr_contract_default_trial_length/models/hr_contract.py
```diff
@@ -0,0 +1,36 @@
...
+
+class HrContract(models.Model):
+    _inherit = 'hr.contract'
+
+    @api.onchange('trial_date_start')
+    @api.multi
+    def onchange_trial_date_start(self):
+        self.ensure_one()
+        if self.trial_date_start:
+            res = self.env.user.company_id.default_contract_trial_length
```
dreispt: In my country's case, default trail duration depends on contract type and, fro non-permanent contract, on contract duration. To fit this a configuration table would be needed. It could have a column with an expression to evaluate, and a duration field.
You may consider this out of scope, but if so please add this idea to the Roadmap section in the README.

hr_contract_default_trial_length/views/res_company.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="hr_timesheet_sheet_company" model="ir.ui.view">
+            <field name="name">res.company.sheet</field>
+            <field name="model">res.company</field>
+            <field name="inherit_id" ref="base.view_company_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//group[@name='account_grp']" position="after">
```
dreispt: You could use the more succinct form: `<group name="account_grp" position="after">`

dreispt:
IMO the feature is not powerful enough, but not a blocking issue. It would be nice if you could improve it as per suggested.
Other than that, only minor notes, look good to go.

saltonmassally:
@dreispt I have agreed with your reasoning and move trial length to contract type... I added the rest to the roadmap... please review when chanced

feketemihai:
I agree that we should also check the type of contract and duration, but if this is in the roadmap, it's ok for now.
:+1:

PR #165, opened 2015-10-15
8.0 hr employee exemption

ghost:


hr_employee_exemption/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Employee Exemption',
+    'version': '1.0',
```
saltonmassally: Use correct versioning convention

hr_employee_exemption/__openerp__.py
```diff
@@ -0,0 +1,47 @@
...
+        'hr_payroll',
+    ],
+    'data': [
+        'security/ir.model.access.csv',
+        'view/hr_employee_view.xml',
+        'view/hr_income_tax_exemption_view.xml',
+        'view/hr_salary_rule_view.xml',
+    ],
+    'test': [],
+    'demo': [],
```
saltonmassally: Avoid empty keys

hr_employee_exemption/models/hr_employee.py
```diff
@@ -0,0 +1,53 @@
...
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    exemption_ids = fields.One2many(
+        'hr.employee.exemption',
+        'employee_id',
+        'Income Tax Exemptions',
+        groups='base.group_hr_manager',
+    )
+
```
saltonmassally, line 34: Wouldn't this be better on hr.contract? In my opinion payroll related features are better off on hr.contract

ghost, line 34: If someone is exempted from a tax, it has nothing to do with his contract. The person would be exempted whatever job he has.

ghost:
Anyone has an idea why travis can not install the module partner_firstname ?

feketemihai:
@dufresnedavid I think the logic is quite different, you can have one salary rule and many exemtions based on the employee, so i suggest to change the relation between salary.rule and hr.income.tax.exemption.

ghost:
@feketemihai I don't understand the use case behind your question.

Example, I have one rule that computes the Canada Income Tax to withdraw for the period. The employee is ether exempted or not from the Canada Income Tax. So I create one exemption which might be named "Canada Federal Income Tax" and then I assign it to the employee.

feketemihai:
Agree with you but instead of having one2many to salary rule i said to have a many2one so you could add different exemption per one salary rule...dependant of some condition of the employee...

Sent from my Samsung device

-------- Original message --------

ghost:
@feketemihai I think that this would be too much complicate for the person who configures the system. Exemptions are rare cases. I prefer let the hr manager attribute an exemption manually to an employee than relying on a script to figure out if the employee is exempted, based on some conditions.

feketemihai:
Ok then...
:+1:

feketemihai:
Could you also add issue paragraph in readme file and Try me on runbot at the end of the Usage paragraph.
Thanks..

PR #174, opened 2015-10-21
hr_skill improves

yelizariev:
I've updated hr_skill module for myself. Probably it could be useful for OCA

### DIFF
--- a/hr_skill/README.rst
+++ b/hr_skill/README.rst
@@ -29,6 +29,7 @@ Contributors
 
 * Savoir-faire Linux
 * Daniel Reis
+* Ivan Yelizariev
 
 Maintainer
 ----------
--- a/hr_skill/__openerp__.py
+++ b/hr_skill/__openerp__.py
@@ -20,7 +20,7 @@
 
 {
     "name": "Skill Management",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.1.0",
     "category": "Human Resources",
     "license": "AGPL-3",
     "author": "Savoir-faire Linux,Odoo Community Association (OCA)",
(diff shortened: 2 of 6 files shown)


max3903:
:+1: 

Can you bump the version ?

yelizariev:
it has 

```
"version": "8.0.1.0.0",
```

Can I change it to 

```
"version": "1.0.1",
```

or I have to keep reference to odoo version?

pedrobaeza:
The idea to put 8.0 before the version number is to know for which Odoo version is the module, because it's not clear sometimes when you get a zip, or is in a forked repository in a branch not named with the version number. As a module update is also needed, the version should be 8.0.1.1.0. See https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

dreispt:
:+1:

feketemihai:
@yelizariev @dreispt Do you think it's worth it to add a type in skills, like on every parent-child relation of type view, normal...basically if you assign a skill to the employee and that skill have 3 childs, this means that the employee should have those 3 skills to get the parent one...

It's just an oppinion.

Beside that, please add you to the authors and contribuitors.
:+1:

yelizariev:
@feketemihai the same can be done via parent_id, child_ids fields. You don't need special field to specify type. The problev with such field is that you have edit it whenever you decide to make skill more detailed. E.g. you have skill "Web-programing" (type=skill) and then you decide to add childs to it:

```
* web-programming (type=view)
  * js
  * css
  * html
```

then you have to change web-programing type to view. In my opinion we can skip it use parent_id, child_ids fields to identify type of a record.

PR #178, opened 2015-10-23
[FIX] Incorrect logic to determine if two periods overlap.

JordiBForgeFlow:
The current logic to determine if there's a timesheet that overlaps with the payslip period is incorrect. 

For example, given a pay period: 
From: 01/07/2015
To: 01/20/2015

And an approved employee timesheet with date ranges:
From: 01/05/2015
To: 01/27/2015

The current logic will fail to identify that the two periods overlap (but it's obvious that they do overlap..).

The fix consists of applyting the simple logic described here:
http://c2.com/cgi/wiki?TestIfDateRangesOverlap

### DIFF
--- a/hr_worked_days_from_timesheet/hr_payslip.py
+++ b/hr_worked_days_from_timesheet/hr_payslip.py
@@ -44,11 +44,6 @@ class hr_payslip(orm.Model):
         worked_days = {}
         # Create one worked days record for each timesheet sheet
         for ts_sheet in timesheet_sheets:
-            # Get formated date from the timesheet sheet
-            date_from = datetime.strptime(
-                ts_sheet.date_from,
-                DEFAULT_SERVER_DATE_FORMAT
-            ).strftime(date_format)
             # Create a worked days record with no time
             worked_days[ts_sheet.id] = {
                 'name': _('Timesheet %s') % date_from,
--- a/hr_worked_days_from_timesheet/hr_payslip.py
+++ b/hr_worked_days_from_timesheet/hr_payslip.py
@@ -146,8 +146,8 @@ class hr_payslip(orm.Model):
             if (
                 # We need only the timesheet sheets that overlap
                 # the payslip period.
-                date_from <= ts_sheet.date_from <= date_to or
-                date_from <= ts_sheet.date_to <= date_to
+                date_from <= ts_sheet.date_to and
+                ts_sheet.date_from <= date_to
 
                 # We want only approved timesheets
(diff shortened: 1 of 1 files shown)


ghost:
@jbeficent yep, the logic is incorrect, but you make things worse. Consider the following:

date_from <= ts_sheet.date_from <= date_to or
date_from <= ts_sheet.date_to <= date_to or
(ts_sheet.date_from <= date_from and date_to <= ts_sheet.date_to)

JordiBForgeFlow:
@dufresnedavid can you point a singlen case where the proposed fix is incorrect? I cannot

JordiBForgeFlow:
Let ConditionA Mean that DateRange A Completely After DateRange B
----------------------------------- |---- DateRange A ------| 
|---Date Range B -----| 
(True if StartA > EndB)

Let ConditionB Mean that DateRange A is Completely Before DateRange B
|---- DateRange A -----|
---------------------------------|---Date Range B ----|
(True if EndA < StartB)

Then Overlap exists if Neither A Nor B is true -
(If one range is neither completely after the other,
nor completely before the other, then they must overlap.)

Now one of De Morgan's laws says that:

Not (A Or B) <=>  Not A And Not B

Which translates to (StartA <= EndB)  and  (EndA >= StartB)

JordiBForgeFlow:
@dufresnedavid I just pushed another fix for the logic to determine the hours worked. Only the timesheet activities that fall within the payslip period shall be considered. The previous logic considered the start date of the timesheet and the end date of the payslip, which generated a ficticious interval.

ghost:
@jbeficent take the case ts_sheet.date_start < date_from and ts_sheet.date_to < date_to. Here your algo does not work.

JordiBForgeFlow:
@dufresnedavid Well, you are specifying a case that does not really test the algorithm, because your test could be either:

----------------------------------- |---- Payslip ------| 
|---Timesheet-----| 

Or:
----------------------------------- |---- Payslip ------| 
----------------------|---Timesheet-----| 

Which one do you refer to? 

The algorithm will test date_from against date_two of the other period. Therefore you should state a case where this comparison is challenged.

gurneyalex (approved):
it would be obvious with an automated test...

pedrobaeza:
Yes, please include a unit test for that!

gurneyalex (changes requested):
needs a test, a rebase and a travis fix

feketemihai:
@jbeficent Any news on this, do you still need it?

feketemihai:

PR #180, opened 2015-10-25
[ADD]  hr_employee_endofservice_benefit

saltonmassally:
This module calculates employee's end of service benefit a benefit paid to
an employee at end of service. Calculation is based on the time spent by
employee in company.

Limited to my use case but improvement suggestions are welcome

### DIFF
--- a/hr_employee_endofservice_benefit/README.rst
+++ b/hr_employee_endofservice_benefit/README.rst
@@ -1,0 +1,73 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============================
+Employee End of Service Benefit
+===============================
+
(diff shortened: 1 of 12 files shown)


ghost:
@tarzan0820 why not using the payroll system ? This could be python computed salary rule. Also, your calculations seem trivial. These kind of benefits rely on actuarial formulas. Also, how do you manage the accounting ?

saltonmassally:
@dufresnedavid like I pointed out in my PR comment this addon is pretty limited to my use case but I am willing to develop it further based on suggestions.
I like your idea of using the payroll system... Are you suggesting that I attach a salary rule to the contract category that will be evaluated whenever we need to calculate terminal benefit?

ghost:
@tarzan0820 this case is complex, I don't have a clear answer on how to develop your module. The problem I see is that we are getting a lot of incomplete/incoherent modules on OCA. If we merge your module, after that, it will be very difficult to upgrade it to a better version. What I think you should do is continue thinking about your module, make your ideas clear and try your module in the real world. Then you might end up proposing something much more interesting.

PR #190, opened 2015-11-10
[FIX] Update tests in hr_employee_firstname - failing travis

feketemihai:
Fix tests in hr_employee_firstname, pr's are falling by this, the method _get_inverse_name return a dict with string indexes, so the values updated on employees were all the time as 'firstname' and 'lastname' string, the indexes from the method, not the values.


feketemihai:
@OCA/human-resources-maintainers I don't get it why on OCB is failing on lxml...

pedrobaeza:
:+1: 

Fast-tracking to allow to greenify all PRs

PR #189, opened 2015-11-06
hr_worked_days_from_timesheet migrate and re-factored to Odoo 8.0

acysos:
Hello,

I have migrate and re-factored this module. Less code and improved for inherit.

Greetings,
Ignacio Ibeas

PR #194, opened 2015-11-18
Update split names at module install to get the value not the key of the dict

feketemihai:
At install of the module, the auto_init is launching the split_name script, which splits the name of the employee in firstname and lastname, but the method gets the key of the dict, not the value returned.

This fix is for getting the value and not the key from the dict, just like the test pr for this module.


andhit-r:
@feketemihai I rebased #189 on top of this PR. Travis check passed. So :+1:

feketemihai:
@OCA/human-resources-maintainers Can you review this, it's raising error depending on the order the modules are installed.

PR #198, opened 2015-12-03
[IMP] make our related fields nonstored

hbrunn:


pedrobaeza:
Isn't by default non-stored?

hbrunn:
Weirdly enough, no. Or maybe not so weirdly: They are not new fields, but we make normal fields from hr related fields. At this stage, we end up with the store-flag being true.

pedrobaeza:
OK, and why that strange way to define related instead of a string ("relational_field.sub_field")?

Anyway :+1:

hbrunn:
@pedrobaeza actually, the list version is the more canonical one: https://github.com/OCA/OCB/blob/9.0/openerp/fields.py#L509

pedrobaeza:
Oh, I see. Interesting to know that!

Is Travis error related?

hbrunn:
yes, and I'm struggling to find out what goes wrong

gurneyalex:
@hbrunn what's the status of this PR?

hbrunn:
still no clue how to solve this to be honest

hbrunn:
@gurneyalex now the status is green :-)

gurneyalex (approved):

moylop260 (approved):

PR #203, opened 2016-03-02
[FIX] hr_employee_state: Exception: object of type 'generator' has no len()

tremlin:
The method `_no_of_employee` raises an exception because 

> object of type 'generator' has no len()

### DIFF
--- a/hr_employee_state/hr.py
+++ b/hr_employee_state/hr.py
@@ -470,8 +470,8 @@ class hr_job(orm.Model):
         res = {}
         count = 0
         for job in self.browse(cr, uid, ids, context=context):
-            count = len(
-                ee for ee in job.employee_ids
+            count = sum(
+                1 for ee in job.employee_ids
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It's easier to just enclose the expression between `[]` and you'll get it.

tremlin:
I know that this alternative solution exists, but I don't see how it's easier.

pedrobaeza:
Well, it's more readable...

hbrunn:
how about `len(job.employee_ids.ids)`?

OSguard:
@hbrunn there is a filter `ee.status != 'pending_inactive'` in next line for the count

hbrunn:
ah, sorry. no commenting after first glance for me today ;-)

hbrunn:
as @tremlin's version is slightly more efficient than @pedrobaeza's proposal, :+1:

pedrobaeza:
OK, no problem for me :+1:

PR #206, opened 2016-03-08
Fixing problem with migration in a function

BT-rmartin:
Setting priority to avoid troubles with views


dreispt:
@BT-rmartin anything wrong?

BT-rmartin:
@dreispt I got some errors about the number of parameters for function _update_employee_names or for search within this. As syntax for v7 and v8 are mixed in the functions this leads to that trouble. As _update_employee_names is only called from _auto_init, I decided to refactor this so that no problem is found.

dreispt:
OK. If you can drop a line stating the reasons when closing a PR, that would be helpful.
In this case, if you're refactoring it, you could have added `[WIP]` to the PR name.

PR #173, opened 2015-10-21
8.0 hr public holidays

saltonmassally:

PR #209, opened 2016-03-09
Update Hr public holidays.

feketemihai:
This is a update from #173 . I will close that one.


andhit-r:
Hi @feketemihai. I tried:
1. Create international public holiday for 2016
   selection_1
2. I tried to create indonesia's public holiday for 2016
   selection_2
3. Then i got this error
   selection_3

andhit-r:
Funny though, i could create 2017 international and 2017 indonesia public holiday. But not 2016

feketemihai:
It depends on the order that you input..if you input by country first the error is not raised..if you input internationally first the error will be raised...i will work on the constraint function.

andhit-r:
Yes i confirm that. It works just like you said.

feketemihai:
It should be fixed now.

andhit-r:
Thanks @feketemihai . Btw is this by design? Or need extra constraints?

selection_5

feketemihai:
There wasn't any constraint on this...we can have a new constraint on this just like for the year country, here to have it by date/states...

andhit-r:
@feketemihai thanks for extra constraints. One more case still bothers me:

selection_6

When a data declared as national holiday then it should not possible to declare that day as state holiday. Is it possible to add more constraint, or just inform it on known issue?

andhit-r:
Well at least in my country (Indonesia) that will not happen. Or it's possible in other countries?

andhit-r:
Btw let me know if you need any help to close this PR or any untouch PR on OCA/hr. Thanks.

andhit-r:
:+1:

azmimr67:
@feketemihai @andhit-r I'm thinking that constraint that you put on public holiday for national and state cannot be for same day.
I think this still can happen example 
17/04/2016       Nyepi
17/04/2016       State Election                                      Tangerang

Because I think the purpose of this module is to record all public holiday on country and state.
Even if we see only from holiday side that day is public holiday.

That's my opinion

But over all I have tested it's OK.

Thanks

mikevhe18:
:+1:

azmimr67:
has been tested 

:+1:

pedrobaeza:
@feketemihai, I only ask you before I merge the module to clean the commit history. Please squash a bit some commits that are only flake8 fixes or similar, and remove the changes in other modules like hr_employee_firstname, because it's not the same in conflicts terms to add something and then remove it, than not touching it.

feketemihai:
@pedrobaeza Done...

PR #215, opened 2016-04-04
Update split names at module install to get the value not the key of dict

feketemihai:
The same with #194 but moved to a new personal branch.

### DIFF
--- a/hr_employee_firstname/models/hr_employee.py
+++ b/hr_employee_firstname/models/hr_employee.py
@@ -45,10 +45,10 @@ class HrEmployee(models.Model):
             ('firstname', '=', ' '), ('lastname', '=', ' ')])
 
         for ee in employees:
-            lastname, firstname = self.split_name(ee.name)
+            split_name = self.split_name(ee.name)
             ee.write({
-                'firstname': firstname,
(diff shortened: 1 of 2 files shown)


feketemihai:
@dreispt Can you check this one too please...when you go on runbot all the employees have lastname firstname as the name...this should fix it...I push a rebuild on the runbot...

feketemihai:
@dreispt Runbot is ready...

dreispt:
LGTM :+1:
It would be nice to have a test added, though.

feketemihai:
@dreispt Basically this is an update from https://github.com/OCA/hr/pull/190/files, the tests are there, only that the tests are not done at module install, so for create, write everythink is ok...only at install of the module i was not updated in 190...

dreispt:
OK, I see. But that could still be tested:
If you have employee demo data you know to be there before the modules is installed, at the beginning of your tests you could check if the name transformation was done properly by the init scripts.

feketemihai:
@dreispt I will added tomorow...

gurneyalex:
:+1:

moylop260:
:+1:

PR #217, opened 2016-04-11
[ADD] hr_employment_termination

saltonmassally:
@andhit-r another dependencies of hr_infraction

Code is extract from hr_employee_state; reason been I needed the termination workflow but not the over functionality in that module

In a case like this how is authorship and copyright attribution handled?


feketemihai:
@tarzan0820 The copyright should belong to the author, it's just like changing a module name, the code is the same but moved to a new module...so first copyright should belong to Michael Telahun Makonnen, you can add yourself(company) as of 2016...

I agree with adding this module to oca, but i think in this moment it's not so general, for example i don't see the role of a cron here, the termination reasons should belong to the company policy, no need to add predefined data, we should choose which contract i want to finish, if the employee is a consultant too, having two contract and i want him to remain a consultant, i should have the ability in the wizard to choose the contract to finish, the one2many field from employee should match linked model, termination_ids...the termination reason should have also a note where you can specify conditions that should be respected..the same with the grouping...i suggest to try to relate them to the base groups...when you add groups for each module...it will be a lot of work to configure one user...

feketemihai:

Issue #224, opened 2016-05-11
List of deprecated modules and removal from 8.0 and 9.0


feketemihai:
Hello @OCA/human-resources-maintainers ,

beside those 2 module already addresed in the issue roadmap of migration to 8.0 #86 i wanted to have your oppinion about other modules that i think should be deprecated and removed from the 8.0 and 9.0 branches and marked in 7.0 desciption as DEPRECATED.

Please review:
- [ ] hr_accrual - Can be configured from hr_employee_benefit
- [ ] hr_experience_analytic - Add only the analytic account field in hr_experience
- [ ] hr_labour_recruitment - It's doing the same things as hr_recruitment base addons
- [x] addons already removed as aeroo dependants: #63 
- [ ] hr_payroll_extension Should be incorporated in modules regarding policies.

It will be more easy to view which modules are maintained.

ghost:
:+1: I agree with the list.

I think also hr_simplify should be deprecated and removed.

andhit-r:
:+1: @dufresnedavid  i think hr_simplify will be depreceted based on conversation on https://github.com/OCA/hr/pull/97

feketemihai:
I think hr_payroll_period or hr_period should be as well deprecated, on v 7.0 modules hr_period and hr_payroll_period brings kind of the same thinks, i think we should decide which one should remain..
hr_period is already in PR: #145

ghost:
@feketemihai we should definitely keep hr_period and remove hr_payroll_period.

hr_payroll_period is dangerous and its original author does not contribute to the repo anymore.

dreispt:
IMO conclusions here should be updated on the list at #86

feketemihai:
@OCA/human-resources-maintainers 
Can i have your opinion on hr_worked_days_activity module, from my point the module is useless, just add a field on the payslip lines, but since the field added is many2one, i don't see the point, since the payslip is computed on total attendance days, holidays days,  timesheet lines in https://github.com/OCA/hr/tree/8.0/hr_worked_days_from_timesheet.
Plus, it just adds a new model for having leave activities by employee...

Can we drop this module?
https://github.com/OCA/hr/tree/8.0/hr_worked_days_activity

ghost:
@feketemihai drop hr_worked_days_activity

PR #63, opened 2015-01-14
remove addons depending on aeroo

gurneyalex:
closes #62

PR #225, opened 2016-05-12
Remove list of deprecated modules, update readme.

feketemihai:


feketemihai:
Follow up of #224

ghost:
:+1:

nhomar:
Hello.

IS this possible you can in the commit message explain why the modules is deprecated?

Deprecate a module without a single explanation (even if it is obvious) is a bad practice.

After a proper explanation of the deprecation +1 for me.

dreispt:
I think we should have a separate PR for each module removed.
That allows to leave a better trail of why it was removed, and allows people interested in each particular module to present arguments on the corresponding discussion.

feketemihai:
I will create separate PR with my opinion why it should be deprecate, i will create one PR for aeroo report removed in #63 from 7.0

PR #227, opened 2016-06-06
[REM] Remove hr_accrual module.

feketemihai:
Remove of module hr_accrual.

The code inside this module is doing a small part from what hr_employee_benefit module is doing, the benefit module is already in v8, so i think there's no need to keep the accrual module.

@OCA/human-resources-maintainers .


andhit-r:
:+1:

pedrobaeza:
:+1:

ghost:
:+1:

PR #229, opened 2016-06-06
[REM] Remove module hr_labour_recruitment.

feketemihai:
Remove of hr_labour_recuitment module.

Since the module is doing the same as the hr_recruitment module available in Odoo base addons, there is no need to keep and maintain this one. Please see:

https://github.com/odoo/odoo/tree/8.0/addons/hr_recruitment


andhit-r:
:+1:

pedrobaeza:
:+1:

ghost:
:+1:

PR #231, opened 2016-06-06
[REM] Remove hr_holidays_extension module.

feketemihai:
Remove of hr_holidays_extension module.

The module was usefull  to compute holidays based on rest days and public holidays, since this options were added in the module https://github.com/OCA/hr/tree/8.0/hr_holidays_compute_days, i don't think we should keep this one, the detailed view of rest_days and public_holidays_days in holidays it's not neccesary, if the company uses attendance and overtime policies, the calculation of paid days should be added in separate module, e.g hr_payroll_holidays, or to link overtime policy module to add rules on rest days and public holiday days.


ghost:
:+1:

dreispt:
:+1:

elicoidal:
:+1:

PR #233, opened 2016-06-06
[REM] Remove of module hr_employee_education.

feketemihai:
Remove of module hr_employee_education.

The module is just adding a selection field on education of employee, the module hr_experience already has this facility, to add employee curriculum, academic titles, diplomas, so i think it's best to use that one, i don't think we should keep 2 modules doing the same thing.


pedrobaeza:
:+1:

elicoidal:
:+1:

ghost:
:+1:

PR #145, opened 2015-10-03
Migrate hr period from 7.0 to 8.0

saltonmassally:

PR #235, opened 2016-06-06
[REM] Remove of hr_payroll_period module.

feketemihai:
Remove of hr_payroll_period module.

The module is not neccesary because of the new module hr_period in v7 which is ported on v8 on #145 .


ghost:
:+1:

andhit-r:
:+1:

dreispt:
It is unfortunate that two overlapping modules were allowed.
We should take more care so that it doesn't happen again.

feketemihai:
@dreispt I agree on this (there was the hr_payroll_period module when the hr_period was merged, one year ago...)...but it happened so at least for v8 and v9 to maintain only one module...and since this one is in PR's for v8, it think it's better to release the other...

ghost:
@dreispt also, just take a look at the code of hr_payroll_period. It's not maintainable at all.

PR #237, opened 2016-06-16
[FIX] Fix issue with 2 invoices on the same partner and the same total amount

kryskool:
On the same expense, when we have 2 or more lines with differents invoices, and each invoices have the same total amount, reconcile is not possible.

The fix is to exclude reconcile account.move.line, and the first time if we have more than one line to reconcile on the same amount, we keep the first.


ghost:
@kryskool thanks for the contribution. Could you add a unit test to cover this bug?

gurneyalex (approved):

PR #241, opened 2016-07-03
FIX contract view

ovnicraft:
correct way to add the header in the right place.


feketemihai:
@ovnicraft Thanks for the patch, do you still need this one for v7?

ovnicraft:
@feketemihai no, btw i didnt check 10.0

PR #246, opened 2016-07-19
[FIX] Issue:245 ensure_one() error on benefit_rate

### LINKED ISSUE #245: ensure_one() error op benefit_rate
We have installed this module but do get an error after applying a benefit rule when opening a contract or opening employee benefit rates.

error:

>   return new_api(self, _args, *_kwargs)
>   File "/opt/odoo/custom/community_addons/hr_employee_benefit/models/hr_employee_benefit_rate.py", line 110, in get_amount
>     self.ensure_one()
>   File "/opt/odoo/odoo-server/openerp/models.py", line 5303, in ensure_one
>     raise except_orm("ValueError", "Expected singleton: %s" % self)
> except_orm: ('ValueError', 'Expected singleton: hr.employee.benefit.rate(1, 2)')

What is the best way to fix this?

astirpe:
Fixes #245

### DIFF
--- a/hr_employee_benefit/models/hr_employee_benefit_rate.py
+++ b/hr_employee_benefit/models/hr_employee_benefit_rate.py
@@ -106,18 +106,18 @@ class HrEmployeeBenefitRate(models.Model):
         for rate in self:
             rate.employee_amount = rate.get_amount(today)
             rate.employer_amount = rate.get_amount(today, employer=True)
-    
+
     @api.multi
     def get_amount(self, date, employer=False):
-        for rate in self:
-            for line in rate.line_ids:
-                if line.date_start <= date and (
-                    not line.date_end or date <= line.date_end
-                ):
-                    return (
-                        line.employer_amount if employer else
-                        line.employee_amount
-                    )
+        self.ensure_one()
+        for line in self.line_ids:
+            if line.date_start <= date and (
+                not line.date_end or date <= line.date_end
+            ):
+                return (
+                    line.employer_amount if employer else
+                    line.employee_amount
+                )
         return False
 
     @api.model
--- a/hr_employee_benefit/models/hr_employee_benefit_rate.py
+++ b/hr_employee_benefit/models/hr_employee_benefit_rate.py
@@ -100,22 +100,24 @@ class HrEmployeeBenefitRate(models.Model):
             ('annual', _('Annual')),
         ]
 
+    @api.multi
     def _get_amounts_now(self):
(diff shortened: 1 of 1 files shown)


hr_employee_benefit/models/hr_employee_benefit_rate.py
```diff
@@ -100,22 +100,24 @@ def get_all_amount_types(self):
...
-        self.employee_amount = self.get_amount(today)
-        self.employer_amount = self.get_amount(today, employer=True)
-
+        for rate in self:
+            rate.employee_amount = rate.get_amount(today)
+            rate.employer_amount = rate.get_amount(today, employer=True)
+    
     @api.multi
     def get_amount(self, date, employer=False):
-        self.ensure_one()
```
moylop260, line 110: IMHO a `api.multi` with a return isn't the best solution.
You could use `ensure_one` from `get_amount` method because with your change now is used with a singleton from `_get_amounts_now`

astirpe, line 110: @moylop260 Thanks!
This PR was only meant to be a quick bug fix. The fix also contains the removal of the constraint of being a singleton, that is not necessary IMHO. Is there any reason to still keep the _self.ensure_one()_  in the _get_amount()_ method?

I agree, that return isn't the best solution. Such refactoring is not meant to be included in this PR though.

moylop260, line 110: I think with the change of this pr in the method `_get_amounts_now` is enough to fix the error.
Because the problem is we had a:

``` bash
@api.multi
def _get_amounts_now(self):
      self.employee_amount = self.get_amount(today)  #  <-  Error here using a singleton method with a non-singleton `self`
```

With your fix 

``` bash
@api.multi
def _get_amounts_now(self):
     for record in self:
          record.employee_amount = record.get_amount(today)  #  <-  Good. Using a singleton method with a singleton `record`
```

moylop260:
Could you add a small test to reproduce the error and check that the fix works, please?

hr_employee_benefit/models/hr_employee_benefit_rate.py
ghost, line 110: @astirpe the ensure_one is important. You should indeed call the method from a singleton.

dreispt:
:+1:

ghost:
:+1:

ghost:
@astirpe thanks for the contribution

PR #260, opened 2016-10-07
[10.0] [MIG] fix imports

yelizariev:
This PR fixes imports. It doesn't fix\migrate anything else. The purpose of this pull request is to make runbot work. It's necessary step to any further migration. Compare it with current status.
[job_10_test_base.txt](https://github.com/OCA/hr/files/515448/job_10_test_base.txt)

```
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3156072-10-0-e3373c/openerp/addons/auditlog/models/rule.py", line 5, in <module>
    from openerp import models, fields, api, modules, _, SUPERUSER_ID, sql_db
ImportError: cannot import name models

```

Some code is absolutely wrong (e.g. `models.except_orm`, `fields.char` with new `fields`) and has to be fixed later. Please accept the Pull Request.

The same PR can be made for all other OCA modules. Script is here https://gist.github.com/yelizariev/147f5db1a28a76ce707589be8b1af09b


yelizariev:
I have created PRs for depended repos

pedrobaeza:
@yelizariev, we are handling runbot in another way to not need to make all these changes, so better to wait to make these fixes until we see what can we do. For now, we have already Travis stuff correct.

yelizariev:
ok

yelizariev:
Actually, I think we still need these updates. Otherwise developers will face import errors on local installation for development. THe same on deployment -- they will need to delete unported modules from repo dir

yelizariev:
The alternative solution is to clean 10.0 branch and merge ported modules only

pedrobaeza:
We are already using it and non installable modules are not loaded, so we don't detect the problem you mention. Are you working with the latest 10.0 revision?

yelizariev:
It seems, that you are right and non installabed modules are not loaded. I will wait for runbot updates

pedrobaeza:
@gurneyalex can you tell us which is the patch you have made?

pedrobaeza:
@yelizariev you have here the needed changes in runbot: https://github.com/gurneyalex/odoo-extra/commits/OCA-runbot

yelizariev:
Few other things that can be made for whole repo:

```

find . -type f -name '*.xml' | xargs sed -i 's/menu_hr_configuration/menu_human_resources_configuration/g'

find . -type f -name '*.csv'  -o -name '*.py' -o -name '*.xml'  | xargs sed -i 's/base.group_hr/hr.group_hr/g'


```

PR #261, opened 2016-10-07
[10.0][MIG] hr_holidays_compute_days

yelizariev:


yelizariev:
I will fix lint tomorrow

yelizariev:
@gurneyalex @pedrobaeza Why runbot doesn't start for this PR?

pedrobaeza:
There are a lot of pending processes since the branch creation, so it has been even detected. I'm not sure if it will be triggered after all... Let's see if we don't need to restart machine and if needed, forcing push can trigger again the rebuild after the restart.

yelizariev:
@pedrobaeza I guess cron on runbot is stucked. Check this build for example: https://runbot.odoo-community.org/runbot/build/3156163 -- started testing 2 days ago and still haven't switched to job_20

yelizariev:
@gurneyalex 
https://runbot.odoo-community.org/runbot/static/build/3156163-10-0-2ed61c/logs/job_10_test_base.txt

```
 > /srv/openerp/instances/openerp-oca-runbot/sandbox/bin/python: can't open file '/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3156163-10-0-2ed61c/bin/openerp-server.py': [Errno 2] No such file or directory
```

pedrobaeza:
@gurneyalex can you please check runbot to see?

gurneyalex:
I'm looking into this

yelizariev:
@gurneyalex have you applied this patch? https://github.com/odoo/odoo-extra/commit/5b29b651ba3fe3604d1ec93870fc98854a023df4

feketemihai:
@gurneyalex I wanted to rebuild runbot but it's the same, complaining about bin openerp-server.py...strange...

gurneyalex:
@yelizariev @feketemihai can you `git commit --amend`  and `git push -f`? this should fix the build.

gurneyalex (approved):

yelizariev:
@gurneyalex I have done, but it still doesn't work

gurneyalex:
@yelizariev it took one more round of fixing on my side. It should get better soon. I thank you for your patience :smile_cat:

hr_holidays_compute_days/models/hr_holidays.py
```diff
@@ -5,7 +5,6 @@
 from openerp import models, fields, api, _
 from openerp.exceptions import ValidationError
 from dateutil.relativedelta import relativedelta
-import math
 
 
 class HrHolidays(models.Model):
```
feketemihai, line 10: @yelizariev Can you please add an onchange method for holiday_status_id, the default view is the calendar view, when you select dates, since the holiday_status_id is not selected and not even in the context, the method is not executed, after you select dates it will open the form view of the holiday where you need to select the holiday_status_id, but the method of computing days is not triggered at tjis change...so if you will not change dates or employee, the resulted number of days will be wrong.

And please add test for the method.

feketemihai (changes requested):
Otherwise ok, :+1:  after onchange on holiday status .

hr_holidays_compute_days/models/hr_holidays.py
yelizariev: @feketemihai good catch!

yelizariev: done

feketemihai: @yelizariev Thanks for the fast reaction, testes on runbot and now it's ok.

feketemihai (approved):

pedrobaeza:
You need to change README file to put version 10.0 and you should also update it to new the template.

EBII:
you let import in hr_public_holidays.py

from openerp import ... please set as from odoo import ... 

regards

PR #267, opened 2016-10-19
[WIP] 9.0 migration of hr_holidays_legal_leave

gurneyalex:


feketemihai:
@damdam-s @gurneyalex Thank for update the module, can i ask to update readme, since the HR config isn't there anymore with the text from version 10.

Configuration

Company setup \* Go to the company and select the leave type to use as legal leave

with 

Configuration

assign legal leave type via Leaves > Configuration:
choose Leave type you need
click [Edit]
set Legal/Annual checkbox
click [Save]

damdam-s:
@feketemihai no problem. I do it right now

feketemihai:
:+1:

hr_holidays_legal_leave/models/hr_holidays_status.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import fields, models, api
+
+
+class HolidaysType(models.Model):
+    _inherit = "hr.holidays.status"
+
+    is_annual = fields.Boolean(
```
pedrobaeza, line 9: Although I know this is what is now in v10, I'm seeing that this is not comfortable for user with access to several companies. We should have a `company_id` field that is filled by default with user company, but with the possibility to be switched. This is similar to Accounting settings.

damdam-s: @pedrobaeza you're right (I think) but is it correct to have a different behaviour between 9 and 10 versions ?
If you want me to do it, I can but I will make a PR on v10 too to have the same behaviour.

@gurneyalex @feketemihai cc

pedrobaeza: Yes, please. I think it's a good change. We can cherry-pick the commit directly in v10 when done in v9.

damdam-s: @pedrobaeza done in  df9830f. Is it what you wanted ?

pedrobaeza (approved):
Yeah, thanks

damdam-s:
great! thanks

pedrobaeza:
Ouch, it lacks the _company_id_ field in the view. I'll add it directly

damdam-s:
aww sorry for that :(

pedrobaeza:
I have include both commits in v10, cc @yelizariev

PR #269, opened 2016-10-24
[FIX] hr_holidays_legal_leave: error on creating new employee

yelizariev:
File "/mnt/files/git/oca_hr-10/hr_holidays_legal_leave/models/hr_employee.py", line 41, in _inverse_remaining_days
    leave.signal_workflow(sig)
UnboundLocalError: local variable 'leave' referenced before assignment

### DIFF
--- a/hr_holidays_legal_leave/tests/test_holidays_legal_leave.py
+++ b/hr_holidays_legal_leave/tests/test_holidays_legal_leave.py
@@ -52,3 +52,11 @@ class TestHolidaysLegalLeave(common.TransactionCase):
         # let's attempt setting remaining leave
         self.employee.write({'remaining_leaves': 20})
         self.assertEqual(self.employee.remaining_leaves, 20)
+
+    def test_create_employee(self):
+        """Check that we are able to create new employee
+        and that _inverse_remaining_days doesn't raise error"""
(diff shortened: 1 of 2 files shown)


feketemihai:
:+1:  Tested and it's ok, employee is created. 

@yelizariev  Can you fix line length, please.

yelizariev:
done

hr_holidays_legal_leave/tests/test_holidays_legal_leave.py
```diff
@@ -52,3 +52,11 @@ def test_setting_remaining(self):
         # let's attempt setting remaining leave
         self.employee.write({'remaining_leaves': 20})
         self.assertEqual(self.employee.remaining_leaves, 20)
+
+    def test_create_employee(self):
+        '''Check that we are able to create new employee
```
pedrobaeza: docstring should be put with double quote `"` instead of simple according PEP257

yelizariev: you are right

ilmir-k:
Tested, the error has been fixed :+1:

pedrobaeza (approved):

PR #271, opened 2016-10-28
[MIG] hr_holidays_leave_auto_approve: Migrated to 10.0

espo-tony:
Migration of the module 'hr_holidays_leave_auto_approve' from V9 to V10.
This module allows the user to define a leave type in order to make the system automatically approving all the leave requests (and leave allocation requests) belonging to that leave type.

### DIFF
--- a/hr_holidays_leave_auto_approve/README.rst
+++ b/hr_holidays_leave_auto_approve/README.rst
@@ -1,0 +1,45 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: https://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===================
+Auto Approve Leaves
+===================
+
+This module allows the user to define a leave type in order to make the system
+automatically approving all the leave requests (and leave allocation requests)
+belonging to that leave type.
+
+Configuration
+=============
+
+If you wish that the system automatically approves all the leave requests
+belonging to a specific leave type, please follow the steps below.
+
+#. Go on the leave type configuration menu
+#. Select the leave type you wish to setup
+#. Mark the flag 'Auto Approve'.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Andrea Stirpe <a.stirpe@onestein.nl>
+* Antonio Esposito <a.esposito@onestein.nl>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Please respect commit history following https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0

hr_holidays_leave_auto_approve/README.rst
```diff
@@ -0,0 +1,48 @@
...
+Get sure that the administrative user belongs at least to the group
+Holidays/Officer.
+
+If you wish that the system automatically approves all the leave requests
+belonging to a specific leave type, please follow the steps below.
+
+#. Go on the leave type configuration menu
+#. Select the leave type you wish to setup
+#. Mark the flag 'Auto Approve'.
+
```
feketemihai, line 25: @espo-tony Can you make readme complaint, usage with Try me on runbot , credits icon, plus Bug Tracker section...

feketemihai (commented):
@espo-tony As an improvement to the module, can you make or at least add in readme a roadmap info,  which should hide the approve button if the selected holiday status is marked as auto approve...

hr_holidays_leave_auto_approve/README.rst
espo-tony: @feketemihai , 
Thank you for reviewing this module. 
In my latest commit I added the missing informations to the readme file.
As for the approve button, I don't think it's a good idea to hide it when the selected status is marked as auto approve, as it would make the request impossible to approve in the following two cases:
- The request was performed, but not approved, before that the module was installed (or after that the module was installed but before that the holiday status was marked as auto approve);
- The administrative user (in contravention with what is suggested in the configuration guide), at the moment of the request creation, doesn't belong to the  Holidays/Officer group.

In both cases, as the functionality is intended to work only during the request creation, the approval will never be automatically triggered and, if we hide the button, also the manual approval won't be possible.
To avoid these issues, and because I don't see any drawback in showing the approval button for auto-approval marked requests, I'd prefer to not add this feature. What's your opinion about that?

feketemihai: @espo-tony This was my fear in the PR for 9.0, but it was merge beside taking this in consideration, https://github.com/OCA/hr/pull/258 , it was about launching the validation on write method...but we can leave like this to keep the same behaviour...Thanks for the changes.

feketemihai (approved):

pedrobaeza:
What about the commit history? It hasn't been respected

espo-tony:
@pedrobaeza 
Sorry, it seems I made a mistake somewhere during the process, as indeed I didn't follow the steps that you linked above. In future I will surely take it in account, but for this PR do you have any suggestion about how to fix it?
Thank you in advance.

pedrobaeza:
My suggestions are that you start from a clean 10.0 branch, make the steps, and then, copy your final code over the resulting branch, and make a new commit with these changes. Finally, push that branch over this one and you will get everything correct.

espo-tony:
@pedrobaeza ,
Is this correct?

pedrobaeza (approved):
Yeah, it seems so. Thanks!

PR #272, opened 2016-10-28
[MIG] 10.0 Porting hr_employee_id

astirpe:
Please notice that I considerably changed the configuration part in order to be compatible with Odoo V10.


astirpe:
Rebased in order to re-launch runbot after https://github.com/OCA/hr/pull/274

hr_employee_id/models/res_company.py
```diff
@@ -25,5 +25,5 @@ class ResCompany(models.Model):
     employee_id_sequence = fields.Many2one(
         comodel_name='ir.sequence',
         string='Sequence',
-        help="Pattern to be used for used for ID Generation",
+        help="Pattern to be used for used for ID Generation"
```
feketemihai: @astirpe Hi Andrea, can you please fix the help message with better description: Pattern to be used for employee ID Generation".

hr_employee_id/models/res_company.py
```diff
@@ -25,5 +25,5 @@ class ResCompany(models.Model):
     employee_id_sequence = fields.Many2one(
         comodel_name='ir.sequence',
         string='Sequence',
```
feketemihai: @astirpe Can you update the strign as well: Employee ID Sequence.

hr_employee_id/models/res_config.py
```diff
@@ -3,53 +3,50 @@
...
+        default=_default_id_gen_method
     )
     employee_id_random_digits = fields.Integer(
-        string='# of Digits',
-        default=5,
-        help="Number of digits making up the ID"
+        related='company_id.employee_id_random_digits',
+        default=_default_id_random_digits
     )
     employee_id_sequence = fields.Many2one(
```
feketemihai, line 49: @astirpe Can you add the model in the field definition...will be more clear...

feketemihai (commented):
@astirpe @OCA/human-resources-maintainers Andrea, is it too much to ask you to make a method on Apply button to generate Identification Number for existing employees, if they don't have it generated?

feketemihai (approved):

feketemihai:
@astirpe Thanks for your quick reaction. :+1:

astirpe:
@feketemihai it's possible and easy to add a method on the Apply button. On the other hand, in the past I had a similar experience with a customer: that didn't want to generate the number for existing employees by button; instead they asked to make the field editable in case no value is set already. That solution was better suitable for them, as they were able to fill the field for existing employees with values already defined.

Maybe we could leave this point out of scope for this PR? I could add this point in the readme for the time being, and track the discussion on  a separate PR. What do you think?

Thank you!

feketemihai:
@astirpe For me it's ok to have this in readme, to know that it will not generate for existing ones...

espo-tony:
:+1:

astirpe:
Rebased branch in order to re-launch runbot

pedrobaeza (approved):

Yopli2k:
Hi, testing v10, especially the new kiosk mode which adds a new field called "barcode". This field has the same meaning as the field ID added by this module. Have you considered using "barcode" field instead of adding a new field?

PR #276, opened 2016-10-28
[MIG] hr_employee_firstname: Migrated to 10.0

espo-tony:


LeartS (approved):
There is a superflous indentation level in the xml file left after the removal of the `<data>` tag, but ok for me.

njeudy (approved):
- Module code look good
- Module install correctly on 10 database and do the job.

hr_employee_firstname/models/hr_employee.py
```diff
@@ -34,10 +33,11 @@ def split_name(self, name):
         clean_name = u" ".join(name.split(None)) if name else name
         return self.env['res.partner']._get_inverse_name(clean_name)
 
-    @api.cr_context
-    def _auto_init(self, cr, context=None):
-        super(HrEmployee, self)._auto_init(cr, context=context)
-        self._update_employee_names(cr, SUPERUSER_ID, context=context)
+    @api.model_cr_context
+    def _auto_init(self):
```
pedrobaeza: This should be done in a post_init_hook

hr_employee_firstname/__init__.py
```diff
@@ -1,23 +1 @@
-# -*- coding: utf-8 -*-
```
pedrobaeza, line 1: Don't remove encoding

hr_employee_firstname/i18n/hr_employee_firstname.pot
```diff
@@ -4,7 +4,7 @@
 #
```
pedrobaeza: Remove this file, as it's not needed and confuses when outdated.

hr_employee_firstname/tests/__init__.py
```diff
@@ -1 +1,22 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Use short header: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

pedrobaeza (changes requested):

njeudy:
@pedrobaeza is there a way to commit on the same merge request ? or @espo-tony should accept me in espo-tony:10_porting_hr_employee_firstname ?

LeartS:
You can do a PR against the branch of this PR in @espo-tony's repo, when he merges it this PR will automatically update

hr_employee_firstname/models/hr_employee.py
espo-tony: I totally agree.
I didn't move this functionality into a post_init_hook because in the previous versions of this module _auto_init is still used. As this is just a PR for migration I thought better to keep the continuity with the previous versions. I think that changing this should be rather done in a separate pool request.
What's your opinion about this?

pedrobaeza: I think this should be done right here. Migration means sometimes a refactorization, and something that is not correct at all (this is not multi-DB aware) should be cut from the beginning.

espo-tony:
@pedrobaeza 
In my last commit I've implemented the post_init_hook as requested.
For the runbot failure do you have any clue on what could be the problem? It doesn't seem to depend on my commit, nor on Odoo at all.

pedrobaeza:
Yeah, runbot is failing due to the change in the infrastructure to Docker. It seems there's still something to fine-tune, @moylop260 and @gurneyalex...

gurneyalex:
No idea... I just triggered a rebuild to see

pedrobaeza:
It's also happening on l10n-spain: runbot.odoo-community.org/runbot/build/3274108

gurneyalex:
could be related to https://github.com/OCA/runbot-addons/issues/111

hr_employee_firstname/init_hook.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import SUPERUSER_ID
+from openerp.api import Environment
```
leemannd: s/openerp/odoo/

hr_employee_firstname/views/hr_view.xml
```diff
@@ -1,27 +1,28 @@
...
-
-            <label for="category_ids" position="before">
+            <xpath expr="//label[@for='name']" position="replace">
                 <label for="firstname" class="oe_edit_only"/>
+            </xpath>
+            <xpath expr="//field[@name='name']" position="attributes">
+                <attribute name="invisible">1</attribute>
+                <attribute name="no_label">1</attribute>
+                <attribute name="required">0</attribute>
+            </xpath>
```
leemannd, line 15: IMHO I find this way to do to be more [readable](https://github.com/OCA/hr/blob/9.0/hr_employee_firstname/views/hr_view.xml#L16) . Just a question, is there any reason to have taken V8 of the module instead of the V9 ?

leemannd (changes requested):
One dependencie change & a question about the version

leemannd:
@pedrobaeza Post init_hook have been updated. Is it possible to have a review? Thank you.
@espo-tony Thanks for the port.

leemannd (approved):

hr_employee_firstname/views/hr_view.xml
pedrobaeza: It's better to not make replace for avoiding losing other attributes (for example, a custom context), so I prefer this one.

pedrobaeza:
Please shorten headers as stated and I'll merge.

pedrobaeza (approved):

espo-tony:
@leemannd Thank you very much for your contribution.

PR #280, opened 2016-10-30
[ADD] new module hr_payroll_total_ytd

bassn:
Add the year-to-date total on the payslip.
This module is a fork:
- We fork module https://github.com/savoirfairelinux/odoo_payroll/tree/9.0/payroll_amount_ytd
- We upgrade it to Odoo 10.0
- We remove all dependencies and make it usable with Odoo core module hr_payroll
- This module sums up the field total while the forked module sums up the field amount


pedrobaeza:
Don't open another PR, but make the operations in the same branch.

PR #290, opened 2016-11-01
Move public holidays menu under Leaves menu.

feketemihai:


astirpe:
I agree with this PR. Thank you @feketemihai 
:+1:

jcdrubay:
+1

LeartS (approved):
Definitely agree, the current positioning in the "Employee" menu is quite weird.

pedrobaeza (approved):
A change in dependencies is not usually good, but as this is at the very beginning of the version cycle and it's very tight to the dependent module, I agree

feketemihai:
@pedrobaeza I agree with you Pedro, but the thing is that it should depend in hr_holidays module, to align the menus...

pedrobaeza:
OK, another reason more then!

PR #291, opened 2016-11-04
10.0 public holidays next year

jcdrubay:
Allow the creation of public holidays for next year through a wizard.

Minor odoolint fixes

Note: Recent merged PR#290 has been taken into account.

### DIFF
--- a/hr_public_holidays/wizards/public_holidays_next_year_wizard.py
+++ b/hr_public_holidays/wizards/public_holidays_next_year_wizard.py
@@ -2,7 +2,6 @@
 # Copyright 2016 Trobz
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from datetime import timedelta
 import logging
 
 from odoo import api, models, fields
@@ -56,7 +55,7 @@ class PublicHolidaysNextYearWizard(models.TransientModel):
                 last_ph_line.copy(new_ph_line_vals)
 
         domain = [['id', 'in', new_ph_ids]]
-        print domain
+
         action = {
             'type': 'ir.actions.act_window',
             'name': 'New public holidays',
--- a/hr_public_holidays/wizards/public_holidays_next_year_wizard.py
+++ b/hr_public_holidays/wizards/public_holidays_next_year_wizard.py
@@ -44,9 +44,6 @@ class PublicHolidaysNextYearWizard(models.TransientModel):
 
         for ph in pholidays:
 
-            # curr_country = ph.country_id # and ph.country_id.id or NO_COUNTRY
-
-            # last_ph_country = last_ph_dict.get(curr_country, False)
             last_ph_country = last_ph_dict.get(ph.country_id, False)
 
             if last_ph_country:
--- a/hr_public_holidays/wizards/public_holidays_next_year_wizard.py
+++ b/hr_public_holidays/wizards/public_holidays_next_year_wizard.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Trobz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from datetime import timedelta
+import logging
+
+from odoo import api, models, fields
+
+
+_logger = logging.getLogger(__name__)
+
+
+class PublicHolidaysNextYearWizard(models.TransientModel):
+    _name = 'public.holidays.next.year.wizard'
+    _description = 'Creates public holidays for the year following the last'
+
+    @api.multi
+    def create_next_year_public_holidays(self):
+
+        ph_env = self.env['hr.holidays.public']
+        pholidays = ph_env.search([])
+
+        NO_COUNTRY = 'NO_COUNTRY'
+        last_ph_dict = {}
+        for ph in pholidays:
+
+            curr_country = ph.country_id and ph.country_id.id or NO_COUNTRY
+            last_ph_country = last_ph_dict.get(curr_country, False)
+
+            if last_ph_country:
+                if last_ph_country.year < ph.year:
+                    last_ph_dict[curr_country] = ph
+            else:
+                last_ph_dict[curr_country] = ph
+
+        new_ph_ids = []
+        for last_ph in last_ph_dict.itervalues():
+
+            new_ph_vals = {
+                'year': last_ph.year + 1,
+            }
+            new_ph = last_ph.copy(new_ph_vals)
+
+            new_ph_ids.append(new_ph.id)
+
+            for last_ph_line in last_ph.line_ids:
+
+                ph_line_date = fields.Date.from_string(last_ph_line.date)
+                new_date = ph_line_date.replace(ph_line_date.year + 1)
+
+                new_ph_line_vals = {
+                    'date': new_date,
+                    'year_id': new_ph.id,
+                }
+                last_ph_line.copy(new_ph_line_vals)
+
+        domain = [['id', 'in', new_ph_ids]]
+        print domain
+        action = {
+            'type': 'ir.actions.act_window',
+            'name': 'New public holidays',
+            'view_mode': 'tree,form',
+            'res_model': 'hr.holidays.public',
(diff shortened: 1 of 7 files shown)


hr_public_holidays/wizards/public_holidays_next_year_wizard.xml
```diff
@@ -0,0 +1,35 @@
...
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+<record id="public_holidays_next_year_wizard_view" model="ir.ui.view">
+    <field name="name">Create Next Year Public Holidays</field>
+    <field name="model">public.holidays.next.year.wizard</field>
+    <field name="arch" type="xml">
+        <form string="Create Next Year Public Holidays">
+            <footer>
```
LeartS, line 51: I would add a little description here that briefly explains what the wizard does. The empty wizard looks, well, empty

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
```diff
@@ -0,0 +1,67 @@
...
+
+class PublicHolidaysNextYearWizard(models.TransientModel):
+    _name = 'public.holidays.next.year.wizard'
+    _description = 'Creates public holidays for the year following the last'
+
+    @api.multi
+    def create_next_year_public_holidays(self):
+
+        ph_env = self.env['hr.holidays.public']
+        pholidays = ph_env.search([])
```
LeartS: why not directly `pholidays = self.env['hr.holidays.public'].search([])`, as you never use `ph_env` again?

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
```diff
@@ -0,0 +1,67 @@
...
+
+        NO_COUNTRY = 'NO_COUNTRY'
+        last_ph_dict = {}
+        for ph in pholidays:
+
+            curr_country = ph.country_id and ph.country_id.id or NO_COUNTRY
```
LeartS: idiomatic python: `curr_country = ph.country_id.id if ph.country_id else NO_COUNTRY`
But you could even put directly the country recordset as the dict keys  and skip the check altoghether. This is possible because [recordset implements the __hash__ method](https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L5157).
`last_ph_country = last_ph_dict.get(ph.country_id, False)`

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @api.multi
+    def create_next_year_public_holidays(self):
+
+        ph_env = self.env['hr.holidays.public']
+        pholidays = ph_env.search([])
+
+        NO_COUNTRY = 'NO_COUNTRY'
+        last_ph_dict = {}
+        for ph in pholidays:
```
LeartS, line 45: Maybe a warning/error if no holidays defined?

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
```diff
@@ -0,0 +1,67 @@
...
+                'year': last_ph.year + 1,
+            }
+            new_ph = last_ph.copy(new_ph_vals)
+
+            new_ph_ids.append(new_ph.id)
+
+            for last_ph_line in last_ph.line_ids:
+
+                ph_line_date = fields.Date.from_string(last_ph_line.date)
+                new_date = ph_line_date.replace(ph_line_date.year + 1)
```
LeartS: Very rare to have a public holiday in that day, but if the date is February 29 of a leap year, you'll get `ValueError: day is out of range for month` when doing this

LeartS (commented):
I think it would be useful to let the user select which `hr.public.holiday` to replicate, defaulting to the most recent one for each country as it is now; and also let the user choose the year (defaulting to the next).
For example if 2016 was a special year with an additional public holiday for the 150th anniversary of the Italian unification, I'd want to replicate the 2015 Italian holidays to 2017.

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
jcdrubay: Mainly because of an habit.
But in next commit, it will make it easier to use later if necessary. In the next commit, it will also simplify a bit to avoid code on 2 lines for this:

```
        ph_env = self.env['hr.holidays.public']
        pholidays = self.template_ids or ph_env.search([])
```

Also, I think it will mean less changes if we need to re-use it later on.

If required, I can change, let me know.

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
jcdrubay: Nice trick! :)

hr_public_holidays/wizards/public_holidays_next_year_wizard.py
jcdrubay: Three cases this year.
image

Source: http://www.qppstudio.net/monthlyholidays/bank-holidays-february-2016.htm

Handling this rare case will mean quite a lot of complexity because previous or next day might also be a public holiday.

I suggest to block the user with a UserError: "You cannot use a the template of a year that includes public holidays on 29th of February (2016, 2020...), please select a template from another year.". Since I take into account your first comment, the users will be able to select the public holidays from 2015 instead for 2017; I guess it is good enough.

jcdrubay:
Thanks @LeartS for your great feedback. I tried to take them into account in the best way. I am a bit scared so about the template and year field on the wizard as they create a lot of complexity for the users. I tried to make it clear on the wizard page with some documentation...

Please let me know if you have any additional suggestion.

hr_public_holidays/wizards/public_holidays_next_year_wizard.xml
```diff
@@ -0,0 +1,70 @@
...
+<odoo>
+
+<record id="public_holidays_next_year_wizard_view" model="ir.ui.view">
+    <field name="name">Create Next Year Public Holidays</field>
+    <field name="model">public.holidays.next.year.wizard</field>
+    <field name="arch" type="xml">
+        <form string="Create Next Year Public Holidays">
+            <sheet>
+                <div>
+                        Use this wizard to create pulic holidays based on the
```
LeartS: s/pulic/public

hr_public_holidays/wizards/public_holidays_next_year_wizard.xml
```diff
@@ -0,0 +1,70 @@
...
+                        "Create".<br/>
+                        By default, the most recent public holidays 
+                        for each country are used as template to create  
+                        public holidays for the year following the templates.
+                        </div> 
+                    </group>
+                    <group string="Optional">
+                        <div colspan="2">
+                        The below optional fields are here only to handle 
+                        special situtaions like "2016 was a special year with
```
LeartS: - s/situtaions/situations
- (The 150th anniversary of Italian unification was actually in 2011, I used 2016 in my example just to make it more current 😁  )
- I would put this group under the first one instead of next to it, it looks clumsy:
  screenshot from 2016-11-07 10-04-54
  Or we could even maybe put the two groups in two notebook pages so that the Optional section is "hidden" unless one explicitly clicks on it; while the "Defaults" explanation is immediately displayed. This could make the wizard look simpler and not overload the user with options/fields, what do you think?

LeartS (approved):

hr_public_holidays/wizards/public_holidays_next_year_wizard.xml
jcdrubay: I like your notebook idea :+1:

feketemihai:
@jcdrubay Thanks for the contribution, good module...

feketemihai (approved):

gurneyalex (approved):

jcdrubay:
Yeah! 3 approvals reached :tada:  Thank you everyone :+1:

PR #293, opened 2016-11-16
[MIG] 10.0 Porting hr_employee_age

Yopli2k:


hr_employee_age/models/hr_employee.py
```diff
@@ -17,10 +17,9 @@
 #    along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 ###############################################################################
-from datetime import datetime
 
-from openerp import fields, models, api
-from openerp.tools import DEFAULT_SERVER_DATE_FORMAT as OE_DFORMAT
+from openerp import api, fields, models
```
pedrobaeza: Use `from odoo` instead of `from openerp`, as now in version 10 is the default namespace.

hr_employee_age/models/hr_employee.py
```diff
@@ -33,9 +32,11 @@ class HrEmployee(models.Model):
     )
 
     @api.one
```
pedrobaeza: `@api.one` is deprecated now (they have decided in this way although here is useful), so better to remove in favor of `@api.multi` and a loop with `for record in self:` and the rest of the stuff replacing `self` by `record`.

hr_employee_age/models/hr_employee.py
```diff
@@ -33,9 +32,11 @@ class HrEmployee(models.Model):
...
+    @api.depends('birthday')
     def _compute_age(self):
         if self.birthday:
-            dBday = datetime.strptime(self.birthday, OE_DFORMAT).date()
-            dToday = datetime.now().date()
-            self.age = dToday.year - dBday.year - ((
-                dToday.month, dToday.day) < (dBday.month, dBday.day))
+            dBday = fields.Date.from_string(self.birthday)
```
pedrobaeza: Don't use upper letters in local variable names (that's something wrong also in the previous version). See PEP8 https://www.python.org/dev/peps/pep-0008/#function-names

hr_employee_age/models/hr_employee.py
```diff
@@ -33,9 +32,11 @@ class HrEmployee(models.Model):
...
+            dToday = fields.Date.from_string(fields.Date.today())
+            self.age = relativedelta(dToday, dBday).years
```
pedrobaeza: You can do directly the substraction of one date minus the other and you will get a relativedelta object.

pedrobaeza (changes requested):
Little changes.

hr_employee_age/models/hr_employee.py
Yopli2k: ok, i use the same variables as they were. I'm aiming for future changes. Thank you

Yopli2k:
Ok, all changes are now made to the new commit.

hr_employee_age/models/hr_employee.py
```diff
@@ -17,25 +17,27 @@
...
+    @api.depends('birthday')
     def _compute_age(self):
-        if self.birthday:
-            dBday = datetime.strptime(self.birthday, OE_DFORMAT).date()
-            dToday = datetime.now().date()
-            self.age = dToday.year - dBday.year - ((
-                dToday.month, dToday.day) < (dBday.month, dBday.day))
+        for record in self:
+            if record.birthday:
+                record.age = relativedelta(
```
pedrobaeza, line 39: The idea here is to do:

```
record.age = (fields.Date.from_string(fields.Date.today()) - fields.Date.from_string(record.birthday)).years
```

and you save also an import.

Yopli2k: The code you propose gives me the following error:

`File "~/Odoo10/hr/hr_employee_age/models/hr_employee.py", line 39, in _compute_age
    record.age = (fields.Date.from_string(fields.Date.today()) - fields.Date.from_string(record.birthday)).years
**AttributeError: 'datetime.timedelta' object has no attribute 'years'**`

pedrobaeza: Right, the returned object is not the same. Well, you can get it with a division, but let's keep this way then

pedrobaeza (approved):

rafaelbn:
Rebuilding for testing

rafaelbn (approved):
Tested in runbot

PR #298, opened 2016-11-25
[MIG] Make module installable and update version 9 to version 10 in version reference

jlaloux:
[MIG] Make module installable and update version 9 to version 10 in version reference. Module works fine. Just need to make it installable


jlaloux:
A PR is already done for this. Haven't seen it

PR #304, opened 2016-12-05
fix is_annual computation remove id

leemannd:


leemannd:
@pedrobaeza @damdam-s We weren't comparing the right things.

pedrobaeza (approved):
Yeah, fast-tracking it!

PR #307, opened 2017-01-03
[ADD] resource_calendar_rrule

hbrunn:
RRules in resource calendars
============================

This module was written to extend the functionality of resource calendars to support more flexible schedules than the default implementation. Per calendar, you can create rules describing the schedule in a very similar manner to what most calendaring applications use for repeating appointments (technically: iCal recurrence rules, short rrule).

Usage
=====

To use this module, you need to:

- go to a resource's calendar
- instead of filling in a day of week, fill in the repetition rule

Known issues / Roadmap
======================

* implement the rrule version of all functions in ``resource.calendar``
* maybe also divert from current standard a bit because the current code assumes weekly repetition at many places
* an init hook to convert existing calendars to rrules would be nice


feketemihai:
@hbrunn Thanks for the contribution, nice module, but i have some concerns, since i made some tests on runbot http://3280920-307-491798.runbot1.odoo-community.org/web?debug=#page=0&limit=80&view_type=list&model=calendar.event&menu_id=155&action=158

Can you please check the calendars that strech over periods when Daylight Saving Time changes...will end up 1 hour before or after the time set on the initial calendar event.

hbrunn:
you might want to elaborate what values you fill in, what the computation results are, and what you expect instead

feketemihai:
@hbrunn Can you enter on runbot, on the calendar please see "test hours - Daylight Saving Time", check the hour of the recorded recurrencies, the hour changes from 09 to 10 and inverse at each change of Daylight Saving Time...the starting hour i think it should remain the same....

feketemihai:
@hbrunn If hours is set at recurency, to check if the time si with Daylight Saving Time ( .tm_isdst) and to add the value get from the comparotion (0 or 1)...

resource_calendar_rrule/models/resource_calendar_attendance.py
```diff
@@ -0,0 +1,78 @@
...
+            text = ''
+            for start, end in this._iter_rrule(
+                    datetime.datetime.now(),
+                    datetime.datetime.now() + datetime.timedelta(days=7),
+            ):
+                text += '%s %s - %s\n' % (
+                    fields.Datetime.context_timestamp(this, start)
+                    .strftime(language.date_format),
+                    fields.Datetime.context_timestamp(this, start)
+                    .strftime(language.time_format),
```
andhit-r, line 49: @hbrunn is it posible to use strftime("%H:%M") instead of language.time_format? The preview field always show different second depend on current time. It may confuse end user.

rrule1

rrule2

andhit-r:
@raditv you might want to check out this module.

hbrunn:
I'll have to think about the summer time issue a bit more. In many other applications, we want the intervals to be stable, so we can't always convert to the local timezones. And for organizations spanning multiple time zones it's always a recipe for pain and suffering to introduce local time zones. I think what we should do is to add a parameter to https://github.com/OCA/server-tools/tree/8.0/field_rrule that developers can use to choose between equal intervals (current behavior) or equal times (wanted behavior here). Patches are welcome!

raditv:
@andhit-r thank you, I'll check it out

feketemihai (approved):
@hbrunn Since the daylight issue will be fixed on the dependant module, it's fine with me.

feketemihai:
@hbrunn Thanks for the update, i have some issues saving resource calendar after changing the morning and afternoon hours, please see error below, probably will be raised on calculation of afternoon hour_to...

/home/odoo/build/OCA/hr/resource_calendar_rrule/models/resource_calendar.py", line 156, in _attendance_from_simplified
    'hour_to': morning_hour + day['morning'],
TypeError: coercing to Unicode: need string or buffer, int found

hbrunn:
you've been too fast, I just now added the integration of the change in server tools. And tests, that revealed you exact problem, which is fixed now

feketemihai:
@hbrunn Still having error in _iter_rule, when changing hours, timezone on Europe/Brussels, 
crop_schedule
it says that hours must be between 0 and 23...

hbrunn:
well, you fill in times, but what you're supposed to fill in here are the *amount of hours* the person works in the morning resp afternoon. The *times* as to when exactly someone begins cannot be set in the simplified widget, that's taken globally from a config parameter. If https://github.com/OCA/hr/pull/307/files#diff-56eb8d60c705c3bd94949566281871e5R19 doesn't make this sufficiently clear, please PR an alternative

feketemihai:
@hbrunn You're right, thanks for quick reaction...and sorry for disturb...

andhit-r (approved):
@hbrun I'm not sure to use simplified view as default. I almost made the same mistake like @feketemihai did. IMO, Information on the simplified-view have a high probability to mislead user. But it's non-blocking since it can be controlled by group.

So :+1:

pedrobaeza:
Please revert the oca-dependencies.txt change and squash at your taste before I merge the PR.

hbrunn:
will do so, but we can't merge this before https://github.com/OCA/server-tools/pull/912 is merged, and there review didn't even start

pedrobaeza:
OK, I have just looked at the module in the PR, but didn't know that there was a later patch.

pedrobaeza:
Now you can act here.

hbrunn:
I will, give it a few hours

hbrunn:
@pedrobaeza long few ours, but now it's mergable

pedrobaeza:
OK, no problem. Merging.

pedrobaeza (approved):

PR #314, opened 2017-02-22
[MIG] hr_emergency_contact: Migrated to 10.0

leemannd:
- Base port of the module
-  Improvement of the base contact emergency view with new fields

### DIFF
--- a/hr_emergency_contact/README.rst
+++ b/hr_emergency_contact/README.rst
@@ -13,7 +13,7 @@ Installation
 
 To install this module, you need to:
 
-1.  Clone the branch 8.0 of the repository https://github.com/OCA/hr
+1.  Clone the branch 10.0 of the repository https://github.com/OCA/hr
 2.  Add the path to this repository in your configuration (addons-path)
 3.  Update the module list
 4.  Go to menu *Setting -> Modules -> Local Modules*
--- a/hr_emergency_contact/README.rst
+++ b/hr_emergency_contact/README.rst
@@ -31,7 +31,7 @@ To use this module, you need to:
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/116/8.0
+   :target: https://runbot.odoo-community.org/runbot/116/10.0
 
 Bug Tracker
 ===========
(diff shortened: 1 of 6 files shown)


hr_emergency_contact/README.rst
```diff
@@ -13,7 +13,7 @@ Installation
```
feketemihai, line 13: @leemannd Please add yourself in the Contributors list...

leemannd: @feketemihai Updated

feketemihai (approved):

hr_emergency_contact/models/hr_employee.py
```diff
@@ -3,20 +3,17 @@
...
 
 
 class HrEmployee(models.Model):
 
     _inherit = 'hr.employee'
 
     emergency_contact_ids = fields.Many2many(
-        string='Emergency Contacts',
         comodel_name='res.partner',
-        relation='rel_employee_emergency_contact',
```
andhit-r: @leemannd if you remove this parameter is it possible there will be an error if user migrate from previous version?

leemannd: @andhit-r Hum, good question. Frankly I can't tell as I don't have enough experience on migrations. But I feel like it's an old fashioned way of describing the relation and we should let the ORM do the job.

andhit-r: Neither do i :smile: @feketemihai do you have opinion regarding this matter?

Well that's my old fashioned line of code :smile: . CMIIW, The reason i use it because on really-really rare occasion the creation of relationship table will raise an exception if you have 2 many2many fields on same object that have same comodel_name. Take a look at code on how odoo name the table https://github.com/odoo/odoo/blob/9.0/openerp/osv/fields.py#L947.

feketemihai: @leemannd @andhit-r without any migration script, the link will be lost without specifying the m2m table name, so it's better to keep it in the definition...

leemannd: OK. Let's keep it.

leemannd:
Updated from @andhit-r comments & removed or updated copyright in files

andhit-r (approved):
:+1:

gurneyalex (approved):

PR #324, opened 2017-03-08
fixed name clash introduced by new api inheritance.

veroc:
_compute_number_of_days clashed with super model method that shouldn't be overridden.


gurneyalex:
fixed in #351

PR #329, opened 2017-03-30
[9.0] [MIG] hr_worked_days_from_timesheet

LoisRForgeFlow:
Migration to v9:
* Rename openerp tags.
* Bump version and update manifest.
* Adapt code to v9. (new analytic account relation with timesheets).
* Add a method to check if the contract is set in the payslip.
* Adapt tests to v9.
* Update README.


hr_worked_days_from_timesheet/README.rst
```diff
@@ -2,11 +2,12 @@
...
+=============================
+HR Worked Days From Timesheet
+=============================
 
-* Adds a button to import worked days from timesheet
+This module allow you to automatically fill the 'Worked Days' of payslips with
+the information of each employee's timesheets for the payslip's period.
 
 Installation
 ============
```
AaronHForgeFlow, line 13: I think this is not needed

AaronHForgeFlow (approved):
It works as expected. Also code soft review :+1: 

Thanks @lreficent :1st_place_medal:

feketemihai (approved):

hr_worked_days_from_timesheet/README.rst
```diff
@@ -16,26 +17,23 @@ To install this module, you need to:
 * clone the branch 8.0 of the repository https://github.com/OCA/hr
```
serpentcs-dev1: > clone the branch 8.0 

should be 9.0

serpentcs-dev1 (commented):
Also please update icon to that of the OCA icon and the [Bug Tracker](https://github.com/OCA/hr/pull/329/files#diff-2967c349d2ea47ff0388938474c3902cR43) part of Readme with a newer version of it, by removing the following lines


Otherwise LGTM :+1:

LoisRForgeFlow:
@darshan-serpent I just updated the PR with your changes, can you check?

hr_worked_days_from_timesheet/README.rst
serpentcs-dev1: I think it should be 9.0 here.

serpentcs-dev1 (approved):
One small change Otherwise LGTM :+1:

PR #331, opened 2017-04-13
[9.0][ADD] hr_holiday_notify_employee_manager

LoisRForgeFlow:
HR Holiday Notify Employee Manager
==================================

This module extends the functionality of the Leaves App to make possible
to notify by mail the manager of the employee requesting the leave.

Configuration
=============

#. Go to *Settings > Users > Companies*.
#. In the company form go to the *Configuration* tab and in the *Leaves
   Management* section.
#. Check *Leave Requests notified to employee's manager* box.

Usage
=====

To use this module, you need to:

#. Go to *Leaves > My Leaves > Leaves Request* and create a leave request to be
   approved.
#. Your manager will be notified by email.


JordiBForgeFlow (approved):
:+1: Technical and functional review

hr_holiday_notify_employee_manager/models/hr_holidays.py
```diff
@@ -0,0 +1,34 @@
...
+
+from openerp import api, models
+
+
+class HrHolidays(models.Model):
+    _inherit = 'hr.holidays'
+
+    def _get_approvers_to_notify(self):
+        """Defines who to notify."""
+        company = self.env['res.company']._company_default_get('hr.holidays')
```
jcdrubay: Should we call directly `self.env['res.users']._get_company()`? 

Or without 'hr.holidays' which is ignored anyway `company = self.env['res.company']._company_default_get()`

hr_holiday_notify_employee_manager/models/hr_holidays.py
```diff
@@ -0,0 +1,34 @@
...
+        if company.leave_notify_manager and self.employee_id.parent_id:
```
jcdrubay: Could we notify someone by default? in case we forget to set the manager?
Normally, there is only one case of person without manager and he might not input his leave request...

This could be another setting in the company form.

hr_holiday_notify_employee_manager/models/hr_holidays.py
```diff
@@ -0,0 +1,34 @@
...
+        res._notify_approvers()
+        return res
+
+    @api.one
+    def _notify_approvers(self):
+        """Input: res.user"""
+        approvers = self._get_approvers_to_notify()
+        if not approvers:
+            return True
+        for aprover in approvers:
```
jcdrubay: aprover > approver
:)

jcdrubay (changes requested):
Nice :+1:

hr_holiday_notify_employee_manager/models/hr_holidays.py
LoisRForgeFlow: what do you mean? by default it notifies the followers of the department, which probably include the department's manager. I don't know if that covers your case...  :confused:

hr_holiday_notify_employee_manager/models/hr_holidays.py
LoisRForgeFlow: what about: `self.env.user.company_id`?

jcdrubay: Or even better:
`self.employee_id.company_id`

Because what is needed is the configuration for the company of that employee.

hr_holiday_notify_employee_manager/models/hr_holidays.py
jcdrubay: I did not know that. Yes it does cover my case. Let's just hope that each department as at least one follower.

hr_holiday_notify_employee_manager/models/hr_holidays.py
LoisRForgeFlow: Yes, that's the best

hr_holiday_notify_employee_manager/models/hr_holidays.py
LoisRForgeFlow: Yes, also if you want you can check the first version of this module (79f80950d63c5c4d56a2adb46099388620ceef8f) in which we had more options in who to notify.

jcdrubay (approved):

hr_holiday_notify_employee_manager/models/hr_holidays.py
```diff
@@ -0,0 +1,36 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, models
+
+
+class HrHolidays(models.Model):
+    _inherit = 'hr.holidays'
+
+    def _get_approvers_to_notify(self):
+        """Defines who to notify."""
```
moylop260: Adds `ensure_one` and @api.multi decorator

hr_holiday_notify_employee_manager/models/hr_holidays.py
```diff
@@ -0,0 +1,36 @@
...
+        company = self.employee_id.company_id
+        if company.leave_notify_manager and self.employee_id.parent_id:
+            return self.employee_id.parent_id
+        else:
```
moylop260: [no-else-return](http://eslint.org/docs/rules/no-else-return)
Use directly `return False` without `else`

hr_holiday_notify_employee_manager/models/hr_holidays.py
```diff
@@ -0,0 +1,36 @@
...
+            return False
+
+    @api.model
+    def create(self, vals):
+        res = super(HrHolidays, self).create(vals)
+        res._notify_approvers()
```
moylop260, line 23: Use a `context` in order to avoid notify if you are importing information

moylop260 (changes requested):
Review lints from travis.

E.g. `api.one` is seudo-deprecated: https://github.com/odoo/odoo/blob/a9da6e9f3f73674df3354e5e6d8f354e8b26eecb/openerp/api.py#L405
(Better use api.multi plus ensure_one)

hr_holiday_notify_employee_manager/models/hr_holidays.py
LoisRForgeFlow: can you provide a example on how to do it correctly?

LoisRForgeFlow:
@moylop260 Thanks for your review, I'm attending it.

hr_holiday_notify_employee_manager/README.rst
```diff
@@ -0,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+HR Holiday Notify Employee Manager
+==================================
+
+This module extends the functionality of the Leaves App to make possible
```
elicoidal, line 9: I am wondering the difference with the standard notification/follower system. I would swear that if you are a follower of an employee as manager you receive their leaves/expenses notifications

elicoidal: In any case, maybe you can explicit the difference here with the standard notification system

elicoidal (approved):

hr_holiday_notify_employee_manager/README.rst
LoisRForgeFlow: That is not true. You would become a follower of the employee's leave if you are following his/her department and you would receive a email notification depending on how you've configured yourself as follower of the department. If you are just follower of the employe and/or his/her manager you will neither follow the leave or be notified.

LoisRForgeFlow: This said, a good use case for this module would be a company with the following requirements:
* There are several followers by department.
* Not all of the followers of the department can approve the leaves, but just the ones that are explicitly managers of the employee requesting the leave.

In this case you will have to configure the system in the following way:
* Configure all the followers of the department to not receive email notifications on leaves requests.
* Check the box *Leave Requests notified to employee's manager* added by this module.

LoisRForgeFlow:
Hello @moylop260 

Can you refresh your review  :grimacing:? Thanks!

moylop260 (approved):

moylop260:
Thank you!

PR #335, opened 2017-04-17
[FIX] payslip move date remove unnecessary update

levkar:
Updating move line is unnecessary and triggers affects_move and todo_date in move line update 

https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L1227
https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L1233

which results update of move and all other move lines. 

https://github.com/odoo/odoo/blob/8.0/addons/account/account_move_line.py#L1265

This brings a 5 second payslip confirm to 2 minutes (in our case).  This fix cuts it back to 5 seconds.


andhit-r (approved):
No test

feketemihai (approved):

congdpt (approved):

PR #336, opened 2017-04-24
[ADD][9.0] new module hr_payroll_report

Wizr:
which add a pivot view of model `hr.payslip.line`

### DIFF
--- a/hr_payroll_report/README.rst
+++ b/hr_payroll_report/README.rst
@@ -1,0 +1,52 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+Payroll Report
+==============
+
+This module extends the functionality of Payroll by adding a BI report view.
+
+Installation
+============
+
+To install this module, you need to:
+
+* have basic module `hr_payroll` installed.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/hr/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Xudong Yang <fanderous@gmail.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/hr_payroll_report/README.rst
+++ b/hr_payroll_report/README.rst
@@ -6,7 +6,8 @@
 Payroll Report
 ==============
 
-This module extends the functionality of Payroll by adding a BI report view.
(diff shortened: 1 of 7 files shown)


Wizr:
cc @elicoidal

hr_payroll_report/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Elico Corp (www.elico-corp.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Payroll Report",
+    "version": "9.0.0.1.0",
```
elicoidal: 9.0.1.0.0

hr_payroll_report/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+    "category": "Generic Modules/Human Resources",
```
elicoidal: Add a summary key

hr_payroll_report/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+    "website": "https://odoo-community.org/",
+    "author": "Elico Corp (www.elico-corp.com), "
```
elicoidal: remove " (www.elico-corp.com)"

hr_payroll_report/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Elico Corp (www.elico-corp.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Payroll Report",
+    "version": "9.0.0.1.0",
+    "category": "Generic Modules/Human Resources",
+    "website": "https://odoo-community.org/",
```
elicoidal: https://www.elico-corp.com

hr_payroll_report/README.rst
```diff
@@ -0,0 +1,54 @@
...
+This allows to create all kind of analysis for the payroll department
+
+Installation
+============
+
+To install this module, you need to:
+
+* have basic module `hr_payroll` installed.
+
+Bug Tracker
```
elicoidal: I miss the runbot section

elicoidal (changes requested):
some details

elicoidal (approved):

elicoidal:
@feketemihai @dufresnedavid your review is welcome

elicoidal:
@feketemihai @dufresnedavid is there any interest in this module?

feketemihai:
@elicoidal @Wizr Thanks for the module, it's o aggod one, but i have some remarks to do, since there is a little code in in can you update to new api, second, can we add a field in salary rules to be showned on this report, i know we already have one for showing in to the payslip, but to have a good report we should be filtering the gross and the net lines, regarding the total amount, ot to add in readme this conclusion, so everybody can add a default filter not to show this lines. Regarding the measures i think that they are too much, i think the first 4 it's enough, since you will probably want to make some comparition between payslips, and if you want to detail on one you can add the measures after...

elicoidal:
@feketemihai thanks for your feedback

> since there is a little code in in can you update to new api

:+1: 

> second, can we add a field in salary rules to be showned on this report, i know we already have one for showing in to the payslip, but to have a good report we should be filtering the gross and the net lines, regarding the total amount, ot to add in readme this conclusion, so everybody can add a default filter not to show this lines.

:-1:  I think it is dependent on every HR manager, country etc. With  current solution, we can be flexible without complexity

> Regarding the measures i think that they are too much, i think the first 4 it's enough, since you will probably want to make some comparition between payslips, and if you want to detail on one you can add the measures after...

I'd rather keep them if necessary (it is useful for some analysis) but not display in the table by default. What do you think?

feketemihai:
@elicoidal @Wizr I'm fine with your comments, and to adapt the last one, not to be shown by default all measures, so you have my :+1:  Can you adapt it so i approve the PR?

elicoidal:
@wizr can you please update the code accordingly?

feketemihai (approved):

elicoidal:
I can safely merge

PR #338, opened 2017-04-28
[ADD] new module hr_ppe

oihane:
PPE stands for Personal protective equipment
This new module is for defining a list of products that will be given
to an employee according to the job title.


mslokhat:
Hi
is there any chances to have it for v9 or v10?
Rgds

gurneyalex (approved):

houssine78:
Just wondering if this module if the module could serve a more generic needs as the material given to the employee could be for something else that just serving ppe purpose...

feketemihai (approved):

PR #339, opened 2017-05-02
[10.0] [bugfix] hr_holidays_compute_days: fix computation of number_of_days

gurneyalex:
# issue 1

*    install hr_holidays_compute_days
*    create a 3 day holiday allocation for any employee
*    go to leave summary and check the allocation

*expected:* the duration is 3 days
*actual:* the duration is 0 days

# issue 2:

*    install hr_holidays_compute_days
*   create a 3 day holiday request in the middle of the week for any employee
*    go to leave summary and check the request

*expected duration:* -3 days
*actual duration:* +3 days.

#Analysis 

There is a name collision between `_compute_number_of_days` defined in this addon
and the method for the `field number_of_days` computation in the base addon `hr_holidays`
which causes the latter to not be called => the field `number_of_days` would be 0 for
holidays allocation, and has the wrong sign for holidays request (because it was
wrongly forced in the the reimplemented version.

This patch:
* adds tests for the expected value of number_of_days
* adds a test for the behavior of the method for holidays allocation
* renames the method _compute_number_of_days to _recompute_number_of_days

### DIFF
--- a/hr_holidays_compute_days/__manifest__.py
+++ b/hr_holidays_compute_days/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'Employee Compute Leave Days',
-    'version': '10.0.1.0.0',
+    'version': '10.0.2.0.0',
     'category': 'Human Resources',
     'license': 'AGPL-3',
     'summary': 'Computes the actual leave days '
--- a/hr_holidays_compute_days/models/hr_holidays.py
+++ b/hr_holidays_compute_days/models/hr_holidays.py
@@ -46,7 +46,7 @@ class HrHolidays(models.Model):
                 raise ValidationError(_("You cannot schedule the end date "
                                         "on a public holiday or employee's "
                                         "rest day"))
-            self.number_of_days_temp = self._compute_number_of_days()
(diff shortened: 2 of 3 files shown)


gurneyalex:
To test the issue mentionned above:

issue 1
* install hr_holidays_compute_days
* create a 3 day holiday allocation for any employee
* go to leave summary and check the allocaiton

expected: the duration is 3 days
actual: the duration is 0 days

issue 2: 
* install hr_holidays_compute_days
* create a 3 day holiday request in the middle of the week for any employee
* go to leave summary and check the request
expected duration : -3 days
actual duration : +3 days.

dreispt (approved):
Looks good and Travis is happy.
PS: those codecov stats can't be right :worried:

feketemihai:
@gurneyalex I wasn't able to test on runbot, seems something is not ok because it is always redirecting to the OCB repo, but the code looks fine for me.

feketemihai (approved):

gurneyalex:
@feketemihai the runbot was having weird issues lately which I fixed yesterday. I just triggered a manual rebuild of that PR.

PR #344, opened 2017-06-03
Extract hr_evaluation from odoo/8.0 and make it unistallable

dreispt:
The purpose is to make the original Odoo module available with all the (quite long) history.
The only actual change made was to make it [uninstallable](https://github.com/OCA/hr/pull/344/files#diff-f1b0ee8247b88567acf36ba278a5103d).
The actual migration should go in its own PR.


hbrunn:
do we have a process for this? We're very adamant for new contributors that every commit must come from a person having signed our CLA, but that's not the case with Odoo core code of course. Or do we have an agreement with Odoo SA that we can hijack their code in such cases?

gurneyalex:
@hbrunn we have. 

Ideally we should have a flag saying that the module is not covered 100% by the OCA's IP rights

gurneyalex (approved):

PR #350, opened 2017-07-14
[ADD][10.0] hr_holidays_hour: allow duration of leaves in hours

astirpe:
The standard Odoo application "Leave Management" allows the employees to set leaves duration in days.
By installing this module, the duration of the leaves will be expressed in hours.


astirpe:
Travis error seems not related with this module.

See PR https://github.com/OCA/hr/pull/351

astirpe:
Another travis error that seems not related with this module.
This time the [tests](https://travis-ci.org/OCA/hr/jobs/254744158) of hr_holidays_imposed_days are failing.

hr_holidays_hour/models/hr_holidays.py
```diff
@@ -0,0 +1,163 @@
...
+        work_hours = self._compute_work_hours(from_dt, to_dt)
+        self.number_of_hours_temp = work_hours
+
+    @api.model
+    def _compute_datetime(self, date):
+        dt = False
+        if date:
+            reference_date = fields.Datetime.context_timestamp(
+                self.env.user,
+                datetime(1990, 2, 8, 12)
```
hbrunn: can you explain this constant?

hr_holidays_hour/models/hr_holidays.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+            dt = fields.Datetime.from_string(date)
+            tz_dt = fields.Datetime.context_timestamp(self.env.user, dt)
+            dt = dt + tz_dt.tzinfo._utcoffset
+            dt = dt - reference_date.tzinfo._utcoffset
```
hbrunn, line 52: I take it this function is supposed to localize datetimes while not really localizing them? Do I understand correctly that this code will have a problem in a time zone with daylight saving when one of `date` and `reference_date` is dst and the other not?
What I'd do is something like `date.astimezone($target_tz).replace(tzinfo=None)` - this gives you a naive datetime with local wall time. Consider using `dateutil.tz.gettz` to load the zone from the user's timezone

hbrunn (commented):
(now I need to run away, so only a partial review) What i wanted to ask: Did you consider making this mostly an UI change? So I fill in 10 hours, and the addon just computes 1.25 days from this and the rest remains unchanged? Or did you encounter any problems with that and that's why you did it this way?

astirpe:
@hbrunn, yes we encountered some problems on developing a widget or some kind of logic to be used only in the view.
One of the problems is that, if we only rely on the float representing the days (or partial days), the system could likely round such value at a certain point of the process: as a resulting scenario, it could be that the employee asking for an amount of hours will see a slightly discrepancy with the effective hours stored in the db. I can provide some examples if you want.
Another problem is that the method _get_number_of_days() returns a float that is the result of a division: the resulting decimals can be periodic, and at a certain point the leave hours requested can slightly differ from the days value stored in the db. So we needed to define a _get_number_of_hours() method.

hr_holidays_hour/models/hr_holidays.py
astirpe: We just needed a reference date that for sure occurs in winter time. We will change this part of code, avoiding such hardcoded value.

hr_holidays_hour/models/hr_holidays.py
astirpe: In Odoo the workschedule time is set as a float value, instead of a time value. This is the reason of such strange time-datas manipulations. The issue we experienced with our customers was in this case the requested leaves start in wintertime and end in summertime (and the other way round).

gurneyalex (approved):
This looks strange to a French guy, but it is probably useful if you need this way of counting holidays..

feketemihai:
@astirpe Isn't it easier to update hr_holidays_compute_days, and to add to config, if it's calculated in days or hours, that way we will not have to check which module to install...

astirpe:
@feketemihai My proposal is the other way round.
This module `hr_holidays_hour` is meant to be a base module that simply replaces the days fields with hour fields. This module is meant to be extended by other modules, in order to provide extra functionalities in terms of hours, instead of days.

If we make this module depending on `hr_holidays_compute_days` which also depends on `hr_public_holidays`, we are creating a blob of modules that is probably "too much" for the companies who are just used to track the duration of leaves in hours.

pedrobaeza:
I also don't see this module as needed. Please check #420, that improves a lot the handling of leaves:

- Partial computation of days according working times.
- Better UI selecting by default full days, but letting you to select also hours.

astirpe:
@pedrobaeza you wrote:

> Better UI ... letting you to select also hours.

How can the user enter hours with your module?

pedrobaeza:
You select start and end hours. The system calculates according your working time how many days it supposes.

Imagine you have a day that you have to work 8 hours. If you introduce your leave only for 4 hours, the system it's going to say you that the leave is 0,5 days. And so for other fractions. As leaves are usually granted also by days (23 days for example in Spain), the logical thing is to decrease by the corresponding fraction.

astirpe:
But this module is meant to allow the employees and the HR managers to handle the duration of leaves and allocations in hours, and not in days or fraction of days.

Let me try to explain better.

It is for who doesn't want to read "`Day(s): 0,5`" but wants to read "`Hours: 04:00`". It's only for them.

All our customers, plus the company I'm working for, plus the company I was previously working for, are used to track the leaves requests and the allocation requests in one way: hours. 
What they want is to replace in every form, tree, calendar, report, etc.. the unit of measure for the duration, from days to hours. In short: they don't want to see "`Day(s): 0,5`" at all. They want to see "`Hours: 04:00`". I got a lot of requests like this, I would say from 100% of the customers that use the Odoo HR.

Is it just a mindset? Yes, it is. The information at the end is same, whether expressed in hours or fraction of days. But if we don't provide the label "Hours", they would probably continue to track hours in their Excel sheet. Or for a new potential customer who is used to write hours, we can do a more convincing smooth demo.

In some countries (eg.: in Italy) the allocation requests are expressed in terms of hours.
The consumed leaves are also tracked in hours (look at an italian payslip for example). Not in fractions of days.

Is module `hr_holidays_compute_days` calculating the duration based on working hours? Yes, nice. But the user will continue to read `Days: 0,125` instead of `Hours: 1:00`.
This `hr_holidays_hour` module is also calculating the duration based on working hours, but the user will read the resulting duration in terms of hours.

We may do not want to install `hr_holidays_compute_days` nor `hr_public_holidays` for this.
If there is a public holiday, the HR officer can simply set it in the proper tab of the employee's Working Schedule, without the need of having `hr_public_holidays` installed.

Regarding the request to make the choice of Days-Hours configurable, I think it's also not needed. The companies who want to track the duration in hours will not switch to days in the future. We install this module since the beginning and they will track the hours from now on. They will never ask to convert hours to fractions of days.

pedrobaeza:
OK, I see it a bit strange, and a lot of changes to make, but if it's your use case...

yung-wang (approved):
Working fine. Hours are correctly taken in the holidays. approved

thomaspaulb:
@astirpe Any chance this one already exists for 8.0? Or is it better to backport?

astirpe:
@thomaspaulb I just published [here](https://github.com/onesteinbv/addons-onestein/tree/8.0/hr_leave_hours) the module we used in V8. The functionality is exactly the same.

The name of the V8 module is different: in this PR we renamed the module to be compliant with the OCA standards. Moreover the code of V8 is old and far from being compliant to the OCA standards of quality.

I hope that module can be of help for you.

thomaspaulb:
@astirpe Thanks so much, probably a backport of the 10.0 one would be better but at least now we have the choice!

hr_holidays_hour/models/resource_calendar.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+        return work_intervals
+
+    @api.multi
+    def get_working_intervals_of_day(
+            self, start_dt=None, end_dt=None,
+            leaves=None, compute_leaves=False, resource_id=None,
+            default_interval=None):
+        """ Override method because of the working intervals not
+        calculating the minutes only the hours"""
```
hbrunn: can you elaborate this? Glancing through super it looks like they do handle minutes (and seconds) there: https://github.com/OCA/OCB/blob/10.0/addons/resource/models/resource.py#L334-L335 - or am I missing something?

hr_holidays_hour/models/resource_calendar.py
```diff
@@ -0,0 +1,144 @@
...
+        for interval in work_intervals:
+            work_interval = self.with_context(
+                change_tz=True
+            ).interval_remove_leaves(interval, leaves)
+            intervals += work_interval
+
+        return intervals
+
+    @api.model
+    def interval_remove_leaves(self, interval, leave_intervals):
```
hbrunn: if we need this, I think there's something wrong with UTC handling at the place the produces the leaves

hr_holidays_hour/tests/test_leave_hours.py
```diff
@@ -0,0 +1,333 @@
+# -*- coding: utf-8 -*-
```
hbrunn, line 1: I think we also should repeat `resource_calendar`'s tests here to be sure what we produced is a drop-in replacement

hbrunn (changes requested):
I found a few things to nag about, but I'm already busy backporting this to 8.0, I might stumble upon other stuff as this goes. PR for the backport is coming soon, I'll rebase then as this is merged

hr_holidays_hour/models/resource_calendar.py
astirpe: @hbrunn yes, I guess you're right. The Odoo code you highlighted was [fixed](https://github.com/OCA/OCB/commit/1dea8a53fbe5d064ea6aeae6506df71be00412f5) in March 2017. Our module was [ported](https://github.com/onesteinbv/addons-onestein/commit/89e13d334b282341fadcc036c9d831c97e935da6) to V10 in September 2016. So the fix in Odoo obsoleted the `def get_working_intervals_of_day()` method here. Makes sense.

Just removing the `def get_working_intervals_of_day()` method from this module should be fine.

Thank you!

hr_holidays_hour/tests/test_leave_hours.py
astirpe: I just copy-pasted the tests from the 'resource' module, but I guess there is an elegant way for repeating those tests avoiding the copy-paste...

hr_holidays_hour/models/resource_calendar.py
astirpe: Method removed

hr_holidays_hour/tests/test_leave_hours.py
hbrunn: just import the test class in your test file. Unittest just runs all classes starting with `Test` within the module, also if you just import it

astirpe:
All the tests of this module passed.
The failure of the tests seems related to module `hr_holidays_validity_date`.
I will open a separate PR containing the fixes for all the pylint/flake8 checks that are failing (modules `hr_contract_default_trial_length` and `hr_expense_sequence`).

hbrunn (approved):

hbrunn:
if you plan to do the extra pr anyways, just add the commit here and squash your other ones a bit, then we can just merge in one go. I to 'rebase and merge' anyways, so the result won't be different

astirpe:
I just opened the following PRs:

- [x] https://github.com/OCA/hr/pull/447
- [x] https://github.com/OCA/hr/pull/448
- [ ] https://github.com/OCA/hr/pull/449

@hbrunn @gurneyalex @tafaRU may I ask you an help to speed up the approval process?

astirpe:
@hbrunn thanks! I cherry-picked the commits.

hbrunn:
(fyi I'm still commited to merge this, I just wait for the CI)

astirpe:
@hbrunn tests of `hr_holidays_validity_date` are not yet ok. Is not something depending on this PR though

astirpe:
Strange. Here tests are passing. Instead in https://github.com/OCA/hr/pull/449 they fail. I don't understand.

astirpe:
Thank you all!

astirpe:
Porting hr_holidays_hour to V11 #450

PR #355, opened 2017-08-04
[10.0] [ADD] hr_payroll_cancel

nikul-serpentcs:
HR Payroll Cancel
=================

This module allows the user to cancel a payslip whatever the previous state is without doing a refund. When the user cancels the journal entry is deleted and  the payslip state is set to reject. Then the user is able to set the state to draft again and later on he/she is able to confirm again the payslip.

If there’s a refund for a payslip the user should not cancel the entry because the refund would still be confirmed. In that case, the user has either to confirm again the payslip or cancel the refund.


nikul-serpentcs:
@aheficent Could you please review.

AaronHForgeFlow (changes requested):
Cannot cancel a payslip when is confirmed:

image

AaronHForgeFlow (changes requested):
You should cancel the accounting entry created:
image

Besides, when there's a refund the user should not cancel the payslip. That was the behavior in v9.

AaronHForgeFlow (changes requested):
I still can cancel a payslip when there's a refund. I should not be able.

nikul-serpentcs:
@aheficent Could you check raise error if refund payslip is cancel.

hr_payroll_cancel/models/hr_payroll.py
```diff
@@ -0,0 +1,19 @@
...
+from odoo import api, models, _
+from odoo.exceptions import ValidationError
+
+
+class HrPayslip(models.Model):
+    _inherit = 'hr.payslip'
+
+    @api.multi
+    def action_payslip_cancel(self):
+        if self.credit_note:
```
AaronHForgeFlow: The credit_note flag is activated in the refund not in the original payslip. I am still able to cancel a payslip that can be refunded.

AaronHForgeFlow (changes requested):

nikul-serpentcs:
@aheficent Could you please review.

nikul-serpentcs:
@aheficent Could you please review.

hr_payroll_cancel/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "name": "Hr Payroll Cancel",
+    "version": "10.0.1.0.0",
+    "author": "Vauxoo, Eficent, "
+              "Serpent Consulting Services Pvt. Ltd.,"
+              "Odoo Community Association (OCA)",
+    "category": "HR",
+    "website": "http://www.vauxoo.com/",
+    "license": "AGPL-3",
+    "depends": [
+        "hr_payroll",
```
feketemihai: I think it should depend on hr_payroll_account, since it depends only on hr_payroll no account_move is created, so you don't have to cancel anything...

feketemihai:
@nikul-serpentcs The module is ok, only i have a remark, if i hit by mistake the Refund button, i cannot cancel none of the payslips, so as a procedure i can cancel first the refund and after the initial payslip, because i don't want to see the balances in the accounts...

AaronHForgeFlow:
@feketemihai I agree, the user may want to cancel the refund and then the payslip. But the user should not be able to cancel a payslip before canceling the refund. @nikul-serpentcs can you do the thing?

aheficent:
@feketemihai I agree with you. The user should be able to cancel the refund first and then the payslip but not the other way around. @nikul-serpentcs can you do the thing?

nikul-serpentcs:
@aheficent @feketemihai Improved code Could you please review.

feketemihai (changes requested):

nikul-serpentcs:
@aheficent @feketemihai Improved dependence Could you please review.

AaronHForgeFlow (commented):
It seems that the related journal entry is not cancelled as well, can you check?

nikul-serpentcs:
@aheficent Could you please review.

hr_payroll_cancel/models/hr_payroll.py
```diff
@@ -27,6 +27,9 @@ def refund_sheet(self):
...
     @api.multi
     def action_payslip_cancel(self):
-        if self.refunded_id and self.refunded_id.state != 'cancel':
-            raise ValidationError(_("Cannot cancel a Refunded payslip."))
-        return self.write({'state': 'cancel'})
+        for payslip in self:
+            if payslip.refunded_id and payslip.refunded_id.state != 'cancel':
+                raise ValidationError(_("""To cancel the Original Payslip the
+                    Refunded Payslip needed to be canceled first!"""))
+            payslip.move_id.button_cancel()
```
feketemihai, line 34: @nikul-serpentcs Can you please unlink the move, when it's cancelled all the move lines are deleted but the move remains as unposted, when you revalidate the payslip a new move is creeated, so i tend to unlink the move after it is canceled. Otherwise, work ok .

feketemihai (changes requested):

nikul-serpentcs:
@feketemihai @aheficent The latest code unlink the moves from the canceled payslip.
Could you please review it ??

nikul-serpentcs:
Fixes leave overlap issues. #425

feketemihai (approved):

AaronHForgeFlow (commented):
There's still an issue here:
I created a payslip and a refund. Then I cancelled the refund. Then tried to cancel the original payslip but got the following:

image

Showing the warning is fine IMO, but the message is misleading because the refund is already cancelled. IMO it should be, "please delete first the refund" or something similar.

AaronHForgeFlow (approved):
Please dismiss my last review. I was setting the refund to draft, not just cancelling it.  In that case it's correct to show a warning. Thus, :+1:

AaronHForgeFlow:
Please fix the commit history before this is merged. 73 commits looks incorrect

nikul-serpentcs:
@aheficent Fixed commit history, Can you Please check it ??

AaronHForgeFlow:
@nikul-serpentcs IMO history is not fixed. This module is merged in v9 and all the history of that branch is missed here.

nikul-serpentcs:
@aheficent Fixed commit history, Can you Please check it ??

AaronHForgeFlow:
@nikul-serpentcs You're getting closer, the problem now it that your changes are missing. If you see commit 57b3c479fb93520bca6e875c7316163969646db0 your changes for the migration to v10 are missing

AaronHForgeFlow:
I think you got it!  :medal_sports:

PR #357, opened 2017-08-07
[add] hr_employee_address_improved

simahawk:
Excerpt from README:

When the HR modules are enabled and we have 1 employee for each users,
some parts of the UI become confusing because
we have several partners for the same person.

Assumption: we would like to exclude home addresses
from the list of available partners in m2o and m2m fields.

/cc @gurneyalex


hr_employee_address_improved/hooks.py
```diff
@@ -0,0 +1,22 @@
...
+from odoo import SUPERUSER_ID, api
+import logging
+
+logger = logging.getLogger('[hr_employee_address_improved.post_init_hook]')
+
+
+def post_init_hook(cr, pool):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    partners_count = env['res.partner'].search_count([])
+    empls = env['hr.employee'].search([('address_home_id', '!=', False)])
```
gurneyalex: can you add "active_test": False to the context of this search?

gurneyalex (approved):
one tiny bitsy improvement suggested, but overall :+1:

leemannd (approved):

feketemihai (approved):

PR #362, opened 2017-08-22
[10.0][MIG] hr_holidays_notify_employee_manager

MiquelRForgeFlow:
This PR includes:
- Migration of hr_holiday_notify_employee_manager module to v10.0:
  - It is renamed as hr_holidays_notify_employee_manager to follow the naming conventions.
  - It depends now from the new module hr_holiday_settings.
- The new module hr_holiday_settings, to enable configurable settings in the Leaves menu.


LoisRForgeFlow (approved):
Code review :+1:  Great job! Thanks!

congdpt (approved):

congdpt:
I'm a newbie with Odoo! I saw your commit files.
But I don't understand 1 point. Could you tell me about them?
- Why we have model `hr.holidays.config.settings` with 2 field `company_id` (m2o) and `leave_notify_manager` (related field)
- Why don't we just use field in `res.company`
If we dont' use that model. I think source become shorter and cleaner. Maybe we can merge 2 module
It's my think. Tks!

congdpt (commented):
I'm a newbie with Odoo! I saw your commit files and module in branch 9.0
But I don't understand 1 point. Could you tell me about them?

Why we have model hr.holidays.config.settings with 2 field company_id (m2o) and leave_notify_manager (related field)
Why don't we just use field in res.company
If we dont' use that model. I think source become shorter and cleaner. Maybe we can merge 2 module
It's my think. Tks!

MiquelRForgeFlow:
Because it could be useful for other people to use the `hr.holidays.config.settings` module if they have a new module that needs to put some field in the settings of the Leaves menu.

congdpt:
Tks for your answer @mreficent ! When I check module. I make a problem in `hr.holidays.config.settings`.
I check True for `leave_notify_manager` field and Apply.
After that, I check `leave_notify_manager` field in Company, it's True.
But I back view in `hr.holidays.config.settings`, `leave_notify_manager` field is False

May be you can refer: https://github.com/odoo/odoo/blob/10.0/addons/account/models/res_config.py#L156

MiquelRForgeFlow:
@congdpt Thanks, it's fixed now!

PR #365, opened 2017-09-06
[ADD] New module hr_attendance_schedule

Jerther:
The base hr_attendance module tracks time with second precision. We needed to round attendances to a schedule so we created this. We were surprised that such addons did not exist already, knowing many industries with attendance systems that work that way.

Also, the time shown in kiosk mode was that of the client computer, which is often offset by several minutes compared to the server's. This was no big deal with second precision, but with roundings we had to make sure the time shown was correct.

I thought about using the web/web_timepicker widget but I didn't since it's in another repository and I didn't know if OCA would like that.


Jerther:
Regarding `Model hr_attendance_schedule.clock has no table` in the travis build:
This is intended. I filed a PR for this https://github.com/odoo/odoo/pull/17386

hr_attendance_schedule/models/hr_attendance_clock.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+import pytz
+from datetime import datetime
+
+from odoo import api, fields, models
+from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
+
+
+class AttendanceClock(models.Model):
```
hbrunn: why not inherit from AbstractModel?

hbrunn (commented):
you duplicate a bit of functionality of [resource calendars](https://github.com/OCA/OCB/blob/10.0/addons/resource/models/resource.py#L17) here. Is this intentional? If yes, I'd be curious about your reasoning. Having to administer both seems quite inconvenient to me.

And yes, those calendars are a bit lacking, I'm working hard on a [module](https://github.com/OCA/hr/pull/307) to introduce a bit more flexibility here.

hbrunn:
oh, and concerning modules from other repos: As that's an OCA repo too, yes, please, we love that. And given no other module depends on this repo, you'll have to add it to [oca_dependencies.txt](https://github.com/OCA/hr/blob/10.0/oca_dependencies.txt)

Jerther:
@hbrunn I was not aware of resource calendars. I'll check them out and see if they could replace my schedule class.

hr_attendance_schedule/models/hr_attendance_clock.py
Jerther: Because get_system_clock is called from the JS. Is that feasible with AbstractModel?

hbrunn: absolutely. And the way I understand `_auto`, it's meant to say: "I'll take care of creating a table myself": https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L2337

hbrunn: ([reports](https://github.com/odoo/odoo/blob/10.0/addons/account/report/account_invoice_report.py) where you create a view manually instead of having the ORM do stuff is a very widespread application of this)

Jerther: I thought abstract classes could not be used unless instanciated via a derived class, or using class methods, right? Then again, that was in Java and C# but Python's concept of OOP and inheritance feels a bit loose and permissive to me. Is there an official doc about models.AbstractModel?

hbrunn: models.py, linked above

Jerther:
@hbrunn After revieweing resource calendars, they could be used for this module once extended with grace periods. One thing though, my version uses a duration rather than an end time to facilitate calculations and allow > 24h shifts. The latter I've never seen though, so reusing resource calendars could be done.

Also, we wanted the attendance module not to interfere with other modules like cost analysis and project management. So, where are the resource calendars used? Are they just a reference or dothey have a functionnal purpose?

Jerther:
Here. Now it's using `resource.calendar`. I even added support for `date_from` and `date_to` on attendances, and I also added support for leaves.

I must admit I like it better this way.

hr_attendance_schedule/models/resource_calendar_attendance.py
```diff
@@ -0,0 +1,63 @@
...
+        self._init_datetimes()
+        if self._is_now_allowed(employee_id):
+            matching_date = self._get_matching_date(in_out)
+            if matching_date:
+                offset = matching_date - (self._start_datetime if in_out == 'in' else self._end_datetime)
+                return self._now - offset, False
+        return None, True
+
+    def _init_datetimes(self):
+        self._now = datetime.strptime(fields.Datetime.now(), DEFAULT_SERVER_DATETIME_FORMAT)
```
hbrunn: how is this different from `datetime.datetime.utcnow()`?

And if you ever really need to parse a datetime in an odoo context, use https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L1541

hr_attendance_schedule/models/resource_calendar_attendance.py
```diff
@@ -0,0 +1,63 @@
...
+        d_day = self._get_next_weekday(int(self.dayofweek))
+        self._start_datetime = d_day + timedelta(hours=self.hour_from)
+        self._end_datetime = d_day + timedelta(hours=self.hour_to)
+        if self._end_datetime <= self._start_datetime:
+            self._end_datetime += timedelta(days=1)
+
+    def _is_now_allowed(self, employee_id):
+        parse_date = lambda d: datetime.strptime(d, DEFAULT_SERVER_DATE_FORMAT).date()
```
hbrunn: this absolutely is a case for the above

hr_attendance_schedule/models/resource_calendar_leaves.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+from odoo import models
+from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
+
+
+class ResourceCalendarLeaves(models.Model):
+    _inherit = 'resource.calendar.leaves'
+
+    def matches(self, datetime):
+        date_from = datetime.strptime(self.date_from, DEFAULT_SERVER_DATETIME_FORMAT)
```
hbrunn: see above

hr_attendance_schedule/models/resource_calendar_attendance.py
hbrunn: I'm a bit concerned about the instance member assignments. Any reason not to pass those values through the call stack in the context?

hr_attendance_schedule/models/resource_calendar_attendance.py
```diff
@@ -0,0 +1,63 @@
...
+        if in_out == 'in':
+            lower = self._start_datetime - timedelta(minutes=self.grace_before_checkin)
+            upper = self._start_datetime + timedelta(minutes=self.grace_after_checkin)
+        else:
+            lower = self._end_datetime - timedelta(minutes=self.grace_before_checkout)
+            upper = self._end_datetime + timedelta(minutes=self.grace_after_checkout)
+        return lower, upper
+
+    def _get_next_weekday(self, weekday):
+        monday = datetime(2017, 1, 2)
```
hbrunn: what's the rationale here?

hr_attendance_schedule/static/src/js/kiosk_confirm.js
```diff
@@ -0,0 +1,33 @@
...
+
+  var Model = require('web.Model');
+  var KioskConfirm = require('hr_attendance.kiosk_confirm');
+
+  KioskConfirm.include({
+    start_clock: function() {
+      var self = this;
+      this.$('.o_hr_attendance_clock').text("...");
+      var server_clock = new Model('hr_attendance_schedule.clock');
+      server_clock.call('get_system_clock')
```
hbrunn: it's a good habit to return the deferred object(s) of asynchronous calls made in functions, so that other modules can attach to them

hr_attendance_schedule/static/src/js/kiosk_mode.js
```diff
@@ -0,0 +1,33 @@
...
+
+  var Model = require('web.Model');
+  var KioskMode = require('hr_attendance.kiosk_mode');
+
+  KioskMode.include({
+    start_clock: function() {
+      var self = this;
+      this.$('.o_hr_attendance_clock').text("...");
+      var server_clock = new Model('hr_attendance_schedule.clock');
+      server_clock.call('get_system_clock')
```
hbrunn: see above

hr_attendance_schedule/models/hr_attendance_clock.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
+
+
+class AttendanceClock(models.AbstractModel):
+    _name = 'hr_attendance_schedule.clock'
+
+    @api.model
+    def get_system_clock(self):
+        now = datetime.strptime(fields.Datetime.now(), DEFAULT_SERVER_DATETIME_FORMAT)
+        return pytz.utc.localize(now).isoformat()
```
hbrunn: this is the use case for https://github.com/odoo/odoo/blob/10.0/odoo/fields.py#L1514

note that `pytz.utc.localize(now)` is basically a no-op, and that your code will return tomorrow if you sit in Amsterdam and it's after 23.00. That's what context_timestamp is meant to fix

hbrunn (changes requested):
nice! A couple of comments and [travis'](https://travis-ci.org/OCA/hr/jobs/275937809#L709) nags

hr_attendance_schedule/models/resource_calendar_attendance.py
Jerther: It's so I can mock the current date and time: https://github.com/OCA/hr/pull/365/files#diff-b98e1f7cba7765407b9bab8dc4403ab4R151 

Nice!

You mean like `self.env.context['_now'] = xxx`? If so, that's interresting; what are the advantages?

hr_attendance_schedule/models/resource_calendar_attendance.py
Jerther: I needed a monday. Just any monday. Is it worth a clarification?

hr_attendance_schedule/models/hr_attendance_clock.py
Jerther: Hmm, not exactly. The goal of the method is to return the current UTC date and time, but including the UTC time zone info at the end of the iso string. No user conversion needed.

`now.isoformat() = '2017-01-01T12:12:12'`
`pytz.utc.localize(now).isoformat() = '2017-01-01T12:12:12+00:00'`

I could also have just returned `now.isoformat() + '+00:00'` but coded this way feld more explicit. I see I may have been wrong.

hr_attendance_schedule/models/resource_calendar_attendance.py
hbrunn: the advantage is that you don't introduce race conditions: All threads (not processes) share the same registry, and if you start relying on values of an instance, you might read a value that some other thread wrote. Very theoretical to nearly impossible, but still worth looking into imho.

hr_attendance_schedule/models/resource_calendar_attendance.py
hbrunn: yes, please. You're writing for yourself in 10 years, then this will puzzle you

hr_attendance_schedule/models/resource_calendar_attendance.py
Jerther: The context is a frozendict. So I need to make a copy, update it, then use it with self.with_env... That's getting messy. I feel I'm missing something. Do you have a good and clean example to illustrate how to do this properly?

hbrunn: `.with_context` is a quite standard pattern in Odoo to pass on stuff through the call stack. You're right, the copying is part of this process, and under the hood this actually causes Odoo to read your record again, because the results can differ per context (we just had a nice conversation about that in https://github.com/OCA/server-tools/pull/969). That indeed might be needlessly inefficient in your case. I pass around some mutable type if I want to split code in functions that shares a set of variables. Usually, I use `namedtuple`s as something like a `struct` equivalent in Python: https://github.com/OCA/server-tools/pull/988/files#diff-f8576cd00057f1477095100637a6bf7aR14

Jerther: I did some modifications and found that my code runs in the order of 50% slower when I use .with_context().

Implying I modified the code so attendance_data is actually passed around as a method argument and I never actually fetch `context['attendance_data']`, consider this:

```
    def get_action_date(self, in_out, employee_id):
        attendance_data = self._create_attendance_data(in_out, employee_id)
        return self._get_action_date(attendance_data)

    def _create_attendance_data(self, in_out, employee_id):
        AttendanceData = namedtuple('AttendanceData', 'in_out employee_id now start_datetime end_datetime')
        now = fields.Datetime.from_string(fields.Datetime.now())
        d_day = self._create_datetime_from_dayofweek()
        start_datetime = d_day + timedelta(hours=self.hour_from)
        end_datetime = d_day + timedelta(hours=self.hour_to)
        if end_datetime <= start_datetime:
            end_datetime += timedelta(days=1)
        return AttendanceData(in_out, employee_id, now, start_datetime, end_datetime)
```
If I call `self._get_action_date` this way instead:

```
self.with_context(attendance_data=attendance_data)._get_action_date(attendance_data)
```

Code runs 50% slower.

Also, if I use `with_context(attendance_data={})`, it does not slow down. So it has something to do with the somewhat heavy attendance_data object. This is no big deal here, but there's no doubt it will impact my future developments. Any thoughts?

Jerther:
I seriously thought about the instance variables problem and came up with a couple of half-solutions that would either be too slow or move the same problem elsewhere. Then I came up with this, which is inspired by _Clean Code_'s solution to Active Records containing business logic. I usually take the advice lightly when it comes to Odoo but I found it well suited for this particular case.

hbrunn:
this is the complicated version of passing your data through the call stack. I'm fine with that, but note that given you also moved the code into your object, you'll make it harder for other modules to modify it.

Jerther:
Good point about inheritance. I think the cleanest solution is what you said at the beginning: use the context. And I'd like to do that for all my future devs but do you have any idea why with_context() is so slow? And what could be done to make it faster? I tried to profile the whole thing but profiling tools for Python on Linux aren't quite the commercial tools I'm used to...

feketemihai:
@Jerther Any news on this, good module but needs to be finished...

Jerther:
@feketemihai I believe it's working fine as is. There is some test data that's conflicting with hr_holidays though so the builds fail. I'll look into that when I get some time. Thanks for the ping.

feketemihai:
@Jerther thanks for the update.

Jerther:
Alright, there was a conflict with hr_timesheet_attendance. Got that fixed.

Anything else, @hbrunn ?

PR #369, opened 2017-09-28
add hr accident at work 

EBII:


chienandalu (changes requested):
Why not to integrate this with the already existing leaves functionality?

gurneyalex (changes requested):
I like the idea. 

Can you rename your branch to 10.0-accident_at_work (the prefix is used by runbot to find the odoo version). 

Then there are some flake8 issues: https://travis-ci.org/OCA/hr/jobs/280882487#L713

EBII:
mv 10.0 mig hr holidays validity date #370

EBII:
mv [ADD] 10.0 accident at work #371

PR #371, opened 2017-10-03
[ADD] 10.0 accident at work

EBII:
follow accident at work


chienandalu (commented):
Shouldn't this be integrated with hr leaves?

EBII:
Probably, wich repo or module you speak about ?  please link

chienandalu:
Odoo base HR Holidays let you configure such leaves 
 as sick leaves https://github.com/odoo/odoo/tree/10.0/addons/hr_holidays

EBII:
i don't use hr_leaves
but i have accident @ work to follow. 
regards

feketemihai:
@EBII Hi there, i think what you did is a good module, but first we should have a general module regarding the SAFETY WORK, to define procedures, and accidents are basically some consequences of not respecting the procedures, so i think you should rethink the module to allow first part and link accidents with the procedures the employee didn't respect.

EBII:
for my users list number of accident by operating unit for follow AT rate 
may be i 'll add generate leaves if module is install

feketemihai:
@EBII I still tend to have first the procedures and accidents as consequences, any from @OCA/human-resources-maintainers have a different opinion?

dreispt:
@feketemihai We should be pragmatic. I agree that we can have a link to procedures, but that is not blocking and can be added later. So, not a reason to block this.
If you are interested in this, I suggest connecting HR with OCA/management-system features.

hr_accident_at_work/README.rst
```diff
@@ -0,0 +1,40 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+========================
+HR collective agreement
```
dreispt: Incorrect title?

hr_accident_at_work/__manifest__.py
```diff
@@ -0,0 +1,33 @@
...
+#   Copyright (C) 2017 EBII (http://www.saaslys.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': "HR Accident at work",
+
+    'description': """
+        HR Accident at work for OCA""",
+
+    'description': """
```
dreispt: Please remove the description key.

hr_accident_at_work/models/hr_accident_at_work.py
```diff
@@ -0,0 +1,59 @@
...
+    employee_id = fields.Many2one(comodel_name='hr.employee', required=True,
+                                  string="Employee")
+    accident_address = fields.Text(string='Accident Adress')
+    accident_date = fields.Datetime(string='Date', required=True)
+    accident_sequence = fields.Selection(selection=[
+        ('EXCEPTIONAL FREQUENCY', 'EXCEPTIONAL FREQUENCY'),
+        ('REGULAR FREQUENCY', 'REGULAR FREQUENCY'),
+    ], string='Frequency Type')
+    accident_department = fields.Integer(string='Department', required=True)
+    accident_location = fields.Char(string='City')
```
dreispt: Why label is "City"? I expected the more generic "Location"

hr_accident_at_work/models/hr_accident_at_work.py
```diff
@@ -0,0 +1,59 @@
...
+    number_of_employee_concerned = fields.Integer(
+        string='Number of Employee Concerned', default=1, required=True)
```
dreispt: Do you mean "Number of employees involved"?

hr_accident_at_work/models/hr_accident_at_work.py
```diff
@@ -0,0 +1,59 @@
...
+    number_of_employee_concerned_acknowledge_by_social_security = \
+        fields.Integer(string='Number of Employee concerned and acknowledge by'
+                              ' the Social Security Authority', default=0)
+    social_security_authority_decision = fields.Text(
+        string='Social Security Authority Decision')
+    accident_almost = fields.Boolean(string='Almost accident ?', default=False)
+    accident_during_commute = fields.Boolean(string='Accident during the '
+                                                    'commuting ?',
+                                             default=False)
+    accident_circumstances = fields.Char(string='circumstances', required=True)
```
dreispt: Text type instead of Char?

hr_accident_at_work/models/hr_accident_at_work.py
```diff
@@ -0,0 +1,59 @@
...
+    wound_description = fields.Text(string='Wound description', required=True)
+    online_wound_declaration_done = fields.Boolean(
+        string='Online Wound Declaration Done?', default=False)
```
dreispt: Why specifically "online"? This may not be a requirement, and sounds too specific to your case.

hr_accident_at_work/views/hr_accident_at_work.xml
```diff
@@ -0,0 +1,71 @@
...
+        <field name="res_model">hr.accident.at.work</field>
+        <field name="view_type">form</field>
+        <!--<field name="view_mode">tree,form,graph</field>-->
+    </record>
+
+    <menuitem id="menu_action_accident_at_work"
+              action="action_hr_accident_at_work_tree"
+              parent="hr.menu_hr_main"
+              sequence="5"/>
+</odoo>
```
dreispt: Add line break.

dreispt:
Is "Accident at Work" more correct than "Work Accident"?

EBII:
both expressions are used

PR #377, opened 2017-10-09
[9.0] hr_public_holidays allow duplicate of year

Martronic-SA:
As it is now, this is not possible to duplicate a year that makes it boring to creates new years.
That PR allows to easily create new years by making the copy of the year and related lines plus one year.
It allows also to deletion the created year if needed.

This will have to be forwarded to 10.0 and 11.0

### DIFF
--- a/hr_public_holidays/models/hr_public_holidays.py
+++ b/hr_public_holidays/models/hr_public_holidays.py
@@ -4,6 +4,7 @@
 
 from datetime import date
 from openerp import fields, models, api
+from openerp.tools.translate import _
 from openerp.exceptions import Warning as UserError
 from dateutil.relativedelta import relativedelta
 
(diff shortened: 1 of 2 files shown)


hr_public_holidays/models/hr_public_holidays.py
```diff
@@ -28,6 +29,7 @@ class HrPublicHolidays(models.Model):
         'hr.holidays.public.line',
         'year_id',
         'Holiday Dates'
+        copy=True
```
gurneyalex: missing comma on line 31

gurneyalex (changes requested):

hr_public_holidays/models/hr_public_holidays.py
Martronic-SA: oh yes, thank you!

Martronic-SA:
The travis errors are not coming from my PR but underlying in the base public holidays module.
I just want to merge the copy part of my PR, I don't want to review the whole module.
What can I do?

feketemihai:
@Martronic-SA Can you please fix travis errors, even that is not coming from your PR, tere are just 2 small details. Thanks.

feketemihai:
@Martronic-SA I changed the file online to pass flake and pylint...hope to be ok...

feketemihai:
@Martronic-SA Can you add a simple test for copy...otherwise :+1:

feketemihai (changes requested):
Please add some tests

PR #389, opened 2017-10-18
[imp] hr_employee_id: ease override w/ custom ID

simahawk:
If you need a difference ID in particular cases you can pass a custom value for identification_id: if you do it no automatic generation happens.

### DIFF
--- a/hr_employee_id/models/hr_employee.py
+++ b/hr_employee_id/models/hr_employee.py
@@ -52,6 +52,6 @@ class HrEmployee(models.Model):
 
     @api.model
     def create(self, vals):
-        eid = self._generate_identification_id()
-        vals['identification_id'] = eid
+        if not vals.get('identification_id'):
+            vals['identification_id'] = self._generate_identification_id()
         return super(HrEmployee, self).create(vals)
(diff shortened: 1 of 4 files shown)


feketemihai (approved):

gurneyalex (approved):

hr_employee_id/models/hr_employee.py
```diff
@@ -52,6 +52,6 @@ def _generate_identification_id(self):
 
     @api.model
     def create(self, vals):
-        eid = self._generate_identification_id()
-        vals['identification_id'] = eid
+        if not vals.get('identification_id'):
+            vals['identification_id'] = self._generate_identification_id()
```
mpanarin, line 56: can't we use a `vals.setdefault('identification_id', self._generate_identification_id())` here?

mpanarin (commented):
otherwise LGTM )

hr_employee_id/models/hr_employee.py
simahawk: to make it less readable? :stuck_out_tongue:

mpanarin (approved):

guewen (approved):

PR #390, opened 2017-10-24
hr_employee_spouse

gfcapalbo:
Adds a spouse (res.partner) to an employee.
Polygamy not considered.

The corresponding res_partner will also have a tag (spouse) on it , but currently considering moving this feature, keeping consistency between spouse on employee and the tag on the partner causes perhaps uneeded computational overhead .


hr_employee_spouse/README.rst
```diff
@@ -0,0 +1,14 @@
...
+======================
+Spouse for employee
+======================
+
+Module adds a field where you can specify the spouse. It also checks that two
+employees do not have the same spouse. A spouse is a partner.
+Spouse partners have a label that designates them as such.
+Security rules have been added in order to show spouse only to those able to
+see the employee they are spouse of.
+
```
hbrunn: delete empty line

hr_employee_spouse/__openerp__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Employee spouses",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV",
```
hbrunn: add oca as author

hr_employee_spouse/models/hr_employee.py
```diff
@@ -0,0 +1,88 @@
...
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    # you can't marry a company, corporations aren't people
+    partner_spouse_id = fields.Many2one(
+        'res.partner', string='Spouse', domain=[('is_company', '=', False)]
+    )
+
+    # implementing  without considering polygamy.
+    # TODO if needed implement polygamy
```
hbrunn: if you want to mention this, do it in the roadmap section of the readme

hr_employee_spouse/README.rst
hbrunn: ... and use the oca readme template

hr_employee_spouse/data/hr_employee_spouse_data.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="1">
+      <record id="res_partner_category_spouse" model="res.partner.category">
```
hbrunn: the record should be forcecreate I guess? Or if not, your code must cope with the reference being empty

hbrunn (changes requested):

gfcapalbo:
@hbrunn all checks passed  please remove wip tag.

hbrunn (approved):

hr_employee_spouse/models/hr_employee.py
```diff
@@ -0,0 +1,69 @@
...
+        for this in self:
+            if 'partner_spouse_id' in vals:
+                spouse_part = self.env['res.partner'].browse(
+                    vals['partner_spouse_id']
+                )
+                # delete "spouse" tag from old partner
+                this.partner_spouse_id.with_context(
+                    auth_spouse_mod=True).write(
+                        {'category_id': [(3, spouse_tag_id)]}
+                    )
```
hbrunn: move this to before calling super

hr_employee_spouse/models/hr_employee.py
```diff
@@ -0,0 +1,69 @@
...
+                # Add to the spouse partner "spouse" tag.
+                spouse_part.with_context(auth_spouse_mod=True).write(
+                    {'category_id': [(4, spouse_tag_id)]}
+                )
```
hbrunn: keep this here, but do the write on `self.mapped('partner_spouse_id')` (above the same) much more efficient

hr_employee_spouse/models/hr_employee.py
```diff
@@ -0,0 +1,69 @@
...
+        return res
+
+    @api.model
+    def create(self, vals):
+        spouse_tag_id = self.env.ref(
+            'hr_employee_spouse.res_partner_category_spouse'
+        ).id
+        new_empl = super(HrEmployee, self).create(vals)
+        # Add to the spouse partner "spouse" tag.
+        if new_empl.partner_spouse_id:
```
hbrunn: you can so this unconditionally because writing an empty field is a noop

hr_employee_spouse/models/res_partner.py
```diff
@@ -0,0 +1,59 @@
...
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import _, api, models
+from openerp.exceptions import Warning as UserError
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    def spouse_check(self, vals):
```
hbrunn, line 11: i still think we don't need this

hbrunn (changes requested):

hr_employee_spouse/models/hr_employee.py
gfcapalbo: @hbrunn the explanation is here: https://github.com/OCA/hr/pull/390/files/982ee2185c30b0de099ad58574549f1a2b9d0ba3#diff-ddf2d9fe4af185cb14a14f2b6e5d7631R34
(tests reveal fail to rollback transaction)
plus doesn't make more sense to update the tag after the write on the hr_employee model?

hr_employee_spouse/models/res_partner.py
gfcapalbo: we have agreed to see if client likes/wants it.

hr_employee_spouse/models/hr_employee.py
hbrunn: ah, that's what the comment means. keep for now, but this points to some underlying problem that should be solved instead

hr_employee_spouse/models/hr_employee.py
gfcapalbo: eb796ef2ffaca58a88537a1e7f8d348e598ec3f7

hr_employee_spouse/models/hr_employee.py
gfcapalbo: eb796ef2ffaca58a88537a1e7f8d348e598ec3f7

hr_employee_spouse/README.rst
gfcapalbo: 3afb945dc917637b41773c037c9be8fb8e48f415

feketemihai:
@gfcapalbo It is the same as #392 , i think we should improve hr_family, they will be to many modules with only one field added...

feketemihai:
@gfcapalbo @hbrunn Can we close this one, as it is included in #409 ?

hbrunn:
yes

PR #392, opened 2017-10-24
[ADD] hr_employee_marital_status

hbrunn:
Marital status
==============

This module was written to replace the fixed selection in `Marital status` on employees by a configurable selection.

Configuration
=============

#. go to `Human Resources/Settings/Marital statuses` to edit marital statuses


LoisRForgeFlow (approved):
Code review and tested on runbot :+1:

feketemihai:
Hi @hbrunn, it is a good module, but my opinion that this should be a good improvement to hr_family, i think that we should try to compact the modules (this is basically only one fields with the init hook), that way we can maintain easier the repo...what do you think?

hbrunn:
I agree, much better. #390 is from our company too, here the same applies. We'll probably update the PRs next week, I'll set a work in progress label meanwhile

feketemihai:
@hbrunn Thanks for quick reaction..

feketemihai:
@hbrunn Can we close this one, as it is included in #409 ?

hbrunn:
sure

PR #393, opened 2017-10-26
[10.0] [MIG] hr_expense_move_date

nikul-serpentcs:
Expense Move Date
=================

Add move date to expense.
Allows user to force a specific move date

### DIFF
--- a/hr_expense_move_date/README.rst
+++ b/hr_expense_move_date/README.rst
@@ -14,7 +14,7 @@ Installation
 
 To install this module, you need to:
 
-1.  Clone the branch 8.0 of the repository https://github.com/OCA/hr
+1.  Clone the branch 10.0 of the repository https://github.com/OCA/hr
 2.  Add the path to this repository in your configuration (addons-path)
 3.  Update the module list
 4.  Go to menu *Setting -> Modules -> Local Modules*
@@ -36,7 +36,7 @@ On expense data with *Approved* status:
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
-   :target: https://runbot.odoo-community.org/runbot/116/8.0
+   :target: https://runbot.odoo-community.org/runbot/116/10.0
 
 Known issues / Roadmap
 ======================
@@ -66,6 +66,7 @@ Contributors
 ------------
 
 * OpenSynergy Indonesia <https://opensynergy-indonesia.com>
+* Serpent Consulting Services Pvt. Ltd. <support@serpentcs.com>
 
 Maintainer
 ----------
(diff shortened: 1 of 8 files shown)


nikul-serpentcs:
@aheficent Could you please review.

Commits are already preserved and can you please check the [history](https://github.com/nikul-serpentcs/hr/commits/10.0/hr_expense_move_date)

AaronHForgeFlow (changes requested):
Cannot post the journal entry from the expense:
image

AaronHForgeFlow (changes requested):
Failing when trying to post the journal entry from the expense:

image

nikul-serpentcs:
@aheficent Improved code Could you please review.

AaronHForgeFlow (approved):
:+1: It works fine now

feketemihai:
My opinion is that the module should not be ported anymore, since the expense have a dedicated accounting_date field, which overwrites the date of the expense, it's nonsense to add a new field move_date which does the same as accounting_date...

AaronHForgeFlow:
@feketemihai what's that accounting_date field? I did not find it on the hr_expense module

feketemihai:
@aheficent After you submit to manager, the expense appears in the Expense Report, where you can input the accounting date which you post the expense...

feketemihai:
On submit to manager, Odoo creates an expense sheet,

https://github.com/odoo/odoo/blob/10.0/addons/hr_expense/models/hr_expense.py#L45

the sheet have a field accounting_date, which you can modify, if not set the move is created with the expense date.

feketemihai:
@aheficent Check Odoo runbot: 
http://276914-10-0-7c486b.runbot12.odoo.com/web?#id=4&view_type=form&model=hr.expense.sheet&menu_id=504&action=686

AaronHForgeFlow:
you're right @feketemihai I'd close this

source

feketemihai:
@nikul-serpentcs @aheficent To make it easier, not to move from expense to expense sheet to post the moves, you can add the related 'accounting_date' field from sheet_id to expense, with oldname=move_date, and after you send to approval, you can modify accounting_date of the sheet from expense, and to post directly from expense.

nikul-serpentcs:
@aheficent @feketemihai  Improved code Could you please review.

feketemihai:
@nikul-serpentcs Sorry for guided you wrongly, i didn't folow all the process of expense, so the accounting_date is already shown in the expense after the state is changed to approve, so no need for this module, apologizes again...can you also try with odoo runbot the workflow...

AaronHForgeFlow:
@feketemihai but if the field in the expense is related it cannot be different to the one in the sheet, right? Two fields have to be maintained then.

Update: @feketemihai did not see your last answer. Do you agree to close this then?

feketemihai:
@aheficent yes... I will add in the migration issue as deprecated

PR #394, opened 2017-10-27
Add Recomputation of employee names

mpanarin:
Basically mimics the recomputation in
https://github.com/OCA/partner-contact/tree/10.0/partner_firstname

As this are completely different repos, the code was added here to be independent of partner-contact repository


damdam-s (approved):
Functionnally tested on customer project :+1: 
Thanks

feketemihai:
My opinion is that we should depend on partner_firstname, because in this way we will have 2 types of splitting the name, and when you llok at the partners you will have one format, if you look at the employee another format...,so :-1:

mpanarin:
@feketemihai the reasoning behind splitting them was to make this module feature independent of the partner_firstname module, which is in whole another repository

feketemihai:
@mpanarin I agree with you...if you look at the history this was present only in 7 version, from 8 version the module was moved to depend on partner_firstname, see https://github.com/OCA/hr/pull/128...that's why i tend to keep the dependecy to partner_firstname...

feketemihai:
@pedrobaeza Can i have your opinion on this, i fear that this will break more, having partners in one style, employees in another style, i tend to leave the config from partner_firstname take care of this for uniformity...

pedrobaeza:
Yeah, I think the order should be the same for partners and for employees.

feketemihai:
@mpanarin I'm sorry for this, but will create some inconsistencies in the database, so i'm afraid i will reject this one.

PR #406, opened 2017-11-14
[MIG] [11.0] Migrate hr_employee_firstname to 11.0

feketemihai:


hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,82 @@
...
+        for employee in self:
+            if employee.firstname and employee.lastname:
+                employee.name = self._get_name(
+                    employee.lastname, employee.firstname)
+
+    def _firstname_default(self):
+        return ' ' if self.env.context.get('module') else False
+
+    firstname = fields.Char(
+        "Firstname", default=_firstname_default)
```
LoisRForgeFlow, line 28: I would add the `required=True` attribute here, without it I can create a user with firstname and then edit it to delete the first name. (while the displayed name still have that deleted first name)

LoisRForgeFlow (commented):
Thanks for the migration.

Tested on runbot -- Just a comment.

LoisRForgeFlow:
Also, Could you squash the transifex commits together?

hr_employee_firstname/models/hr_employee.py
feketemihai: @lreficent required=True will brake the init_hook on already existing datas...it was required in history (see https://github.com/OCA/hr/commit/196e0db397d119213c01d3261ff6c174ac1b88a6#diff-0d3e28c86aee91593e80482604cd462d)

feketemihai:
@lreficent I would like to keep all the commits as history, if ok, we can squash all when we merge...

hr_employee_firstname/models/hr_employee.py
LoisRForgeFlow: Ok, it's not a big deal.

LoisRForgeFlow:
@feketemihai I think that the idea is to keep the previos versions history of the modules on migration. But, Transifex commits are not very important and they introduce too much noise in some cases, for instnace in this PR you can see two series of 5 and 3 transifex commits that could be harmless squashed. This way we respect the history and keep it the smallest possible.

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_name(self):
+        for employee in self:
+            if employee.firstname and employee.lastname:
+                employee.name = self._get_name(
+                    employee.lastname, employee.firstname)
+
+    def _firstname_default(self):
+        return ' ' if self.env.context.get('module') else False
+
+    firstname = fields.Char(
```
etobella, line 27: Could you change the order of fields and functions following the conventions?
I think it is weird to see the fields so below.
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models

etobella (approved):
LGTM :+1:

hr_employee_firstname/models/hr_employee.py
feketemihai: done

feketemihai:
@lreficent @etobella @OCA/human-resources-maintainers Can i have a last review...

etobella:
LGTM :+1:

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,82 @@
...
+            vals['lastname'] = self.split_name(vals['name'])['lastname']
+            vals['firstname'] = self.split_name(vals['name'])['firstname']
+        res = super(HrEmployee, self).write(vals)
+        if set(vals).intersection(UPDATE_PARTNER_FIELDS):
+            self._update_partner_firstname(self)
+        return res
+
+    @api.model
+    def split_name(self, name):
+        clean_name = u" ".join(name.split(None)) if name else name
```
moylop260: u"" is not needed for py3.5

moylop260 (approved):
👍

PR #409, opened 2017-11-15
8.0 hr family with spouse and marital status

gfcapalbo:


hr_family/README.rst
```diff
@@ -1,16 +1,35 @@
...
+
+#. go to `Human Resources/Settings/Marital statuses` to edit marital statuses
+
 
 Usage
 =====
 
 To use this module, you need to go to any employee form, and look for the
 *Family* tab. You will be able to enter family information there.
+The "spouse" tag will be unmodifiable from the partner form. it is managed by
```
hbrunn: It

hr_family/README.rst
```diff
@@ -1,16 +1,35 @@
...
+editing the spouse selection on HR employee. The error messages when you try to
+delete such tag are verbose.
+So if I try to delete or add the spouse tag from res partner A, i will receive a
```
hbrunn: I

hr_family/migrations/8.0.1.3/post-migration.py
```diff
@@ -0,0 +1,26 @@
...
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    cr.execute(
+        ''' select id, fam_spouse_bk, fam_spouse_employer_bk, 
+        fam_spouse_tel_bk from hr_employee where fam_spouse_bk is not null'''
+    )
+    for empl_id, spouse_name, spouse_empl, spouse_phone in cr.fetchall():
+        spouse = env['res.partner'].create({
+            'name': spouse_name,
+            'employer':spouse_empl,
```
hbrunn: i don't think this is a good idea, see below in the field definition

hr_family/models/hr_employee.py
```diff
@@ -36,3 +15,62 @@ class HrEmployee(models.Model):
...
+    )
+    marital_status_id = fields.Many2one(
+        'hr.employee.marital.status', string='Marital status',
+    )
+
+    @api.constrains('partner_spouse_id')
+    def _check_spouse_uniqueness(self):
+        for this in self:
+            spouse = self.search(
+                [('partner_spouse_id', '=', this.partner_spouse_id.id),
```
hbrunn: what happens if there is no spouse, but another employee without a spouse?

hr_family/views/res_partner.xml
```diff
@@ -0,0 +1,38 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
hbrunn: don't copy other people's copyright

hr_family/models/res_partner.py
```diff
@@ -0,0 +1,64 @@
...
+from openerp import _, api, models, fields
+from openerp.exceptions import Warning as UserError
+
+
+class ResPartner(models.Model):
+    _inherit = 'res.partner'
+
+    # added to port old field fam_spouse_employer on hr_employee
+    # from previous version
+    employer = fields.Char("Employer")
```
hbrunn: I'm not so sure about this. If somebody really wants to save stuff here, she'd add a company partner I guess, this would also then contain the fields to record what is relevant about the employer. I'd either not touch the employer field at all or make it a m2o to res.partner too

hbrunn (changes requested):
travis has an opinion too: https://travis-ci.org/OCA/hr/jobs/302553985#L506

hr_family/models/res_partner.py
gfcapalbo: I thought about doing that. I would have to search for the name and assign the many2one.  if no match is found, then create another new partner . I will do that now. It is good for the OCA to have a moduel that preserves data.

gfcapalbo: 3722b8005ad3a28827b89dbe0563bdeb77be68fa

hr_family/models/hr_employee.py
gfcapalbo: https://github.com/OCA/hr/commit/3722b8005ad3a28827b89dbe0563bdeb77be68fa#diff-9815fc7e7d4f54c5c9e87c6bd11aa486R29

hbrunn (approved):

feketemihai:
@gfcapalbo Tested and the functionalities work, one small suggestion, since the parents and childs have dedicated fields for their birthdays, i think you add the same for the spouse, spouse_birthday...so we can be alligned...

hbrunn:
if we didn't have to not lose data, I'd say also remove the family birthdays because all this can be set in the partner. Then we'd also have to migrate parents and children to partners, which seems a bit out of scope for the PR to just add two smaller modules' functionality to this one

feketemihai:
@hbrunn Yep this will change a little bit the module, so instead changing a lot, parents and childs to be linked as partners, make the module dependant on "partner_contact_birthdate" from partner-contact repo, make the birthdays related to contacts, i would choose adding now a field and on future version to redesign the module, since in many countries for HR declaration we will have to declare name, birthday, ssnid for coinsured persons...so i tend to add one more field now for the birthday...

hr_family/views/res_partner.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<openerp>
+    <data>
+
+        <record id="respartner_form_add_employer" model="ir.ui.view">
+            <field name="name">res.partnere.view.form.add.employer</field>
```
thomaspaulb, line 7: typo in "partner"

thomaspaulb:
@hbrunn How do we normally deal with translations here? As part of the PR, or later in Transifex?

hbrunn:
I think we're still transitioning away from transifex, so better directly here. When you work on this anyways, could you maybe implement https://github.com/OCA/hr/pull/409#issuecomment-345997799 (just add the one field) to speed up merging?

gfcapalbo:
Ok for now we just add the birthday field to spouse. no dependencies , to speed up merge.
I will add a TODO in manifest roadmap with "Include birthday in partner, dependency  on partner_contact_birthdate + migration script for data"  which is the nice solution. Adding field for speed now.

gfcapalbo:
@hbrunn  @feketemihai  thanks to @thomaspaulb  we have what we wanted + translations and some text improvements. Added TODO in README for improved version.

feketemihai (approved):

thomaspaulb (approved):

PR #410, opened 2017-11-24
translations nl for hr.skill

thomaspaulb:


elicoidal (approved):
Usually the translations are handled via www.transifex.com but I think this does not hurt
@pedrobaeza what do you think?

thomaspaulb:
Yes, I wonder what is a good way - Transifex looks great, but I needed the translations in my buildout right away and would have been strange to do the work twice.

Another question while I am at it - is there any reason to make an effort to fill Dutch (nl) as well as Dutch (general)?

hbrunn (approved):

feketemihai (approved):

PR #419, opened 2017-12-22
[IMP] Tri-state calendar (empty, weekly, even/odd weeks), with buttons to switch between either.

thomaspaulb:


resource_calendar_rrule/models/resource_calendar.py
```diff
@@ -78,15 +59,12 @@ def _simplified_from_attendances(self):
...
-                odd = all(
-                    bool(rule._dtstart.isocalendar()[1] % 2)
-                    for occurrence in rule[:2]
-                )
                 key = float(self.env['ir.config_parameter'].get_param(
                     'resource_calendar_rrule.simplified_afternoon',
                     DEFAULT_AFTERNOON_HOUR,
                 )) > attendance.hour_from and 'morning' or 'afternoon'
 
+                odd = bool(rule._dtstart.isocalendar()[1] % 2)
```
thomaspaulb: @hbrunn I don't quite understand what was meant here, but since in the [previous version](https://github.com/OCA/hr/pull/419/files#diff-a560eff6b0bcdbd3621c082965c7cbdcL81) `occurrence` is not used, I think this line of code is equivalent?

thomaspaulb:
@hbrunn Review welcome, see also the PR on Gitlab.

resource_calendar_rrule/models/resource_calendar.py
hbrunn: it's not. This piece of code is meant to detect if this is an rrule implementing the simplified occurrence only for odd weeks. The original looked at two consecutive occurrences, and if both are odd weeks, we know that this is a rule for odd weeks. Otherwise, there would have been an even week in between

hbrunn: so change `rrule._dtstart` to `occurrence`, then the code does what it's meant to be ;-)

resource_calendar_rrule/static/src/xml/resource_calendar_rrule.xml
```diff
@@ -12,11 +12,6 @@
         <t t-if="value.type == 'odd'">
             <t t-set="data_keys" t-value="['data', 'data_odd']" />
         </t>
-        <t t-if="value.type == undefined">
```
hbrunn: also with your changes, we can end up here. A forced other default, a value just happening to be null, that kind of stuff. I think what we need here is what the TODO says, a little link/button that makes a call to create some default calendar.

hbrunn (changes requested):

resource_calendar_rrule/static/src/xml/resource_calendar_rrule.xml
thomaspaulb: Just wish to be a bit more convinced that that is the better solution... currently when value.type is `undefined` we will show a calendar filled with zeros, and people can fill them in as they please. In what situation can that go wrong? If it's only about a link/button, that can also be achieved by having it execute a Javascript function that unhides the input elements?

hbrunn: think about what happens if some code writes a null value in the field

thomaspaulb: I just wrote an empty dict to the `simplified_attendance` field. `record.read(['simplified_attendance'])` now gives me `[{'id': 90, 'simplified_attendance': {}}]`. The form view for the calendar shows me:

image

hbrunn: yes, and that's wrong. We need to be able to make a difference between: "Doesn't work at all for this contract" (zero hours contract) and "No schedule set [yet]" (null)

What you propose here is basically the in my opinion very bad Idea of float and integer fields that null is 0. Bad.

thomaspaulb: Alright, agreed, but that also means we should scrap [this default](https://github.com/OCA/hr/blob/8.0/resource_calendar_rrule/models/resource_calendar.py#L23) and replace it by the null-equivalent?

hbrunn: not so sure actually. Defaults are nice. It's just that for other modules using us, they should be able to change it if they want. Or an admin can override this with some configuration. And then handling null values the way people should expect is a good thing :-)

What I notice is that a possibility to clear out a calendar then should be added too some time, but as I consider it bad style to ask for features during review, I wont

thomaspaulb: Have implemented the null state, and the button to return to it after user had filled something.

hbrunn: nice! now we just need to get travis green

hbrunn:
can you squash commits and fix flake8's complaints here?

resource_calendar_rrule/models/resource_calendar.py
```diff
@@ -225,12 +225,30 @@ def _attendance_from_simplified(self):
...
     def get_attendances_for_weekdays(self, id, weekdays):
+        start_dt = self.env.context.get('start_dt_res_calendar')
+        end_dt = self.env.context.get('end_dt_res_calendar')
         return [
             attendance
             for attendance in self.browse(id).attendance_ids
-            if any(
+            if (
+                not attendance.contract_id.date_end or
+                fields.Date.from_string(attendance.contract_id.date_end) >= end_dt.date()
```
Kiplangatdan: @thomaspaulb using the date from the context and comparing it with date_end of the contracts. Your input please..

feketemihai:
@thomaspaulb Do you still need this one?

thomaspaulb:
@feketemihai Yes, why the question?

feketemihai:
@thomaspaulb because i didn't saw any reaction on comments.

thomaspaulb:
@feketemihai Now addressed
@hbrunn Review please, and advise me on whether to include the "end date of attendance" support here also (re: our other conversation)

hbrunn (approved):

feketemihai (approved):

thomaspaulb:
@hbrunn @feketemihai Thanks a lot for the review. Today I finished another task that was still pending, which is [this commit](https://github.com/OCA/hr/pull/419/commits/dfd6e2ba164006c9557734c03482c1fa09b5e0eb) with some modifications that let the Odoo function `get_working_hours` take the `dt_start` and `until` information into account when trying to find out the total working hours for a given date range.

hbrunn:
like it

hbrunn:
@feketemihai should we merge then?

feketemihai:
@hbrunn Didn't tested last commit, but it's covered by tests, so from my PoV, it's fine to be merged...

PR #420, opened 2017-12-27
[RFR] hr_holidays_compute_days: General refactoring and improvement

pedrobaeza:
* To not depend on hr_contract. Working times are set also in employee, and v11 rely on this data, not in contract working time. This simplifies also the code.
* To not include all the checks while selecting start or end leave date. At the end, it's the same for you: what counts is the number of leave days. As the compute algorithm already handles overlapping leaves, rest days and public holidays, there's no problem with this.
* It handles fraction of days!! New algorithm put the part of the working day that you are leaving. There's a new option at leave type level for keeping the behavior of including full days. This option is checked by default also for keeping backwards compatibility. This code has been made similar to v11.
* It includes UI helps for being better to select dates in the form view, allowing to select only dates instead of date + time.
* The number of days is not manually modifiable by leave users, but to leave managers. If this behavior is not desirable, you can manually add other groups to the view, as this won't be overwritten with module update.

@Tecnativa


hr_holidays_compute_days/models/hr_holidays.py
```diff
@@ -1,101 +1,121 @@
...
-        if (self.date_to and self.date_from) \
-           and (self.date_from <= self.date_to):
-            self.number_of_days_temp = self._recompute_number_of_days()
+    def _inverse_date_to_full(self):
+        """Put end of the day in employee's user timezone, or user timezone
+        as fallback.
+        """
+        for record in self.filtered('date_to_full'):
+            tz_name = record.employee_id.user_id.tz or record.env.user.tz
+            dt = fields.Datetime.from_string(record.date_to_full).replace(
```
feketemihai: This i think this we have to change, to allow user select, because on partial holidays, they can select a holiday at beginning of the day, let's say from 08 to 12. Even if i remove the calculation in holiday status to allow partial results, remove from holiday view the full day checkbox to select hours, i cannot select hours for date_to, the hours are changed to 23:59:59 every time...

feketemihai (dismissed):

hr_holidays_compute_days/models/hr_holidays.py
pedrobaeza: Yeah, I have noticed this also. I'm on it.

hr_holidays_compute_days/models/hr_holidays_status.py
```diff
@@ -1,20 +1,29 @@
...
     exclude_rest_days = fields.Boolean(
-        'Exclude Rest Days',
+        string='Exclude Rest Days',
+        default=True,
         help="If enabled, the employee's day off is skipped in leave days "
              "calculation.",
     )
     exclude_public_holidays = fields.Boolean(
-        'Exclude Public Holidays',
+        string='Exclude Public Holidays',
```
jcdrubay, line 19: I am being picky, but for those 3 fields you should not use the string attribute as the technical name is the same as the label.

hr_holidays_compute_days/README.rst
```diff
@@ -6,30 +6,49 @@
...
-Computes the actual days for which employee will be on leave taking into 
-account both rest days and public holidays
+Computes the actual days for which employee will be on leave taking into
+account:
 
-Installation
-============
-
-No additional steps necessary
+* Other leaves in the same interval.
```
jcdrubay: Remove the dot at the end of this line.

jcdrubay (changes requested):
Tested on runbot, it worked well!

2 comments for very minor request of changes :)

Beautiful code :+1:

Yajo:
It has merge conflicts.

yajo:
I rebased to fix some merge conflicts.

hr_holidays_compute_days/README.rst
chienandalu: Why? :thinking:

hr_holidays_compute_days/models/hr_holidays_status.py
chienandalu: You'd loose control over capitalization of the words.

pedrobaeza: Well, in this case it would be the same, but for this I was looking the "better explicit than implicit".

hr_holidays_compute_days/README.rst
jcdrubay: I think it should be consistent between the 3 bullet points.

chienandalu: In that case it's the period in the next line what's missing :slightly_smiling_face:

rafaelbn (approved):
Tested in runbot :+1:

rafaelbn:
We have here 2 approvals, just one missing please @feketemihai :smile: Thanks!

hr_holidays_compute_days/models/hr_holidays.py
pedrobaeza: This has been solved with latest code.

pedrobaeza:
I have solved the comment from @feketemihai, so I proceed with the merge for finally unlocking this.

PR #421, opened 2018-01-03
[ADD] hr_attendance_report_theoretical_time

pedrobaeza:
Depends on:
- [x] hr_holidays_compute_days refactorization in #420 

Theoretical vs Attended Time Analysis
===============================

This module adds a new report called "Theoretical vs Attended Time Analysis" that compares worked time, measured through attendances records, with the theoretical time, computed from employee's working calendar, public holidays and employee specific leaves.

There is the possibility of counting as theoretical time some leave types if specified in them.

Configuration
=============

You need to be at least "Attendance / Manual Attendance" for being able to see
the attendances report.

For including some leave types in the teorethical time, you have to:

* Go to Leaves > Configuration.
* Select leave type you want to include.
* Check the mark "Include in theoretical hours".

Usage
=====

* Go to *Attendances > Reports > Theoretical vs Attended Time Analysis*.
* Check pivot table or look at the graph view.

Known issues / Roadmap
======================

* Employees with less than 1 week in the company will show full week theoretical hours.
* Activate ORM cache for improving performance on computing theoretical hours, but assuring that the cache is cleared when the conditions of the computation changes.
* If you change employee's working time, theoretical hours for whole employee's life will be computed according this new calendar. You have to define start and end dates inside the calendar for avoiding this.

@Tecnativa


rafaelbn:
@pedrobaeza could we add here an ACL for all employees? In order that all employees can use this report (by default is only for "Official" or "Manager"). The report will show fo each employee their own hours and theoretical hours not of other employees. Thanks

rafaelbn:
Hi @pedrobaeza once https://github.com/OCA/hr/pull/420 is merge could you rebase please?

Could you please add https://github.com/OCA/hr/pull/421#issuecomment-409301518

pedrobaeza:
This one is ready to be reviewed. About the menu access, that can be done at DB level, adding the group on the menu, or in other module, but not here.

hr_attendance_report_theoretical_time/readme/CREDITS.rst
```diff
@@ -0,0 +1,3 @@
+**Images**
+
+* Font Awesome: `Icon <http://fontawesome.io>`_.
```
tarteo, line 3: Is this needed? It's already included by Odoo, right?

hr_attendance_report_theoretical_time/models/hr_attendance.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+            calendar = record.employee_id.calendar_id
+            resource = record.employee_id.resource_id
+            # Compute manually leaves for excluding the desired ones (as it
+            # can't be done through inheritance on calendar.get_leave_intervals
+            # method)
+            leaves = []
+            for leave in calendar.leave_ids:
+                holiday = leave.holiday_id
+                if (holiday.holiday_status_id.include_in_theoretical or
```
tarteo: ```if (not holiday.holiday_status_id.include_in_theoretical or```?

tarteo (changes requested):
Cool module! include_in_theoretical seems to be doing the opposite.

hr_attendance_report_theoretical_time/readme/CREDITS.rst
pedrobaeza: Yes, because the icon is made from one Font Awesome glyph and I give credits for that.

hr_attendance_report_theoretical_time/models/hr_attendance.py
pedrobaeza: As we want to include this leave in theoretical hours, we shouldn't exclude this leave in the computation. That's why I include this condition for continuing the loop.

pedrobaeza:
I'm finishing a huge refactoring of this for making an isolated report and other surprises :wink:

I will push this later and give more explanations.

pedrobaeza:
Main comment updated with the new README of the module

pedrobaeza:
@tarteo I have finished fine-tuning of the module. Performance is not the best, but it's acceptable for a controlled amount of dates/employees.

tarteo (approved):

AaronHForgeFlow (commented):
I like the idea of this report. :+1: 

I could not include some leaves as worked time in the report. Included the leave type as "Include in theoretical hours" and approved the leave request for the current month. The Theoretical time was not reduced in the report. Not sure what I am missing.

pedrobaeza:
@aheficent you are misunderstanding the check: it's for not reducing, but keep the theoretical hours. Not checking the mark is when they are reduced....

What is the idea behind that mark? To have a leave type for putting leaves that are only for compensating extra hours you have worked before. Imagine this situation in a work week with 40 hours:

- Monday: I worked 10 hours
- Tuesday: I worked 10 hours
- Wednesday: I worked 10 hours
- Thursday: I worked 10 hours
- Friday: I ask for a compensation leave (type with that mark), as I have already worked 40 hours. And this will make that people knows that I'm not at the office.

On the report, whole week should put 40 theoretical hours against 40 worked hours - 8 per day - (although they were on previous days, not on Friday).

On contrary, if you want to take a holiday one of that days, you should ask for a leave type without the check, and then the whole week will be 32 theoretical hours against 32 worked hours.

Do you understand? Maybe we can add this explanation to the readme...

AaronHForgeFlow:
Ok I see the point now. I think the Readme is fine.

AaronHForgeFlow (approved):
Functional test :+1:

pedrobaeza:
I have added an extra feature for manually selecting the start date of generated date series according working schedule, and also improve even more the README, with the use case I have written in comments.

pedrobaeza:
Merging as no news.

PR #423, opened 2018-01-05
[MIG][11.0] Hr Phone validation

fmdl:


hr_phone_validation/models/hr_employee.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class HrEmployee(models.Model):
+    _name = 'hr.employee'
+    _inherit = ['hr.employee', 'phone.validation.mixin']
+
+    @api.onchange('work_phone', 'country_id')
+    def _onchange_work_phone_validation(self):
+        if self.work_phone:
+            self.work_phone = self.phone_format(self.work_phone)
```
feketemihai, line 11: Hi @fmdl , small module but interesting, i have a remark that here you should add the employee country as parameter, since it can be different from the company country...

hr_phone_validation/models/hr_employee.py
```diff
@@ -0,0 +1,18 @@
...
+
+    @api.onchange('mobile_phone', 'country_id')
+    def _onchange_mobile_phone_validation(self):
+        if self.mobile_phone:
+            self.mobile_phone = self.phone_format(self.mobile_phone)
```
feketemihai, line 16: @fmdl The same here...

feketemihai (changes requested):

fmdl:
Thank for your remark, I will update that quickly.

hr_phone_validation/models/hr_employee.py
fmdl: HI @feketemihai you don't need to add country=self.country it is already built in here : https://github.com/odoo/odoo/blob/11.0/addons/phone_validation/models/phone_validation_mixin.py#L11

feketemihai (approved):

hr_phone_validation/models/hr_employee.py
feketemihai: You're right, thanks for the explanation...

hr_phone_validation/models/hr_employee.py
feketemihai: You're right, thanks for the explanation...

hr_phone_validation/README.rst
```diff
@@ -0,0 +1,50 @@
...
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/116/11.0
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/server-tools/issues>`_. In case of trouble, please
```
simahawk: https://github.com/OCA/hr

simahawk: https://github.com/OCA/hr

hr_phone_validation/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
```
simahawk: remove encoding from all py files

hr_phone_validation/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+{
+    'name': 'HR Phone Numbers Validation',
+    'summary': 'Validate and format phone numbers for leads and contacts',
+    'version': '11.0.1.0.0',
+    'category': 'Human Resources',
+    'author': "Florent de Labarre,"
+              "Odoo Community Association (OCA)",
+    'website': 'https://odoo-community.org/',
```
simahawk: https://github.com/OCA/hr

simahawk (changes requested):

simahawk:
can you reword commits meaningfully? `[add] hr_phone_validation` or `[MIG][11.0] hr_phone_validation` would be nice ;)

feketemihai:
@fmdl can you please update @simahawk comments, to have it clean.

fmdl:
@feketemihai it is update

fmdl:
Close in https://github.com/OCA/hr/pull/444

PR #427, opened 2018-01-27
11.0 fix employee firstname test

feketemihai:


feketemihai:
@pedrobaeza Can i have some help please, i've pushed https://github.com/OCA/hr/pull/406 into main branch( but by mistake i choose merge commits, not squash), and now the tests fail, but i don't get it why, on the PR everything was ok, all were green, but now, i really don't know why...

feketemihai:
Stange is that locally i don't have any error on tests, both modules...with Odoo latest revision..

hbrunn (approved):
you could have also just explicitly called the decorated function - we're not testing if the framework works, but if the functions we use do what they should.  But I think it's okay the way it is

feketemihai:
@pedrobaeza @espo-tony Can i have your review here, since the module is merged and we have a bug, and other PR's are red without this..

hr_employee_firstname/tests/test_hr_employee_firstname.py
```diff
@@ -46,12 +46,22 @@ def test_onchange(self):
         """
         Validate the get_name method is not failing
         """
-        field_onchange = self.employee1_id._onchange_spec()
+        field_onchange = self.employee_model._onchange_spec()
```
pedrobaeza: You are running the onchange over an empty recordset, which is for sure invalid. If you want to set some kind of default, do it over a `NewId` recordset (created with `new()` method)

pedrobaeza (changes requested):

hr_employee_firstname/tests/test_hr_employee_firstname.py
feketemihai: @pedrobaeza It's done on 59 line, https://github.com/OCA/hr/pull/427/files/45f54dd343a4f219ddcc6fa4c356a78839705889#diff-c2742713bb31aafa07bcfdfb4244cb82R59

pedrobaeza: No, you are making some `assert` after this call, so it's not correct.

feketemihai: @pedrobaeza I inspired from Odoo base: https://github.com/odoo/odoo/blob/cd48a806ac636806130efcf5c055f73758cb0902/odoo/addons/test_new_api/tests/test_onchange.py

pedrobaeza: Which line? Any way, what I'm proposing is

```python
field_onchange = self.employee_model.new({})
field_onchange._onchange_spec()
self.assertEqual(field_onchange.get('firstname'), '1')
self.assertEqual(field_onchange.get('lastname'), '1')
```
It's for assuring no weird behavior is got over versions.

pedrobaeza (approved):

PR #430, opened 2018-02-26
10.0 resource calendar rrule

hbrunn:
relevant commit is dad3ad6, the rest already was merged in https://github.com/OCA/hr/pull/307


resource_calendar_rrule/models/resource_calendar.py
```diff
@@ -0,0 +1,250 @@
...
+            return super(ResourceCalendar, self).get_weekdays(
+                default_weekdays=default_weekdays
+            )
+        return list(set(
+            date.weekday()
+            for attendance in self.mapped('attendance_ids')
+            for date, _ in attendance._iter_rrule(
+                *self._get_check_interval()
+            )
+        ))
```
thomaspaulb: We are having a potentially nasty bug here. 

If we run `get_working_hours()` on an employee's calendar, it goes and runs `get_working_hours_per_day()` on each of the weekdays that are returned by this function `get_weekdays()`. These weekdays, according to the code here, are determined by collecting the weekdays from the attendances in the interval of `_get_check_interval()`.

Therefore, with this default code, if it's now september and the employee has attendances on monday, tuesday and wednesday, but we are getting his working hours from a week in April where he was also working on Thursday, the Thursday is truncated from the total summation of working hours.

Seeing the following solutions:

1. We will not use the check interval in determining the weekdays but we use **all** the attendances from the calendar, like the default Odoo function does,
2. We will not use the check interval but we use an interval taken from eg. the first week of the attendance that we are getting the weekdays for,
3. We forget about restricting weekdays altogether and check for each weekday

I don't have the overview for the impact of each of these choices, could you help brainstorm?

thomaspaulb (changes requested):

resource_calendar_rrule/models/resource_calendar.py
hbrunn: I currently can't find a good reason why we didn't do (3) in the first place. Maybe that was a well meant but misguided optimization.
Concerning optimization, I think we should try to create more efficient rrules, especially when using the simplified calendar. I noticed quite a speed penalty since we introduced this. So for example if a person works full time, it should not generate 10 rules, one for every morning, one for every afternoon, but just one for all afternoons and one for all mornings. Or (and I think that's not as crazy as it sounds) just precompute asynchronously all working intervals for [-N, +M] years, and replace all of the expensive computations by an SQL lookup. And if a request comes in for a not yet computed interval, it's computed on the fly.

thomaspaulb: Hm, (3) in practice (simply returning all days from get_weekdays) interferes with other uses of get_weekdays eg in [get_previous_day](https://github.com/OCA/OCB/blob/10.0/addons/resource/models/resource.py#L215). Went for 2 in the end.

thomaspaulb: https://github.com/sunflowerit/hr/commit/782e18083c22d2dc14d0c33fda3f026c3ecf4fb6

bennals:
@hbrunn This module looks extremely useful.  Do you have plans to migrate it to version 11?

resource_calendar_rrule/models/resource_calendar.py
```diff
@@ -0,0 +1,275 @@
...
+
+                if rule._dtstart < start:
+                    start = rule._dtstart
+                if rule._until and (not stop or stop > rule._until):
+                    stop = rule._until
+
+                odd = all(
+                    bool(rule._dtstart.isocalendar()[1] % 2)
+                    for occurrence in rule[:2]
+                )
```
thomaspaulb, line 84: @hbrunn Turns out this goes wrong in cases like December 2009 - January 2010 where there is 'week 53' followed by 'week 1'. There needs to be an extra check on `_interval` being 2.

resource_calendar_rrule/models/resource_calendar.py
```diff
@@ -0,0 +1,275 @@
...
+                result['rdate'].append(fields.Date.to_string(rdate))
+
+        def sum_all(x):
+            return x['morning'] + x['afternoon']
+
+        if not sum(map(sum_all, result['data'])):
+            result['data'] = result['data_odd']
+            result.pop('data_odd')
+        elif not sum(map(sum_all, result['data_odd'])):
+            result.pop('data_odd')
```
thomaspaulb, line 108: This code is also questionable to me - what if people want to specify an "even" or "odd" schedule with 0 hours in the other week?

thomaspaulb (changes requested):

PR #432, opened 2018-03-08
[ADD] Hr contract amendment

EBII:
automatic backup hr contract changes visible in contract view.


hr_contract_amendment/README.rst
```diff
@@ -0,0 +1,70 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/agpl
+   :alt: License: AGPL-3
+
+=====================
+HR CONTRACT AMENDMENT
```
feketemihai: @EBII Good module, but still need to be improved a little bit...Use title formating here, no upper...

hr_contract_amendment/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+This module extends the functionality of HR contract to keep history of change
+and to allow you to saw when and who made contract change
+
+Installation
+============
+
+To install this module, you need to:
+
+ just install it
```
feketemihai: Not all persons are technical, so expand this with a standard instalation steps:
+To install this module, you need to:
+
+* clone the branch 11.0 of the repository https://github.com/OCA/hr
+* add the path to this repository in your configuration (addons-path)
+* update the module list
+* search for "HR Contract Amendment" in your addons
+* install the module

hr_contract_amendment/README.rst
```diff
@@ -0,0 +1,70 @@
...
+Known issues / Roadmap
+======================
+
+may add print button on amendment
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/HR/issues>`_. In case of trouble, please
```
feketemihai: <https://github.com/OCA/hr/issues>

hr_contract_amendment/data/sequence.xml
```diff
@@ -0,0 +1,9 @@
+<odoo>
```
feketemihai: Usefull to have a noupdate on it <odoo noupdate="1">

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+#   Copyright (C) 2018 EBII (https://www.saaslys.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models, _
+
+
+class HrAmendmentContractChanges(models.Model):
```
feketemihai: use the same name as the model HRContractAmendmentChanges

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+    _name = 'hr.contract.amendment.changes'
+    _description = 'HR contract amendment change'
+
+    amendment = fields.Many2one(
```
feketemihai: amendment_id since is many2one..

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+        comodel_name='hr.contract.amendment', string="Amendment",
+        Require=True, ondelete="cascade")
```
feketemihai: Require not an option, required yes

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+    value_name = fields.Char(string="Contract attribute name")
+    value_before = fields.Char(string="Value before change")
+    value_new = fields.Char(string="New value")
+
+
+class HrAmendmentContract(models.Model):
```
feketemihai: use the same order as in name of the model

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+    _name = 'hr.contract.amendment'
+    _description = 'HR contract amendment'
+
+    contract_id = fields.Many2one(
+        comodel_name='hr.contract', string="Referenced Contract",
+        required=True)
+    amendment_number = fields.Char(string="Amendment name")
```
feketemihai: just call it name

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+    date = fields.Date(required=True, default=fields.Date.context_today)
+    user_id = fields.Many2one('res.users', string='Registred by',
```
feketemihai: string="Registered by"

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+                              default=lambda self: self.env.user)
+    amendment_type = fields.Selection(selection=[
```
feketemihai: for better inheritance, use a method to get the values

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
```
feketemihai, line 1: Use a different file for contract and amendment, split it at least in 2

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+class HrContract(models.Model):
+    _inherit = 'hr.contract'
+
+    amendment_ids = fields.One2many(
+        comodel_name='hr.contract.amendment', inverse_name="contract_id",
+        string="contract", required=False)
+
+    def write(self, vals):
+        sequence = self.env['ir.sequence'].next_by_code("amendment.number")
+        if 'job_id' in vals.keys() and len(vals) == 1:
```
feketemihai: I would add some config to allow users to select fields for what they want to create an amendment, i don't want a new amendment for adding something is notes, advantages, change visa_expire...and probably to be linked with the type of change...

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            type = 'CONTRACT_SITUATION'
+
+        amendmentVals = {
+            'contract_id': self.id,
+            'amendment_number': sequence,
+            'amendment_type': type,
+        }
+
+        id_amendment = self.env['hr.contract.amendment'].create(amendmentVals)
```
feketemihai: use proper notation, it's not an Id, it's a record..

feketemihai (changes requested):

hr_contract_amendment/models/hr_contract.py
EBII: name of file ? 
make two hr_contract.py & hr_amendment.py  ?

hr_contract_amendment/models/hr_contract.py
EBII: sure test a list of fields
 how simple to do that ?

hr_contract_amendment/models/hr_contract.py
EBII: seen many time in odoo code with 
    grep -rn --include="*.py" "Many2one('res.users'"
but have not found somethings like Many2one('res.users', compute="_get_current_user"
does an example or not understand what u mean

hr_contract_amendment/models/hr_contract.py
EBII: ok  it is done 
Thanks for reviews !

EBII: made

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,68 @@
...
+
+class HrContract(models.Model):
+    _inherit = 'hr.contract'
+
+    amendment_ids = fields.One2many(
+        comodel_name='hr.contract.amendment', inverse_name="contract_id",
+        string="contract", required=False)
+
+    @api.multi
+    def write(self, vals):
```
legalsylvain: i see this algorithm a little complex. (I mean Overloading write function to make such changes.
What about 
* set job_id, wage as readonly on hr.contract
* create a wizard to allow to change the values

hr_contract_amendment/README.rst
```diff
@@ -0,0 +1,80 @@
...
+For usability field 'notes' is not save has save until an other fields is change,
+see write method in contract.py for comments.
+
+so you can add notes to every amendment and set notes outside of amendment system
+
+
+Installation
+============
+
+To install this module, you need to:
```
legalsylvain: I'm not in favor to write such part.

* It is useless to write the same things in the 2000 OCA modules.
* it is wrong, because there are a lot of ways to install a module. (appstore, pip, etc...)

Installation part is generally removed, except if there are specifics things to do before / during the installation. (key in openerp.cfg file, lib installation, etc...)

regards.

hr_contract_amendment/models/hr_contract.py
```diff
@@ -0,0 +1,68 @@
...
+
+        EXCLUDED_FIELDS = ['notes']
+        # don't save notes if it's the only one changes
+        if EXCLUDED_FIELDS[0] in vals.keys() and len(vals) == 1:
+            return super(HrContract, self).write(vals)
+        else:
+
+            for res in self:
+                sequence = res.env['ir.sequence'].next_by_code("name")
+                amendmentVals = {
```
legalsylvain: could you set this code out the function, with the classical _prepare_contract_amendment.
It allows easy overload.

hr_contract_amendment/models/hr_contract_amendment.py
```diff
@@ -0,0 +1,39 @@
...
+    _description = 'HR contract amendment'
+
+    contract_id = fields.Many2one(
+        comodel_name='hr.contract', string="Referenced Contract",
+        required=True)
+    name = fields.Char(string="Amendment name")
+    date = fields.Date(required=True, default=fields.Date.context_today)
+    user_id = fields.Many2one('res.users', string='Registered by',
+                              default=lambda self: self.env.user)
+    amendment_type = fields.Selection(selection=[
```
legalsylvain: Design question. Generally, in a amendment, it is possible to change many things in a same time. (salary, working hours, etc...)
a Many2Many implementation will suit better to the use case, don't you think ?

In france, an amendment (avenant au contrat) can belong severals changes.

legalsylvain (changes requested):
Hi @EBII. 
thanks for sharing this module. One main design question inline (about one2many relation), and some cosmetic ones.
kind regards.

hr_contract_amendment/models/hr_contract_amendment.py
andhit-r: I agree with @legalsylvain . Amendment should cover change for several changes. I have tested on runbot, this module does not cover multiple changes. IMO, similar result can be achieve using auditlog module.

IMO, amendment document should be created first. When amendment approved, it will change contract value.

EBII: :| all kind of change at the same time on contract are save in this module. 

you got list of change in the amendment ? 

Only the fields note is'nt write if it is alone in vals 
i don't uderstand what you claim on ?

hr_contract_amendment/models/hr_contract.py
EBII: it is automatic process to follow change on contract.

keep each change between you create it and the time you look in is saved to know what about changes.
Who , when which values before and after ..  without one boring mouse click for user. 
where add this either than on write method ?

hr_contract_amendment/README.rst
EBII: @feketemihai   @legalsylvain  

ok you got to choose what is the rule and apply for all  
i've write 
"just install it" 
I change one a time with standard (no really more value for the user)
Now you ask to remove  are working for the same repository ?

feketemihai:
@EBII Can you please follow @andhit-r @legalsylvain comments, mostly on the design of the amendment and procedure of updating contract.

EBII:
@feketemihai @andhit-r @legalsylvain  thanks a lot for review, i follow your recommendations with prepare. Multiple settings already exists. the if elif else sequence is only for naming contrat modification.

Regards

hr_contract_amendment/README.rst
alexey-pelykh: This obvious instructions should be omitted I guess at all, unless there are any module-specific steps

alexey-pelykh:
@EBII any plans for moving this towards 12.0 as well?

PR #435, opened 2018-03-30
[10.0] [FIX] Fix error when journal don't have update_posted checked.

feketemihai:
When the payslip journal is not set as update_posted it will raise an error, so instead we are reversing the move of the payslip and set move_id field to False.


feketemihai:
Codecov not passing since no tests are in the module.

AaronHForgeFlow (approved):
:+1: Thanks

MiquelRForgeFlow (approved):

pedrobaeza:
Changing module version on merge.

PR #438, opened 2018-04-09
[11.0][FIX] Dangerous use of 'replace' from hr.employee view.

ddufresne:
This issue was flagged by travis in the following build https://travis-ci.org/OCA/hr/jobs/364219109

### DIFF
--- a/hr_employee_firstname/__manifest__.py
+++ b/hr_employee_firstname/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'HR Employee First Name, Last Name',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change module version number

pedrobaeza:
And better to put a high priority any way in the view, as you are changing a label. I would even prefer to hide the label and add a new one, so both works.

ddufresne:
@pedrobaeza I agree

feketemihai (approved):

pedrobaeza (approved):

PR #441, opened 2018-04-23
[UPD] use updated .travis.yml from mqt

hbrunn:
this is an update from https://github.com/OCA/maintainer-quality-tools/pull/538 that should give us green builds in 10 again


pedrobaeza (approved):
Indeed green. Fast-tracking it.

PR #442, opened 2018-04-27
[IMP][8.0]hr_skill:add states, approve or reject

Kiplangatdan:


hr_skill/views/hr_employee.xml
```diff
@@ -8,7 +8,14 @@
...
             <notebook position="inside">
                 <page string="Skills">
-                    <field name="skill_ids" />
+                    <field name="skill_ids" domain="[('state', '=', 'approved')]">
+                        <tree editable="top">
+                            <field name="display_name" readonly="1"/>
+                            <field name="state" readonly="1"/>
+                            <field name="free_text" attrs="{'readonly': [('is_free_text_field', '=', False)]}"/>
+                            <field name="is_free_text_field" invisible="1"/>
+                        </tree>
```
thomaspaulb: Here also, could you check if the effect is the same if this inline view is removed here and instead you change [the generic tree view for `hr.skill`](https://github.com/OCA/hr/blob/8.0/hr_skill/views/hr_skill.xml#L10)?

If you want to have some deviations in that one, you can create a new tree view for that, assign a high priority (eg. 101) and reference that one explicitly using `view_id` in [the action](https://github.com/OCA/hr/blob/8.0/hr_skill/views/hr_skill.xml#L57).

hr_skill/models/hr_skill.py
```diff
@@ -50,3 +56,23 @@ def name_get(self):
...
+
+    @api.onchange('free_text')
+    def _onchange_free_text(self):
+        if self.is_free_text_field:
+            hr_skill_obj = self.env['hr.skill']
+            hr_skill_obj.create({
+                'name': self.free_text,
+                'state': 'pending',
+                'parent_id': self._origin.id,
+            })
```
thomaspaulb: Hmmmm i think this is a bad idea.. onchange gets executed very often, like in, any time you change the free_text field. So many skill records will be created, when in fact, you want the "create" action for the new skill to be done only once. So I think it's better to go the way I said on GitLab, remove the onchange function and instead inherit the "write" function of `hr.skill` and check for "free_text" to be in vals. If yes, you will add `state=pending` and call super of the write, thereby creating the record.

thomaspaulb (changes requested):

thomaspaulb (approved):

thomaspaulb:
@Kiplangatdan I guess we can close this, since it's 8.0 and pretty specific functionality. Perhaps someday someone can pick up where we left off and push to 14.0

PR #445, opened 2018-05-16
[11.0][MIG] hr_holidays_settings

tarteo:


pedrobaeza:
Please squash together adjacent "OCA Transbot.." commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

tarteo:
@pedrobaeza Squashed them :)

hr_holidays_settings/views/res_config_settings.xml
```diff
@@ -0,0 +1,40 @@
...
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="base.res_config_settings_view_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//div[hasclass('settings')]" position="inside">
+                <div class="app_settings_block"
+                     data-string="Leaves management" string="Leaves Management"
+                     data-key="hr_holidays" groups="hr.group_hr_manager">
+                    <h2>Leaves Management</h2>
+                    <div class="row mt16 o_settings_container" title="Leaves Management">
+
```
hieulucky111, line 14: why do you put the empty div in here? Can we remove it?

feketemihai:
@tarteo @pedrobaeza Do you think one should be autoinstalled, if you install hr_holidays?

pedrobaeza:
Well, this is not needed unless you install one module that uses it, so better to let it not auto-installed, but installed as dependency when installing one of the other modules.

feketemihai (approved):

hr_holidays_settings/views/res_config_settings.xml
tarteo: We can remove it. But this way we have a 'default section' for settings. People who depend on this module can just add a div with class o_setting_box to extend it.

astirpe (approved):

hr_holidays_settings/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "HR Holidays Settings",
+    "summary": "Enables Settings Form for HR Holidays.",
+    "version": "11.0.1.1.0",
```
MiquelRForgeFlow, line 6: should have been `11.0.1.0.0`

PR #450, opened 2018-06-21
[11.0][MIG] hr_holidays_hour

astirpe:


hr_holidays_hour/__manifest__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Leave Management in hours',
+    'summary': 'Holidays, Allocation and Leave Requests in Hours',
+    'author': 'Onestein, Odoo Community Association (OCA)',
+    'website': 'http://www.onestein.eu',
```
JayVora-SerpentCS: website to be repo link plz.

hr_holidays_hour/models/hr_holidays_status.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    @api.multi
+    def name_get(self):
+        if not self._context.get('employee_id', False):
+            # leave counts is based on employee_id, would be
+            # inaccurate if not based on correct employee
+            return super(HrHolidaysStatus, self).name_get()
+
+        res = []
```
JayVora-SerpentCS: This line should better be at the start of the def()

JayVora-SerpentCS (changes requested):

hr_holidays_hour/__manifest__.py
astirpe: Done!

hr_holidays_hour/models/hr_holidays.py
```diff
@@ -0,0 +1,147 @@
+# Copyright 2017 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError, Warning
```
MiquelRForgeFlow: `from odoo.exceptions import UserError, ValidationError`

hr_holidays_hour/report/hr_holidays_report.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2017 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models, tools
+
+
+class HrHolidaysRemainingLeavesUser(models.Model):
```
MiquelRForgeFlow, line 7: `HrHolidaysRemainingLeavesUserReport` or `HrHolidaysReport`

MiquelRForgeFlow (approved):
LGTM

JayVora-SerpentCS (approved):

rafaelbn (approved):
Thanks! :+1: 

Who could merge this one?

JayVora-SerpentCS:
I have no access unfortunately

hr_holidays_hour/models/hr_holidays.py
```diff
@@ -0,0 +1,147 @@
...
+        employee = self.employee_id
+        if not employee and (self.date_to or self.date_from):
+            raise UserError(_('Set an employee first!'))
+
+    @api.multi
+    def _compute_work_hours(self):
+        self.ensure_one()
+        employee = self.employee_id
+        from_dt = fields.Datetime.from_string(self.date_from)
+        to_dt = fields.Datetime.from_string(self.date_to)
```
hieulucky111: i think we should put this code under `if self.date_from and self.date_to:`

hr_holidays_hour/models/hr_holidays.py
```diff
@@ -0,0 +1,147 @@
...
+                leave.holiday_status_id.name,
+                leave.number_of_hours_temp
+            )))
+        return res
+
+    @api.multi
+    def _prepare_create_by_category(self, employee):
+        values = super(HrHolidays, self)._prepare_create_by_category(employee)
+        values.update({
+            'number_of_hours_temp': self.number_of_hours_temp,
```
hieulucky111: you should put `self.ensure_one()`

hieulucky111 (changes requested):
Can you check my comment?

astirpe:
@hieulucky111 done. Thanks!

hieulucky111 (approved):

hr_holidays_hour/models/hr_employee.py
```diff
@@ -0,0 +1,39 @@
...
+            ('state', '=', 'validate')],
+            fields=['number_of_hours', 'employee_id'],
+            groupby=['employee_id']
+        )
+        mapping = dict(
+            [(leave['employee_id'][0], leave['number_of_hours'])
+             for leave in leaves]
+        )
+        for employee in self:
+            employee.leaves_count = mapping.get(employee.id)
```
pedrobaeza: Default to 0 if non existing:
```python
employee.leaves_count = mapping.get(employee.id, 0)
```

hr_holidays_hour/models/hr_employee.py
```diff
@@ -0,0 +1,39 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    @api.multi
+    def _compute_leaves_count(self):
```
pedrobaeza, line 11: This is something out of the scope of the module. And there's already an smart-button for leaves in v11:

seleccion_006

hr_holidays_hour/report/hr_holidays_report.py
```diff
@@ -0,0 +1,45 @@
...
+
+    no_of_hours = fields.Float('Approved hours')
+    virtual_hours = fields.Float('Virtual hours')
+    no_of_leaves = fields.Integer('Remaining hours')
+    employee_id = fields.Many2one('hr.employee', 'Employee')
+
+    def init(self):
+        cr = self._cr
+        tools.drop_view_if_exists(cr, 'hr_holidays_remaining_leaves_user')
+        cr.execute("""
```
pedrobaeza: This is not inheritable. Please use a similar technique like the one I use in:

https://github.com/OCA/purchase-reporting/blob/9.0/purchase_reporting_weight/reports/purchase_report.py

pedrobaeza (changes requested):

hr_holidays_hour/models/hr_employee.py
astirpe: I don't think is out of scope. I could remove this method, no problem, but then the user will read the computed leaves in days instead of hours. At the end I will create for our customers an extra module to convert this amount to hours. So if for you is important to remove it, I will do it ... do you confirm?

hr_holidays_hour/report/hr_holidays_report.py
astirpe: Interesting, thanks!

hr_holidays_hour/models/hr_employee.py
pedrobaeza: OK, I see, then it's not out of scope indeed. Leave it :smiley:

pedrobaeza:
@astirpe have you seen https://github.com/OCA/hr/pull/465#issuecomment-417972491? Is it OK for you?

astirpe:
@pedrobaeza done! Thanks.

pedrobaeza (approved):

PR #453, opened 2018-06-25
[MIG][11.0]hr_contract_multi_jobs: Migration to 11.0

Trivedi-Vacha-SerpentCS:
[MIG][11.0]hr_contract_multi_jobs: Migrated module in 11.0


hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,48 @@
...
+    @api.constrains('contract_job_ids')
+    def _check_one_main_job(self):
+        for contract in self:
+            # if the contract has no job assigned, a main job
+            # is not required. Otherwise, one main job assigned is
+            # required.
+            if contract.contract_job_ids:
+                main_jobs = contract.contract_job_ids.filtered('is_main_job')
+                if len(main_jobs) != 1:
+                    raise Warning(
```
nikul-serpentcs: @Trivedi-Vacha-SerpentCS Remove `Warning` use `userError`

hr_contract_multi_jobs/tests/test_hr_contract_multi_jobs.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_one_main_jobs(self):
+        """
+        Validate the _check_one_main_job method
+        when contract has one assigned job
+        and check is the job_id is set.
+        """
+        self.contract_id.write({'contract_job_ids':
+                                [(0, 0, {'job_id': self.job_id.id,
+                                         'is_main_job': True})]})
+        self.assertTrue(self.contract_id.job_id.id == self.job_id.id)
```
nikul-serpentcs: @Trivedi-Vacha-SerpentCS IMO `assertTrue` replace `assertEqual`.

Ex. self.assertEqual(self.contract_id.job_id.id, self.job_id.id)

hr_contract_multi_jobs/tests/test_hr_contract_multi_jobs.py
```diff
@@ -0,0 +1,78 @@
...
+        Validate the _check_one_main_job method
+        when contract has two assigned jobs
+        and check is the job_id is set as main job.
+        """
+        self.contract_id.write({'contract_job_ids':
+                                [(0, 0, {'job_id': self.job_id.id,
+                                         'is_main_job': True}),
+                                 (0, 0, {'job_id': self.job_2_id.id,
+                                         'is_main_job': False})]})
+        self.assertTrue(self.contract_id.job_id.id == self.job_id.id)
```
nikul-serpentcs: same as here

nikul-serpentcs (commented):
Improve Code

nikul-serpentcs (approved):
Code Review LGTM

Trivedi-Vacha-SerpentCS:
@pedrobaeza Commit squshed, Ready to merge

hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,48 @@
...
+    @api.depends('contract_job_ids')
+    @api.multi
+    def _compute_main_job_position(self):
+        """
+        Get the main job position from the field contract_job_ids which
+        contains one and only one record with field is_main_job == True
+        """
+        for contract in self:
+            main_job = contract.contract_job_ids.filtered('is_main_job'
+                                                          ) or False
```
espo-tony: You can remove `or False` because the filtered method will return an empty recordset if no record matches the condition. An empty recordset will be evaluated as False in the following conditional statement.

hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,48 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import UserError
+
+
+class HrContract(models.Model):
+    _inherit = 'hr.contract'
+
+    contract_job_ids = fields.One2many('hr.contract.job',
```
espo-tony: Please, break the line just after `(` in order to get a nicer and more readable indentation

hr_contract_multi_jobs/models/hr_contract_job.py
```diff
@@ -0,0 +1,27 @@
...
+
+    This model may look trivial for now, but the purpose is that other modules
+    add fields to it, e.g. a salary class or a wage scale. These are fields
+    that depend on both the contract and the job position.
+    """
+    _name = 'hr.contract.job'
+    _description = 'Relational object between contract and job'
+
+    name = fields.Char(string='Job', related='job_id.name', index=True)
+    job_id = fields.Many2one('hr.job',
```
espo-tony: Like earlier, please, break the line just after `(`

hr_contract_multi_jobs/models/hr_contract_job.py
```diff
@@ -0,0 +1,27 @@
...
+                             string='Job',
+                             required=True,
+                             ondelete='cascade')
+    contract_id = fields.Many2one('hr.contract',
```
espo-tony: Like earlier, please, break the line just after `(`

hr_contract_multi_jobs/models/hr_job.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2014 Savoir-faire Linux. All Rights Reserved.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class HrJob(models.Model):
+    _inherit = 'hr.job'
+
+    contract_job_ids = fields.One2many('hr.contract.job',
```
espo-tony: Like earlier, please, break the line just after `(`

hr_contract_multi_jobs/views/hr_contract_view.xml
```diff
@@ -0,0 +1,28 @@
...
+<odoo>
+        <!-- Inherited Contract view -->
+        <record id="hr_contract_view_form" model="ir.ui.view">
+            <field name="name">hr.contract.view.form</field>
+            <field name="model">hr.contract</field>
+            <field name="inherit_id" ref="hr_contract.hr_contract_view_form"/>
+            <field name="arch" type="xml">
+
+                <field name="job_id" position="replace">
+                    <field name="job_id" readonly="1"/>
```
espo-tony: Please, if you only wish to add a readonly attribute to this element, use XPath attributes functionality instead than remove. It should be something like:

```
<field name="job_id" position="attributes">
    <attribute name='readonly'>1</attribute>
</field>
```

espo-tony (changes requested):
Just some lesser and cosmetic notes

hr_contract_multi_jobs/models/hr_contract.py
nikul-serpentcs: @espo-tony Is not need to remove, because after @Trivedi-Vacha-SerpentCS  check `if condition`

pedrobaeza:
And we need to remove the plural on the module name.

hr_contract_multi_jobs/models/hr_contract.py
```diff
@@ -0,0 +1,49 @@
...
+    @api.depends('contract_job_ids')
+    @api.multi
+    def _compute_main_job_position(self):
+        """
+        Get the main job position from the field contract_job_ids which
+        contains one and only one record with field is_main_job == True
+        """
+        for contract in self:
+            main_job = contract.contract_job_ids.filtered('is_main_job'
+                                                          )
```
espo-tony: Please, put `)` on the same line than the rest of the statement it belongs

Trivedi-Vacha-SerpentCS:
@pedrobaeza Module Renamed Done

hr_contract_multi_job/models/hr_contract.py
```diff
@@ -0,0 +1,48 @@
...
+    @api.multi
+    def _compute_main_job_position(self):
+        """
+        Get the main job position from the field contract_job_ids which
+        contains one and only one record with field is_main_job == True
+        """
+        for contract in self:
+            main_job = contract.contract_job_ids.filtered('is_main_job')
+            if main_job:
+                main_job = main_job[0].job_id.id
```
espo-tony: According to the `api.constraint` just below, I don't think this `if` statement is required. That's because:
- We are sure there is at least 1 `main_job`, which means `if main_job` is always True;
- We are also sure there is exactly 1 `main_job`, therefore `main_job` is a recordset containing exactly 1 record, which means its fields can be directly accessed through dot-notation (no need to use `[0]` to access 1st and only element)
- If you don't explicitely invoke a `write` method, you don't necessarily need to pass the ID (an integer), you can directly pass the record.

I'd rather write it as:

```
            main_job = contract.contract_job_ids.filtered('is_main_job').job_id
            contract.job_id = main_job
```
Or even write the same thing in a single line (if you manage to stay within the max line length)

nikul-serpentcs: @espo-tony Here if not get `main_job` then ..?
if Your Conditions are False, so @Trivedi-Vacha-SerpentCS Used If a condition.

If Dev. Removed  main_job in UT that time use our scenario.

espo-tony: @nikul-serpentcs I argue that the condition is never false, because in the code there is a constraint ensuring there is always 1 and only 1 `main_job` . Am I wrong?

espo-tony (approved):
Code Review LGTM

nikul-serpentcs:
@pedrobaeza if it's Done then Move Forward

PR #401, opened 2017-11-12
[MIG] [11.0] Migrate hr_public-holidays to 11.0

feketemihai:

PR #465, opened 2018-07-10
[11.0][MIG] hr_public_holidays > hr_holidays_public

MiquelRForgeFlow:
Superseeds https://github.com/OCA/hr/pull/401.

cc @Tecnativa


MiquelRForgeFlow:
Hey everybody! something on Odoo changed that makes a test fail. Previously was all green some months ago in the commit of @feketemihai . My changes didn't affect the tests but the PR is red. I can't find what's the cause or what's wrong, so please, can someone help to fix this?

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -0,0 +1,131 @@
+# ©  2015 2011,2013 Michael Telahun Makonnen <mmakonnen@gmail.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from datetime import date
+
+from odoo import api, fields, models, _
+from odoo.exceptions import Warning as UserError
```
astirpe: Just use:
`from odoo.exceptions import UserError`

hr_holidays_public/wizards/holidays_public_next_year_wizard.py
```diff
@@ -0,0 +1,103 @@
+# Copyright 2016 Trobz
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import logging
+
+from odoo import api, models, fields, _
+from odoo.exceptions import Warning as UserError
```
astirpe: `from odoo.exceptions import UserError`

hr_holidays_public/tests/test_holidays_public.py
```diff
@@ -0,0 +1,224 @@
+# ©  2015 Salton Massally <smassally@idtlabs.sl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.exceptions import ValidationError
+from odoo.exceptions import Warning as UserError
```
astirpe: `from odoo.exceptions import UserError, ValidationError`

hr_holidays_public/views/hr_holidays_public_view.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+        <record id="view_holidays_public_tree" model="ir.ui.view">
+            <field name="name">hr.holidays.public.tree</field>
+            <field name="model">hr.holidays.public</field>
+            <field name="arch" type="xml">
+                <tree string="Public Holidays">
+                    <field name="year"/>
```
astirpe: Could you replace `<field name="year"/>` with `<field name="display_name"/>` here?

hr_holidays_public/wizards/holidays_public_next_year_wizard.xml
```diff
@@ -0,0 +1,74 @@
...
+                        Only the last templates of each country 
+                        will be taken into account (If you select templates 
+                        from 2012 and 2015 of the same country; '
+                        only the templates from 2015 will be taken into 
+                        account).
+                </div>
+                
+                <notebook>
+                    <page name="defaults" string="Defaults">
+                        <div>
```
astirpe: There are some trailing spaces in this line. Please check also the other lines.

hr_holidays_public/models/hr_holidays.py
```diff
@@ -0,0 +1,103 @@
...
+    ).astimezone(tz).replace(tzinfo=None)
+
+
+class HrHolidays(models.Model):
+    _inherit = 'hr.holidays'
+
+    show_full_days = fields.Boolean(
+        string="Show Full Days",
+        related="holiday_status_id.compute_full_days")
+    number_of_hours_temp = fields.Float(
```
astirpe: Shouldn't this module inherit from `hr_holidays_hour`?  The module `hr_holidays_hour` already defines those fields for managing the hours. There would be an overall simplification of code by inheriting from `hr_holidays_hour`, for example you could take advantage of methods for the conversion to hours, the onchanges and the views.

Migration of `hr_holidays_hour` to V11 is proposed here: https://github.com/OCA/hr/pull/450.

astirpe (changes requested):
Thank you for this PR! Please check my comments for the code review.

I personally would prefer a pure porting of this module from https://github.com/OCA/hr/tree/10.0/hr_public_holidays to V11, without adding so much other stuff. The module version =< 10 was meant to be a technical module, useful for being extended for further customization. But with the latest PRs this module is becoming something else.

The calculation of the hours should be delegated to a specific module *hr_holidays_hour*, which migration to V11 is proposed here: #450.

Odoo V12 will have the code for the leave management almost completely rewritten, supporting also the hours calculation, so do we really want to spend so much effort to rewrite the hours management in this module, with the risk to restart all over in V12 (will be released next October :open_mouth:  )?

Anyway I will not block this PR if you have different needs, just think about it.

hr_holidays_public/models/hr_holidays.py
MiquelRForgeFlow: Ok, I will wait then until that module is merged.

pedrobaeza: @astirpe I have been reviewing the module `hr_holidays_hour` and the dependency should be just the contrary: to add public holidays, we don't need to make the computation by hours first, but to adjust current Odoo core methods for obtaining the days in float. The hour one should go on top of this. I'm reviewing right now this one and proceed to merge for unlocking this, but I'm available to review any adjustment you need to make over this one for correctly pasting both together. I hope you understand.

pedrobaeza:
After been thinking about this a lot, this is the best option I have found:

* Leave `hr_holidays_public` with the public holidays definition, the `exclude_public_holidays` field in leave types, and the computation of the days intervals according this definition. This will be transparent for other modules not depending on this one, as the working intervals won't have public holidays and they are able to handle this interval days the way they want.
* Rescue `hr_holidays_compute_days` module for adding day selection helpers (full day selectors for example) and `exclude_rest_days` and `compute_full_days` options on leave types, touching the working days computation the way it's done right now in this PR (although avoiding full method overriding).
* `hr_holidays_hour` can plug together with this in 2 ways:
  * Putting a hook in this second module and depending of it for not repeating computations.
  * Being totally isolated, but repeating some of the computations.

I'm implementing this tomorrow. What do you think?

astirpe:
@pedrobaeza I would prefer to keep `hr_holidays_hour` independent from other modules, since the idea is to  keep it as a technical module that replaces the days with hours. I understand that we prefer to avoid repetition of code, but in this case I don't see any real benefit in doing that.

pedrobaeza:
OK, no problem, as stated, you can do it repeating some of the code. Just take a look for not screwing up inheritance for allowing the installation of the 3 of them.

pedrobaeza:
@mreficent I'm doing the other changes, but thanks for the lint fixup.

pedrobaeza:
I have finished the adaptation, reducing the code to the minimal and decoupling all modules from each other. I think it has been a good job :smiley: 

For handling module renaming in OpenUpgrade, there's https://github.com/OCA/OpenUpgrade/pull/1509

hr_holidays_public/views/hr_holidays_public_view.xml
```diff
@@ -0,0 +1,57 @@
...
+        </record>
+
+        <record id="view_holidays_public_form" model="ir.ui.view">
+            <field name="name">hr.holidays.public.form</field>
+            <field name="model">hr.holidays.public</field>
+            <field name="arch" type="xml">
+                <form string="Public Holidays">
+                    <group>
+                        <field name="year"/>
+                        <field name="country_id"/>
```
astirpe: We already have a module that extends this one. Would be nice if you write this block this way (to ease the extensions):

```
                <group>
                    <group name="name">
                        <field name="year"/>
                        <field name="country_id"/>
                    </group>
                    <group name="extra">
                    </group>
                </group>
```

hr_holidays_public/wizards/holidays_public_next_year_wizard.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2016 Trobz
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+
+<record id="holidays_public_next_year_wizard_view" model="ir.ui.view">
```
astirpe: Could you indent the records more to the right, making this xml consistent with the other views?

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -0,0 +1,196 @@
...
+    variable_date = fields.Boolean('Date may change', oldname='variable',
+                                   default=True)
+    state_ids = fields.Many2many(
+        'res.country.state',
+        'hr_holiday_public_state_rel',
+        'line_id',
+        'state_id',
+        'Related States'
+    )
+    calendar_leave_id = fields.Many2one(
```
astirpe: Why this field was added?

astirpe: Why this field was added?

hr_holidays_public/models/hr_holidays.py
```diff
@@ -0,0 +1,28 @@
...
+    def _get_number_of_days(self, date_from, date_to, employee_id):
+        if (self.holiday_status_id.exclude_public_holidays or
+                not self.holiday_status_id):
+            obj = self.with_context(
+                employee_id=employee_id,
+                exclude_public_holidays=True,
+            )
+        else:
+            obj = self
+        return super(HrHolidays, obj,)._get_number_of_days(
```
astirpe: Could you remove the trailing comma after `obj`?

astirpe (changes requested):
Review of module  hr_holidays_public only

pedrobaeza:
Hehe, you have just reviewed the untouched part. Checking the comments...

hr_holidays_public/models/hr_holidays_public.py
pedrobaeza: No idea...

pedrobaeza:
Changes done

astirpe:
Thanks!

We are not using `hr_holidays_compute_days`, so I think is better to let someone else (more interested) doing the functional review. I could do a code review later when I will have more time.

Module `hr_holidays_public` :+1:

pedrobaeza:
I think we are going to move forward, and if any problem, we can fill new PRs. This is done for not interfering with your module, so I don't think you will find troubles.

PR #467, opened 2018-07-25
[FIX][10.0] Take back group in view & add test because idendification_id is not mandatory

leemannd:
The field was setted with a group in standard. I ported it back.
The `t-if .... raw_value` was failing because there are records without it.

### DIFF
--- a/hr_employee_id/views/hr_employee_views.xml
+++ b/hr_employee_id/views/hr_employee_views.xml
@@ -11,7 +11,7 @@
                 <h1 position="after">
                     <label for="identification_id" class="oe_edit_only"/>
                     <h3>
-                        <field name="identification_id" readonly="1"/>
+                        <field name="identification_id" readonly="1" groups="hr.group_hr_user"/>
                     </h3>
                 </h1>
             </data>
@@ -24,7 +24,7 @@
         <field name="arch" type="xml">
             <data>
                 <field name="name" position="after">
-                    <field name="identification_id"/>
+                    <field name="identification_id" groups="hr.group_hr_user"/>
                 </field>
             </data>
         </field>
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM

leemannd:
Done

Tonow-c2c (approved):
LGTM

JayVora-SerpentCS (approved):

hr_employee_id/views/hr_employee_views.xml
```diff
@@ -45,9 +45,11 @@
         <field name="inherit_id" ref="hr.hr_kanban_view_employees"/>
         <field name="arch" type="xml">
             <li id="last_login" position="after">
-                <li t-if="record.identification_id.raw_value">
-                    <field name="identification_id"/>
-                </li>
+                <t t-if="record.identification_id">
+                    <li t-if="record.identification_id.raw_value" >
+                        <field name="identification_id" groups="hr.group_hr_user"/>
```
espo-tony, line 50: It's not better to add the `groups` option 2 lines above, into the `t t-if` tag?

espo-tony (changes requested):

hr_employee_id/views/hr_employee_views.xml
leemannd: Hum, it could be better, but I wanted to keep it as close as possible to the standard, as another module may want to update this field and it could lead to broken view/xpath.

espo-tony: I just wonder if, for users non-belonging to the given group, this would not display as a dangling dash due to an empty `<li>` tag.

grindtildeath (approved):
Can you please ~~rebase and~~ bump version number ?

leemannd:
@rafaelbn Can you tell me if it's mergeable?
If not, feel free to close it

leemannd:
Closing as it was never merged. Sidenote: has been successfully used in production

PR #469, opened 2018-08-02
[11.0] [MIG] hr_attendance_rfid

HviorForgeFlow:
Minor changes

PR #470, opened 2018-08-02
[10.0] [FIX] hr_attendance_rfid UserError catch on system + test

HviorForgeFlow:
FF if #469


hr_attendance_rfid/models/hr_employee.py
```diff
@@ -66,7 +66,5 @@ def register_attendance(self, card_code):
                 res['error_message'] = msg
                 return res
         except Exception as e:
-            msg = e.message
-            _logger.error(msg)
```
pedrobaeza, line 70: I assume you are removing this for having tests correct, but what you can do it's a `mute_logger` in the test for that purpose.

hveficent:
Aren't UserError exceptions running automatically in logs?

hr_attendance_rfid/models/hr_employee.py
HviorForgeFlow: Aren't UserError exceptions running automatically in logs?

pedrobaeza: No, the Travis status is determined if there's an `ERROR` text parsed in the log.

rafaelbn (approved):

carlosdauden (approved):

PR #471, opened 2018-08-07
[MIG] hr_public_holiday 10.0 to 11.0

rokealva83:
hr_public_holiday migrate to 11.0


pedrobaeza:
There's already another migration task in #465, and it's correctly preserved the commit history, so please review that one instead.

PR #472, opened 2018-08-07
[8.0] [FIX] rrule[0] does not always exist

thomaspaulb:


hbrunn (approved):

resource_calendar_rrule/tests/test_resource_calendar_rrule.py
```diff
@@ -19,17 +19,41 @@ def setUp(self):
...
+        hours_per_week = sum(
+            x['morning'] + x['afternoon'] for x in
+            self.calendar.simplified_attendance['data'],
         )
+        self.assertEquals(hours_per_week, 40)
+
+    def test_60_simplified_attendance_even_odd(self):
+        orig_attendance = {
+            'type': 'odd',
+            'start': fields.Date.to_string(date(2018, 10, 8)),
```
nikul-serpentcs: @thomaspaulb Here use Server Time

nikul-serpentcs (commented):
Improve Code

resource_calendar_rrule/models/resource_calendar.py
```diff
@@ -179,9 +180,18 @@ def _attendance_from_simplified(self):
             odd_even = _type == 'odd'
         result = []
         rule = self.env['resource.calendar.attendance']._default_rrule()[0]
-        start = fields.Datetime.from_string(attendance_start)
+        # convert start to UTC for SerializableRRuleSet
+        start = tz.localize(fields.Datetime.from_string(attendance_start))
```
hbrunn: what you get back here is already UTC, and `localize` doesn't do what you think it does. Consider

```python
from datetime import datetime
from pytz import utc, timezone

tz = timezone('Europe/Amsterdam')
# 12th of feburary in utc
date = datetime(2019, 2, 11, 23, 0, 0)
print date
localized_date = tz.localize(date)
print localized_date
print localized_date.astimezone(utc)

# this is the right way to do it
date = datetime(2019, 2, 11, 23, 0, 0)
print utc.localize(date).astimezone(tz)
```

hbrunn (changes requested):
I think to fix the problems you mention elsewhere, what we need here is to inject the right `tz` field for field_rrule, not messing with the dates.

resource_calendar_rrule/models/resource_calendar.py
thomaspaulb: Are you sure about that? I wrote this code assuming that `simplified_attendance['start'] and ['stop']` contain naive dates and times.

```
>>> fields.Datetime.from_string('2018-10-15')
datetime.datetime(2018, 10, 15, 0, 0)
```

In that case, `tz.localize(date)` does exactly that what I wanted to do, namely to convert this naive time into the same time but then for the user's timezone, and then converting it to UTC in order to be saved in a `SerializedRRuleset`.

I would like to save the timezone in the `SerializedRRuleset` as well, but [it always resets the tz](https://github.com/OCA/server-tools/blob/8.0/field_rrule/field_rrule.py#L33) and in many places I read that it is better to store UTC in your structures and work with timezone only for displaying. I see the sense in that, so I tried to heed it.

thomaspaulb: I now [assert](https://github.com/OCA/hr/pull/472/commits/6452c851ef8bcb8211d34e6002437f15294c62c8) that start and stop are dates and not datetimes.

PR #476, opened 2018-09-07
11.0 hr holidays public generator

yweng8111:
~~This Pull request adds 2 modules:~~
~~1. hr_holidays_public_generator~~
  ~~This module brings the shell for generating public holidays~~
  ~~based on year, country and state.~~
  ~~The module can not be used at its own and have to be extended.~~
  ~~The module adds menu "Generate Public Holidays" under "Leaves/Public Holidays".~~
  ~~From this menu one can:~~
  ~~* generate public holidays for specific country (if there is no template set)~~
  ~~* copy public holidays for specific country~~
~~2. hr_holidays_public_generator_de~~
  ~~This module extends the hr_holidays_public_generator module~~
  ~~and implements the functionality needed to generate and copy~~
  ~~the public holidays in Germany.~~

Now the wizard is integrated on `hr_holidays_public`.


pedrobaeza:
`hr_holidays_public` already includes a wizard for generating public holidays on the future:

https://github.com/OCA/hr/blob/11.0/hr_holidays_public/wizards/holidays_public_next_year_wizard.py

Why is this needed?

ntodorova:
The OCA wizard actually does not generate holidays, but copy the holidays from the last year template, or am I wrong? The problem comes when we have holidays with different date each year, for example Easter, then one should manually change the date. The new module adds functionality to generate the Public Holidays for Germany and a base functionality  to be extended from other countries.

hr_holidays_public_generator/wizard/hr_holidays_public_generator_view.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
nikul-serpentcs: @yweng8111  IMHO Replace to `<odoo>`

pedrobaeza:
The holidays generator should be inside `hr_holidays_public`, and the German specific part in OCA/l10n-germany.

ntodorova:
@pedrobaeza, I just have a short question - do you mean that we don`t need new module "hr_holidays_public_generator", but we should integrate the functionality in the  "hr_holidays_public"?

pedrobaeza:
Yes, Nikolina. I think it can be acceptable to have that part inside the main module, as a complement of the other generator (even as an option inside it).

hr_holidays_public_generator/wizard/hr_holidays_public_generator_view.xml
yweng8111: Thanks for the review. I will fix it now.

yweng8111:
@pedrobaeza I have just finished moving the wizard into module hr_holidays_public. If it is merged, I will create an new PR in OCA/l10n-germany for the German implementation.

ntodorova:
Hi @pedrobaeza, are you going to have some time for a review here? : ))

gurneyalex (approved):

gurneyalex:
@yweng8111 @ntodorova  do you have a link to the PR implementing the german holidays which works with this one?

hr_holidays_public/wizards/hr_holidays_public_generator.py
```diff
@@ -0,0 +1,85 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from datetime import datetime
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+COUNTRY_GENERATORS = []
+
+
+class HrHolidaysPublicGenerator(models.TransientModel):
```
gurneyalex, line 12: I think the computation of easter sunday should be in this class : otherwise most western european localizations will end up recomputing it.

yweng8111:
> @yweng8111 @ntodorova do you have a link to the PR implementing the german holidays which works with this one?

It can only be done, if this PR is approved and the commits are merged. That implementation depends on these changes.

pedrobaeza:
@yweng8111 you can already submit the PR, although this PR is not merged. You can even include this branch reference in oca_dependencies.txt file, and then everything will work (Travis, Runbot...).

hr_holidays_public/wizards/hr_holidays_public_generator.py
yweng8111: Yes, It's a good Idea. I will try to improve this wizard in next PR.

yweng8111:
The PR for German implementation was created.
https://github.com/OCA/l10n-germany/pull/37

HaGuesto (approved):
LGTM

I reviewed the code and check the corresponding German generator.

mharenz:
I reviewed the code and found no problems.

mharenz (approved):
I reviewed the code and found no problems.

yweng8111:
Hi @pedrobaeza,
This PR has already the Label approved. When can it be merged? Or there are something still incorrect.

PR #478, opened 2018-09-22
hr_holidays_hour: fix HrHolidaysRemainingLeavesUser VIEW query

CGenie:
1. It wasn't checked whether the query was patched already or not
2. Python's string.replace is case-sensitive. Newer Odoo versions did change
   cases of query parts ('as' -> 'AS') and so the code didn't fully apply
   new query which resulted in a subtle bug of 'employee_id' column missing.

### DIFF
--- a/hr_holidays_hour/report/hr_holidays_report.py
+++ b/hr_holidays_hour/report/hr_holidays_report.py
@@ -59,5 +59,7 @@ class HrHolidaysRemainingLeavesUser(models.Model):
             view_def += self._holidays_hour_group_by()
         # Re-create view
         tools.drop_view_if_exists(cr, self._table)
-        sql = 'CREATE OR REPLACE VIEW {} as (%s)' % view_def
-        cr.execute(SQL(sql).format(Identifier(self._table)))
+        sql = SQL('CREATE OR REPLACE VIEW {} as (%s)' % view_def).format(
+            Identifier(self._table)
+        )
+        cr.execute(sql)
--- a/hr_holidays_hour/report/hr_holidays_report.py
+++ b/hr_holidays_hour/report/hr_holidays_report.py
@@ -1,6 +1,8 @@
 # Copyright 2017 Onestein (<http://www.onestein.eu>)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+import re
+
 from odoo import fields, models, tools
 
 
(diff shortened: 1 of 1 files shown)


astirpe:
@CGenie Thanks!
Could you check the failures in Travis ?

CGenie:
It's some SQL injection possibility, isn't directly related to my code but I can try to fix this.

astirpe:
You are adding `psycopg` as an external dependency, but Odoo uses `psycopg2`.

I think you can get tips for fixing the travis warnings by following the the [no-sql-injection](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#no-sql-injection) paragraph of the OCA Guidelines.

CGenie:
I'll look into why `psycopg` is there instead of `psycopg2`. As for SQL injection -- this code was there already and actually it's hard to avoid. We generate a view and there is just modification of the view code there. I would have to parse the view with some SQL parser, then modify appropriate objects, then add into SQL query. I already fixed this a bit by using `psycopg.sql.Identifier` for table name, however it's difficult to do the same for view code.

CGenie:
I surely don't understand how my code is worse than mainstream's
https://github.com/OCA/hr/blob/11.0/hr_holidays_hour/report/hr_holidays_report.py#L50-L52
The build on `master` passes but mine does not...

hr_holidays_hour/report/hr_holidays_report.py
```diff
@@ -38,15 +40,32 @@ def init(self):
...
             view_def = view_def[:-1]
-        view_def += self._holidays_hour_group_by()
+        holidays_hour_group_by = self._holidays_hour_group_by()
+        if holidays_hour_group_by not in view_def:
+            view_def += self._holidays_hour_group_by()
         # Re-create view
         tools.drop_view_if_exists(cr, self._table)
-        cr.execute("create or replace view {} as ({})".format(
-            self._table, view_def,
+        sql = SQL('CREATE OR REPLACE VIEW {} as {}')
```
pedrobaeza: Using `AsIs` method you can do it directly. See for example: https://github.com/OCA/OpenUpgrade/pull/1589/files#diff-438d8381a8d27b194556dd3c7cbbe766R140

CGenie:
@astirpe @pedrobaeza Is it possible to merge this now? :)

astirpe (approved):

pedrobaeza (approved):

Issue #484, opened 2018-10-17
[8.0][hr_employee_firstname] Modifying inherited view using '<label>' does not work


gabrielo77:
When I try to install the module the inheritance of view __hr.view_employee_form__ is broken because the plan '<label>' does not work, it has to be changed to an '<xpath>'. See:

https://github.com/OCA/hr/blob/8.0/hr_employee_firstname/views/hr_view.xml#L9

PR #485, opened 2018-10-17
[8.0] Two small fixes in hr_employee_firstname module (fixes #483 and #484)

### LINKED ISSUE #483: [8.0][hr_employee_firstname] Field address_home_id has changed to address_id
If you take a look at this line:

https://github.com/OCA/hr/blob/8.0/hr_employee_firstname/models/hr_employee.py#L57

You can see `mapped()` using the field `address_home_id` on *res.employee* which has been renamed to `address_id` since v8.0. This should be fixed as it raises an error when `_update_partner_firstname()` is called.

gabrielo77:
I found this two issues: #483 and #484.

This *PR* solves them both.

Thank you!


gurneyalex (approved):

hr_employee_firstname/views/hr_view.xml
```diff
@@ -6,9 +6,9 @@
         <field name="model">hr.employee</field>
         <field name="inherit_id" ref="hr.view_employee_form"/>
         <field name="arch" type="xml">
-            <label for="name" position="replace">
+            <xpath expr="//label[@for='name']" position="replace">
```
gurneyalex, line 10: if you use 'replace' you need to change the priority of the view definition to 100 or more to avoid breaking other views depending on the label for="name" anchor.

gurneyalex (changes requested):

gabrielo77:
@*gurneyalex you're right. What do you suggest? *

--
Gabriel

hr_employee_firstname/views/hr_view.xml
gabrielo77: @gurneyalex you're right. The 'replace' position was there before my **PR** actually. What do you suggest to correct this?

gurneyalex: change the priority of the view (this field: https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/ir/ir_ui_view.py#L126)

hr_employee_firstname/views/hr_view.xml
```diff
@@ -6,9 +6,9 @@
         <field name="model">hr.employee</field>
```
gurneyalex, line 6:
```suggestion
        <field name="model">hr.employee</field>
        <field name="priority" eval="100"/>
```

hr_employee_firstname/views/hr_view.xml
gurneyalex: see suggestion below

PR #458, opened 2018-07-03
[MIG][11.0]hr_employee_age:Migration to 11.0.

Trivedi-Vacha-SerpentCS:
[Mig][11.0]hr_employee_age:Migration to 11.0.

PR #489, opened 2018-10-25
[12.0][ADD] hr_employee_relatives: alternative to hr_family

alexey-pelykh:
Alternative to #487 in order to avoid painful migration of `hr_family@11.0` to `hr_family@12.0`


pedrobaeza:
On this one that is new, I encourage you to not use plurals in module name. `hr_employee_relative` is OK for a technical name.

alexey-pelykh:
Thanks, slipped my mind and fixed

pedrobaeza:
Thanks, I wish all the new contributors were so reactive and mind-opened as you :smiley:

hr_employee_relative/models/hr_employee_relative.py
```diff
@@ -0,0 +1,59 @@
...
+
+class HrEmployeeRelative(models.Model):
+    _name = 'hr.employee.relative'
+    _description = 'HR Employee Relative'
+
+    employee_id = fields.Many2one(
+        string='Employee',
+        comodel_name='hr.employee',
+    )
+    relation = fields.Selection(
```
nikul-serpentcs: @alexey-pelykh @pedrobaeza  IMO here add one M2O field for `Relation`, Because a user can be adding and maintain an easily `Relation`.
Selection field give only a few data, so

hr_employee_relative/tests/test_hr_employee_relatives.py
```diff
@@ -0,0 +1,29 @@
...
+            'relatives': [
+                (0, 0, {
+                    'relation': 'other',
+                    'name': 'Relative',
+                    'date_of_birth': datetime.now() + relativedelta(years=-42),
+                })
+            ],
+        })
+        relative = self.EmployeeRelative.browse(employee.relatives[0].id)
+        self.assertEqual(int(relative.age), 42)
```
nikul-serpentcs, line 31: @alexey-pelykh here static age, next year it's changed, IMO here change like this
`age = relativedelta(datetime.now(), relative.date_of_birth)`
`record.age = age.years + (age.months / 12)`

`assertEqual(relative.age, age)`

alexey-pelykh: Not sure what you mean since 42 years are added to the test data initially, thus resulting the same. And copying code from the tested method itself has to meaning to me

nikul-serpentcs: > thus resulting the same.

Yes @alexey-pelykh, Because this yesr `age` is same but next year different or increasing one.
So you can use Compute code here and check assertEqual.

hr_employee_relative/models/hr_employee_relative.py
alexey-pelykh: 1. Won't that be an overkill? Third opinion needed
2. Would adding default set to M2O field be good idea?

hr_employee_relative/tests/test_hr_employee_relatives.py
alexey-pelykh: Still don't get it, every year age will be 42 due to `'date_of_birth': datetime.now() + relativedelta(years=-42),`

nikul-serpentcs: @alexey-pelykh Could you please check this link https://github.com/OCA/hr/pull/458#discussion_r221875110

here Calculation are Done by me.

alexey-pelykh: Still not applicable. During creation of test data the relative is given _exactly_ 42 years of age by subtracting 42 from current date:
```
'date_of_birth': datetime.now() + relativedelta(years=-42),
```
Couple of milliseconds later it's checked if calculation returns 42, effectively checking if `relative.age` is computed as 42. Checking if `relative.age` equals to anything but constant dismisses the meaning of the test. The only change that would make sense is to put 42 in a separate variable to avoid it being a magic constant twice, yet for a such small piece of code it's not that relevant.

hr_employee_relative/models/hr_employee_relative.py
nikul-serpentcs: @pedrobaeza @tafaRU @Yajo Could you please check and suggest here.

CC @alexey-pelykh

pedrobaeza: You can provide as data usual relation types, but it's good to have the extensibility through a table. This is more and more becoming my initial idea of partner_multi_relation with employees :wink:

alexey-pelykh: @pedrobaeza how do you suggest to prepopulate this table for common usecases?

pedrobaeza: That's done with XML or CSV data (first is better in this case). Examples: 

https://github.com/odoo/odoo/blob/12.0/addons/account/data/data_account_type.xml
https://github.com/odoo/odoo/blob/12.0/addons/stock/data/stock_data.xml

alexey-pelykh: Maybe just make it a char field?

pedrobaeza: Well, I prefer selection over char for grouping and so on.

alexey-pelykh: and configure list of available relations in configuration?

alexey-pelykh: done!

nikul-serpentcs: Thanks, @alexey-pelykh @pedrobaeza for added a new feature :smile:

hr_employee_relative/models/hr_employee.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    relatives = fields.One2many(
```
yajo: `relative_ids`, as per guidelines.

hr_employee_relative/models/hr_employee.py
```diff
@@ -0,0 +1,14 @@
...
+        string='Relatives',
+        comodel_name='hr.employee.relative',
```
yajo, line 12: Would it make sense to do a many2many to `res.partner` instead? That table already contains all possibly required fields to identify somebody, and addons that add the missing ones you need...

hr_employee_relative/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
```
yajo, line 2: It's common nowadays to remove licensing and copyright from `__init__.py` files that amount just a few imports.

hr_employee_relative/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
yajo, line 1: :fire:

hr_employee_relative/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
yajo, line 1: :fire:

yajo (changes requested):

hr_employee_relative/models/hr_employee.py
pedrobaeza: You need to indicate also the type of relation, so  you need a new model.

hr_employee_relative/__init__.py
pedrobaeza: Copyright is usually removed, but not license, which should be in all files as stated by the AGPL license.

hr_employee_relative/models/hr_employee.py
alexey-pelykh: Changed

hr_employee_relative/models/hr_employee.py
alexey-pelykh: It's not a `res.partner` in a matter of speaking, especially when it comes to notes like "what gifts relatives prefer" if such policies exist in company, plus relation type is required

pedrobaeza: @alexey-pelykh any way, it would be great to have a link to res.partner in that one2many table for relatives with persons.

yajo: OK I understand the relation stuff, so maybe the question should have gone inside the `name` field of `hr.employee.relative`. Would it make sense to link there to a partner instead?

alexey-pelykh: I still think that creating a partner in order to enter a name is an over-complication. And how to limit access to that partner only to HR users?

pedrobaeza: You can do both:

- To have an string for direct relations
- To have a partner_id link, with an onchange that dumps the name to `name` field, for advanced uses. If you put contact type as "private", it will only be seen by HR managers or the affected person (through portal).

alexey-pelykh: @pedrobaeza a code snipped where similar approach is used would be very helpful

pedrobaeza:
```python
class HrEmployeeRelative(models.Model):
    ...
    name = fields.Char()
    partner_id = fields.Many2one()

    @api.onchange('partner_id')
    def _onchange_partner_id(self):
        if self.partner_id:
            self.name = self.partner_id.name
```

alexey-pelykh: Done, please check if I got your idea correctly

hr_employee_relative/views/hr_employee_relative.xml
```diff
@@ -0,0 +1,23 @@
...
+      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+    -->
+
+    <record id="hr_employee_relative_view_tree" model="ir.ui.view">
+        <field name="name">hr.employee.relative.view.tree</field>
+        <field name="model">hr.employee.relative</field>
+        <field name="arch" type="xml">
+            <tree string="Employee Relatives" editable="bottom">
+                <field name="relation_id"/>
+                <field name="name"/>
```
yajo, line 14: You forgot to add `partner_id` here.

yajo (approved):
Code OK

hr_employee_relative/views/hr_employee_relative.xml
alexey-pelykh: Thanks, true! Fixed

alexey-pelykh:
@nikul-serpentcs please review the changes requested here as well

nikul-serpentcs (approved):
Code Review LGTM

alexey-pelykh:
@pedrobaeza there's also a suggestion to hide 
* Spouse Complete Name
* Spouse Birthdate
* Number of Children

if this module is installed, yet I'm not sure

pedrobaeza:
Well, it seems redundant information, so I would say yes.

alexey-pelykh:
Another question: are there an easy way to add a self-service feature when `employee.user_id` (Related User that is) can also edit this data?

alexey-pelykh:
@luistorresm I'd like to request your review as well

gurneyalex (approved):

hr_employee_relative/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows storing information about employee's family.
```
gurneyalex, line 1: Can you add a note about hr_family which provides a similar feature with a different implementation?

alexey-pelykh: Absolutely, what text/link should I add?

hr_employee_relative/data/data_relative_relation.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <data noupdate="1">
+        <!--
+          Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+          License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+        -->
+
+        <!-- hr.employee.relative.relation -->
+        <record model="hr.employee.relative.relation" id="relation_spouse">
```
luistorresm, line 10: Odoo already has the field to set spouse information when the employee is married.

We could use that field when is the spouse?

alexey-pelykh: Please elaborate

alexey-pelykh:
@pedrobaeza seems this is good to be merged and 5 days after approval have passed?

hr_employee_relative/readme/DESCRIPTION.rst
pedrobaeza: But why maintaining `hr_family` if this one is more complete?

hr_employee_relative/data/data_relative_relation.xml
pedrobaeza: @luistorresm any news about this?

pedrobaeza:
I have commented the 2 opened comments to see if we solve them and if not replied in 2 days, please ping me again for merging.

hr_employee_relative/data/data_relative_relation.xml
alexey-pelykh: The only thing I can think of is to compute that field depending if there's a valid relation, yet not sure if that totally OK, or that should go to a separate module

luistorresm: I propose to avoid load the spouse data in this new feature,

And for that information use the Odoo fields,

What do you think?

Regards

alexey-pelykh: That would be 100% confusing for end-user. I'd hide the spouse UI field and set it to be related to data stored here in order to maintain compatibility with other modules that expect that field to be filled in with valuable data

alexey-pelykh: Actually, I gave that some thought - that particular feature should go to some `hr_employee_relatives_compat` or `hr_employee_relatives_spouse`

alexey-pelykh:
@pedrobaeza half-replied that is

pedrobaeza:
OK, let's merge it as it seems good enough.

PR #497, opened 2018-10-30
[12.0][ADD] hr_contract_currency: non-default currency for contract

alexey-pelykh:
image


hr_contract_currency/models/hr_contract.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrContract(models.Model):
+    _inherit = 'hr.contract'
+
+    currency_id = fields.Many2one(
```
alexey-pelykh, line 10: @pedrobaeza is this correct way of converting a related field to be non-related?

alexey-pelykh: another question: for some reason, changes on this field are not "tracked" in the bottom of the screen:
image
a hint would be much appreciated

pedrobaeza: Yes, although changing fields this way sometimes can be a problem with other modules. Let's hope it isn't in this case. For having that changes, you have to add `track_visibility` argument

alexey-pelykh: Awesome, thanks!

alexey-pelykh:
This one will be simple and separate from hr_payroll_currency

hr_contract_currency/views/hr_contract.xml
```diff
@@ -0,0 +1,30 @@
...
+            </xpath>
+        </field>
+    </record>
+
+    <record id="hr_contract_view_tree" model="ir.ui.view">
+        <field name="name">hr.contract.view.tree.inherit.currency</field>
+        <field name="model">hr.contract</field>
+        <field name="inherit_id" ref="hr_contract.hr_contract_view_tree"/>
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='wage']" position="after">
```
nikul-serpentcs: @alexey-pelykh use field directly

nikul-serpentcs (approved):
Minor change Otherwise LGTM

hr_contract_currency/views/hr_contract.xml
alexey-pelykh: Thanks, updated!

alexey-pelykh:
@feketemihai @mreficent I'd like to request your reviews as well, if you don't mind

AaronHForgeFlow:
Is the currency field passed to payslips? or this is just info?

alexey-pelykh:
@aheficent the field is there initially:
https://github.com/odoo/odoo/blob/9dbb7d199cfe085ea5e3a4980de2bd766953ab93/addons/hr_contract/models/hr_contract.py#L79
but it's set as being mirror of default company currency.

The `hr_payroll` does not include currency or conversion at all, to answer your question directly - that's something different to address:
https://github.com/odoo/odoo/blob/9c3ec026c541901a9fe75db9bca5f9850276cb41/addons/hr_payroll/data/hr_payroll_data.xml#L46

AaronHForgeFlow (approved):
:+1: Thanks for the explanation. LGTM.

PR #499, opened 2018-10-30
[FIX] hr_holidays_leave_auto_approve tracking handling

### LINKED ISSUE #498: hr_holidays_auto_approve: overwrites tracking_disable=True
If I create a holiday with tracking_disable = True and a holiday type with auto approve false, I expect that no emails are going to be sent. However this line:

https://github.com/OCA/hr/blob/10.0/hr_holidays_leave_auto_approve/models/hr_holidays.py#L21

bypasses the value I put in the context.

gurneyalex:
If tracking is disabled through the context, the module should not
reenable it.

closes #498


leemannd (approved):

yvaucher (approved):

yvaucher:
Merging as already approved by 3 person on version 11.0

PR #501, opened 2018-10-31
[12.0][ADD] hr_holidays_accrual_advanced: Advanced accrual leave allocations

alexey-pelykh:
image
image

PR #502, opened 2018-11-01
[12.0][ADD] hr_employee_service: employee service information & duration

alexey-pelykh:
image


hr_employee_service/models/hr_employee.py
```diff
@@ -0,0 +1,58 @@
...
+    service_start_date = fields.Date(
+        string='Start Date',
+        groups='hr.group_hr_user',
+        track_visibility='onchange',
+        help=(
+            'Start date is the first day the employee actually works and'
+            ' this date is used for accrual leave allocations calculation'
+        ),
+    )
+    service_termination_date = fields.Date(
```
dreispt, line 32: Aren' these a duplication on the Contract start and end dates?
At least the service start duration should hire date, if no service start date is provided.

hr_employee_service/models/hr_employee.py
```diff
@@ -0,0 +1,58 @@
...
+        help=(
+            'Termination date is the last day the employee actually works and'
+            ' this date is used for accrual leave allocations calculation'
+        ),
+    )
+    service_duration = fields.Float(
+        string='Service Duration',
+        digits=(2, 1),
+        readonly=True,
+        compute='_compute_service_duration',
```
dreispt, line 45: Explain to the user that this is months.

hr_employee_service/models/hr_employee.py
alexey-pelykh: Not quite, since Contracts may be expired, renewed, prolonged etc., thus not a reliable source of information. Difference between hire and start date is as well significant - hire date is when paperwork was actually finished (offer accepted and signed e.g.) while start date may not yet be defined, thus assuming if there's no start date yet there's hire date then hire is start - is a bit long-shot.

hr_employee_service/models/hr_employee.py
alexey-pelykh: Using `help`? And actually these are years

hr_employee_service/models/hr_employee.py
dreispt: Yes contracts can be renewed, and contract start date is not the same as hire date.
But the date should be the first day of work, and the start date of the first contract.
If you need e date for when the paperwork was finished, that should be a different field.

alexey-pelykh: First contract date - that would work except for the re-hire cases

alexey-pelykh: Regrading date when paperwork is done - that's commonly referred to as a hire date, so not sure if I got you correctly. (I did some googling on the matter)

hr_employee_service/models/hr_employee.py
dreispt: OK. Using help would be fine, but using the String label would be better.

I don't want to complicate your design, but in the public sector I've seen service times represented in years, months and days (they use it for career progression). Suggestion: maybe an additional computed field with a text description of the service time.

dreispt: s/"Service Duration"/"Service Years" ?

alexey-pelykh: Better, actually I liked your idea about years/months/days

alexey-pelykh: Done! Thanks for the idea!

hr_employee_service/models/hr_employee.py
```diff
@@ -0,0 +1,93 @@
...
+        readonly=True,
+        compute='_compute_service_duration_display',
+    )
+
+    @api.multi
+    @api.depends('service_start_date', 'service_termination_date')
+    def _compute_service_duration(self):
+        for record in self:
+            if record.service_start_date:
+                service_since = record.service_start_date
```
dreispt, line 76: Can we default to hire date if not service start date is given?

hr_employee_service/models/hr_employee.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @api.multi
+    @api.depends('service_start_date', 'service_termination_date')
+    def _compute_service_duration_display(self):
+        for record in self:
+            if record.service_duration:
+                service_duration = relativedelta(
+                    record.service_termination_date or fields.Date.today(),
+                    record.service_start_date
+                )
```
dreispt, line 96: I can see the reason why, but it's not nice to duplicate the service duration computation here.
I suggest this design:
- `_compute_service_duration` becomes a `_get_service_duration_timedelta` helper method, returning a timedelta (making it easy to extend the basic computation - I've seen tricky rules to handle 28/29/31 day months).
- the `_compute_service_duration` calls the `_get_service_duration_timedelta` helper and assigns value to all four computed fields.

hr_employee_service/views/hr_employee.xml
```diff
@@ -0,0 +1,36 @@
...
+        <field name="model">hr.employee</field>
+        <field name="inherit_id" ref="hr.view_employee_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//page[@name='hr_settings']/group" position="inside">
+                <group name="service_information" string="Service Information">
+                    <field name="service_hire_date"/>
+                    <field name="service_start_date"/>
+                    <field name="service_termination_date"/>
+
+                    <field name="service_duration" invisible="1"/>
```
dreispt, line 19: 👍

hr_employee_service/models/hr_employee.py
alexey-pelykh: Well, I'd avoid that since case when offer is signed yet start date is negotiated won't be handled properly. What are your thoughts on that?

hr_employee_service/models/hr_employee.py
alexey-pelykh: Got your idea! Yet not sure if it's better to proceed with `relativedelta` or with `timedelta`, what's your vision on that? First one would give super exact as far as I understand, yet does not provide the "total days" option

hr_employee_service/models/hr_employee.py
dreispt: Start date and Hire date are the same thing, that's what I've seen in all HRMSs I worked with.
Contract signed date is something else.

hr_employee_service/models/hr_employee.py
dreispt: IIIC timedelta is the object resulting from the relativedelta operation.
The point is that the helper function return a delta object, not usable directly by the ORM, but to be used by the actual _compute method.

hr_employee_service/models/hr_employee.py
dreispt: (there is still a point to give the possibility to manually give a "service start date" to use for service time. For example, if the worker was transferred from a related company/organization, and is entitled to keep his service time since the hire date in the origin organization).

hr_employee_service/models/hr_employee.py
alexey-pelykh: Timedelta has days and below (hours, minutes) and total_seconds(), while relativedelta gives exact years, months and days, the actual thing that is. So there's really to different calculations. I understand what you want to achieve yet still don't get how to accomplish that taking into account the different of calculation

hr_employee_service/models/hr_employee.py
alexey-pelykh: I'd still keep those two fields separate, while maybe prefilling in the UI the Start Date when Hire Date is altered and Start Date is empty. Can we call that a solution?

dreispt: It is, but I think it's easier to add an "or" to the computation start date, like you did for the termination date.

hr_employee_service/models/hr_employee.py
dreispt: I see. 
Is it reasonable for you to compute `service_duration` from the years/months/days fields? 
If in years (as you originally had): = years + months/12 + days/360

alexey-pelykh: Or vice-versa as years * 365 + months * 30 + days
Personally, I'd keep as is since days don't loose any precision, leaving for other module consumers the decision how to use that data while those 3 extra fields are for UI purpose only

hr_employee_service/models/hr_employee.py
alexey-pelykh: Easier yet not transparent for the end-user (who does not read the code?) - let's try the via-UI approach

alexey-pelykh:
@luistorresm I'd like to request your review as well

hr_employee_service/README.rst
```diff
@@ -0,0 +1,73 @@
...
+.. |badge4| image:: https://img.shields.io/badge/weblate-Translate%20me-F47D42.png
+    :target: https://translation.odoo-community.org/projects/hr-12-0/hr-12-0-hr_employee_service
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/116/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module adds employee service information.
```
luistorresm: Hello @alexey-pelykh 

Could you add more information to this README file?

Which is the reason and maybe a use case where is used this feature?

alexey-pelykh:
Does anyone have a hint about those 2 new checks that are failing?

alexey-pelykh:
@dreispt what are your thoughts on the latest code version?

alexey-pelykh:
@pedrobaeza I know I shouldn't summon you, yet should this go to the 12.0 milestone as possible replacement of https://github.com/OCA/hr/tree/10.0/hr_employee_seniority ?

pedrobaeza:
Milestone set, although this doesn't mean this replaces any module.

alexey-pelykh:
(the codecov checks - what are those? why those triggered on this particular PR?)

pedrobaeza:
It's always triggered. It shows you the current code covered by the execution of install/tests and it's a metrics of how much untested code there is.

alexey-pelykh:
Fixed, should be good for review now

luistorresm (changes requested):

hr_employee_service/README.rst
alexey-pelykh: @luismontalba updated the readme a bit

alexey-pelykh:
@luistorresm #501 is related and based on this, if you could look there as well it would be awesome

hr_employee_service/README.rst
pedrobaeza: Incorrect pong :wink:

alexey-pelykh: indeed :) @luistorresm ^

luistorresm: A last question @alexey-pelykh 

Why you do not use the dates in the contract?

alexey-pelykh: Modularity that is: #506

alexey-pelykh: In other words, I don't want to enforce contract usage if Odoo is not used for _everything_

luistorresm (approved):

alexey-pelykh:
@dreispt your word?

alexey-pelykh:
@nikul-serpentcs and your review please 👍

nikul-serpentcs (approved):
Code Review LGTM

pedrobaeza:
@dreispt as you had some objections, is not OK for you to merge?

alexey-pelykh:
@pedrobaeza what are next steps on this?

pedrobaeza:
As @dreispt is not answering, let's merge it and he can later recall.

PR #505, opened 2018-11-07
[12.0][FIX] hr_employee_ssn: fix author naming

alexey-pelykh:


hr_employee_ssn/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     'category': 'Human Resources',
     'website': 'https://github.com/OCA/hr',
     'author':
-        'Braibean Apps, '
+        'Brainbean Apps, '
```
rvalyi, line 10: could you also move the comma out of the quoted string?

rvalyi (changes requested):

hr_employee_ssn/__manifest__.py
alexey-pelykh: Sure, what's the guidelines for this field?

alexey-pelykh: Since this one has a comma https://github.com/OCA/maintainer-tools/blob/646608c2916cb30aa2325fc292aaf69b5ab993b5/template/module/__manifest__.py#L12

pedrobaeza (commented):
Comma is correct as is. It's part of the string.

hr_employee_ssn/__manifest__.py
pedrobaeza: Comma is correct as is. It's part of the string.

pedrobaeza:
Merging as trivial

hr_employee_ssn/__manifest__.py
rvalyi: my bad, the syntax was correct indeed sorry I forgot author was a single string.

PR #508, opened 2018-11-09
[10.0][IMP] Refactor create_next_year_public_holidays wizard

grindtildeath:
This small refactoring allows to keep better track of what is created
by preparing all the values before calling the create method,
instead of doing copies of the records sequentially.


gurneyalex (approved):
can you forward port the change to 11.0 (if the module was migrated)?

pedrobaeza:
Please check #476 that improves the holidays generator in 11.0.

dreispt (approved):

PR #509, opened 2018-11-13
[ADD][10.0]hr_attendance_reason

AaronHForgeFlow:
Odoo removed this feature from v9 onwards. IMO this feature is still useful. Employees may have some issues that make them not to register their attendance. With this feature the can notify the HR manager. The manager will set the reason of the manual attendance.

PR #510, opened 2018-11-13
[10.0][ADD]hr_attendance_autoclose

AaronHForgeFlow:
This module close stale attendances automatically. Without this module, when an employee forgets to check out at the end on the day and the next day the employee does not realize of that, then the error is propagated and all the attendances are wrong. With this module only the attendance with the issue is
wrong, and the manager knows the system closes the attendance, not the employee


pedrobaeza:
I think you should check my module in #421 for avoiding all of this stuff for filling missing attendances on leaves, worked hours, and so on. cc @jbeficent

AaronHForgeFlow:
Thanks @pedrobaeza  I will check. Is that report compatible with [hr_timesheet_holiday](https://github.com/OCA/hr-timesheet/commits/10.0/hr_timesheet_holiday)? 

The total hours are computed in the timesheet but not in the attendance itself. I can use this value to auto close stale attendances for example. If nobody is interested I'll close this.

pedrobaeza:
@aheficent with my module, you should uninstall that module, because you don't need that dirty trick of adding attendances when there are leaves, so we can say that they are incompatible, as if you create that fake attendances, they were taken as worked hours when they are not.

hr_attendance_worked_hour/tests/test_hr_attendance_worked_hour.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def setUp(self):
+        super(TestHrAttendanceReason, self).setUp()
+        self.att_model = self.env['hr.attendance']
+        self.employee = self.browse_ref('hr.employee_al')
+
+    def test_employee_edit(self):
+        att = self.att_model.create(
+            {'employee_id': self.employee.id,
+             'check_in': time.strftime("%Y-%m-01 07:%M:%S"),
```
nikul-serpentcs: @aheficent IMO here use `DEFAULT_SERVER_DATETIME_FORMAT`

nikul-serpentcs (commented):
minor change

hr_attendance_worked_hour/tests/test_hr_attendance_worked_hour.py
AaronHForgeFlow: Thanks @nikul-serpentcs. Done

pedrobaeza:
But I don't get why do you want to add this field when there's another field for the same purpose:

https://github.com/odoo/odoo/blob/8cdfe1d8fe09b4f5bf808fc38ac812780e807c40/addons/hr_attendance/models/hr_attendance.py#L23

AaronHForgeFlow:
Now the field is empty until the check out is performed. I redefine the field so it is not stored and it is computed any time. Then, we compute the hours so we are able to see the worked hours before the check out. We can use this value on opened attendances to perform other actions as automatic closing of stale attendances or send a message to the manager.

Making that field not stored may cause problems when grouping by that field but I have not see that in any module. Perhaps it may cause other problems but I have not noticed that.

pedrobaeza:
I think the approach of this is wrong because it creates false expectations on the shown value. I prefer to have it empty until both sign in and sign out are done. If this goes ahead, please specify very clear the intention of the module and the known issues.

AaronHForgeFlow:
I agree the readme could be more clear.

That approach will include the use of #509 so the manager knows the attendance had an issue by the "attendance reason". I think that is better than coming the next day and having an attendance of many hours. Even worse, in the case the next day I have a visit to a customer or I just simply forget to check in again then the situation gets worse.

pedrobaeza:
I don't follow your thoughts. If you don't sign out, next day the signing will be out instead of in, but you will have the same problem with this. You will have to manually adjust the attendance and ask for the real sign out hour...

AaronHForgeFlow:
Yes, but if the next day I don't realize I didn't check out yesterday, I will be performing the check out in the morning. And If I don't realize of that, the error is propagated over the next days. With the new approach only one attendance is wrong.

pedrobaeza:
I think you are wrong, as the system doesn't compensate that. You will still be in signed in state, so you when you sign, it will be a sign out.

AaronHForgeFlow:
This module does not compensate that. But allows another module to use this value to perform the action.

pedrobaeza:
Then I think everything should be packed in only one module, as this value doesn't serve for other thing.

AaronHForgeFlow:
Well, I thought maybe other modules can use this one as base, for example to send a message to the user when the worked hours are higher than expected. But yes, the feature to autoclose stale attendances is the most important reason of this module, I will include the changes here.

hr_attendance_autoclose/README.rst
```diff
@@ -0,0 +1,90 @@
...
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+#. Go to *Attendances > Manage Attedances > Attendances*.
+#. Go to *Attendances > Manage Attedances > Attendances*.
```
pedrobaeza: Duplicated text

hr_attendance_autoclose/README.rst
```diff
@@ -0,0 +1,90 @@
...
+#. Attendance are autoclosed after the value specified.
+
+Known issues / Roadmap
+======================
+
+* Known Issues
+
+This module overwrites the standard behavior of worked_hours field in
```
pedrobaeza: I think there's no overlap as the name of the field is different. Any way, I would say to use a more explicit name for the field.

hr_attendance_autoclose/README.rst
```diff
@@ -0,0 +1,90 @@
...
+=====
+
+#. Go to *Attendances > Manage Attedances > Attendances*.
+#. Go to *Attendances > Manage Attedances > Attendances*.
+#. Attendance are autoclosed after the value specified.
+
+Known issues / Roadmap
+======================
+
+* Known Issues
```
pedrobaeza: Not needed text

pedrobaeza (changes requested):
This is better!

hr_attendance_autoclose/README.rst
AaronHForgeFlow: Actually I am using the same field now. I think it is no need for duplicating the field. Is it?

AaronHForgeFlow: Making the standard field not stored was wrong so I go back to the previous approach and use a better field name.

bjeficent (changes requested):
It's not working properly. The attendances are not closed after cron is executed

hr_attendance_autoclose/README.rst
AaronHForgeFlow: Removed from readme

AaronHForgeFlow:
@bjeficent Introduced some changes for the fix. Can you review again please?

hr_attendance_autoclose/README.rst
AaronHForgeFlow: Is ok `open_worked_hours` as name for the new field?

bjeficent (approved):
Functional test 👍

AaronHForgeFlow:
With the last commit we can close old stale attendances. Previous to it a constraint was invalidating the change

AaronHForgeFlow:
Hi @pedrobaeza, can you please update your review? I think comments are attended. Thank you!

pedrobaeza (approved):

nikul-serpentcs:
@aheficent @pedrobaeza squash Commits :-1:

pedrobaeza:
@Nikul-Chaudhary I squashed commits on merge. Please check before commenting.

nikul-serpentcs:
> @Nikul-Chaudhary I squashed commits on merge. Please check before commenting.


@pedrobaeza You talking about commit history I see https://github.com/OCA/hr/commits/10.0/hr_attendance_autoclose here


But In This PR not squashed, Anyway Sound Good :+1:

AaronHForgeFlow:
Hi @pedrobaeza I am sorry. I forgot to remind you that one of the commits added an explicit dependency to another module that is nor merged: https://github.com/OCA/hr/pull/509 
I included the fix in #509 itself. I think the best thing is to merge #509 (if CI is green). Otherwise we have to do a Pull request that reverse this merge.

pedrobaeza:
@Nikul-Chaudhary I'm talking about "Squashing on merge" option. It's what it sounds.

@aheficent that's not cool. You have to be more aware of that. Do these things:

1. In #509, rebase and delete that.
2. Ping me there when done, and I'll merge.

pedrobaeza:
And next time, set clearly WIP title and a "Depends on" text on the main comment, @aheficent

AaronHForgeFlow:
Yes I will, thanks for the advice.

#509 is rebased. Just waiting for the CI.

nikul-serpentcs:
> @Nikul-Chaudhary I'm talking about "Squashing on merge" option. It's what it sounds.

Yes, Done

> @aheficent that's not cool. You have to be more aware of that. Do these things:

> 1. In #509, rebase and delete that.
> 2. Ping me there when done, and I'll merge.

@pedrobaeza  IMO You Directly Merged #509 , dependency automatic Removed.

So no need to New PR for specific Remove dependency.

Or it's mandatory for  that

pedrobaeza:
@Nikul-Chaudhary exactly, we are going to avoid a new PR in this case.

FYI, the modification in oca_dependencies.txt is useless, as it doesn't affect to the current repository (Nikul, you have already experienced that in other PRs), but better to clean that "little mess".

nikul-serpentcs:
Yes, I Aggree @pedrobaeza 

Or Added dependency Ref. also So It's very Clear for Reviewer 

Next PR,  @aheficent And I noted your Good Tips

AaronHForgeFlow:
@pedrobaeza It seems #509 is green so better to fix this asap. I'll make sure this won't happen again.

PR #511, opened 2018-11-14
[10.0][ADD] hr_public_holidays : Allow to change dates in wizard

grindtildeath:
When creating next year's public holidays using the wizard, holidays being flagged
with variable=True (Date may change) were created using last year's date although
this date will not be the same in 99% of the cases.
This commit adds a new model with an onchange to allow the user to set next years
holiday date in the wizard before the creation of related holidays.

Based on :
 - [x] #508 

TODO :
 - [x] Add tests
 - [ ] Add a function to compute easter related days ? Not sure it's right to do it here though :thinking:

### DIFF
--- a/hr_public_holidays/README.rst
+++ b/hr_public_holidays/README.rst
@@ -79,6 +79,7 @@ Contributors
 * Tecnativa - Pedro M. Baeza
 * Nedas Zilinskas <nedas.zilinskas@xpansa.com> (Ventor, Xpansa Group <https://ventor.tech/>)
 * Artem Kostyuk <a.kostyuk@mobilunity.com>
+* Akim Juillerat <akim.juillerat@camptocamp.com>
 
 Maintainer
 ----------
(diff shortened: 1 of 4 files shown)


grindtildeath:
@yweng8111 @ntodorova Since you work on a similar feature in #476, maybe you can give a look here ?

grindtildeath:
@pedrobaeza I answer here for your comment on #508. I guess we can either merge or close it to continue the discussion here where are the biggest changes.
I had a look at #476, but have the feeling it brings unnecessary complexity by using country specific dependencies. Whereas this PR allows the user to change the dates before the generation, what gives probably enough flexibility for a process being usually done only once a year. (we actually needed in a project where public holidays creation is heavily overriden to add costly specific computations)

IMO the method to calculate easter sunday and the related 'easter dependent' holidays could be great though, but it's already somehow localization specific, and I'm not sure it's the right place to implement it.

tschanzt (approved):

yweng8111:
@grindtildeath We have just the same problem, Instead of improving the existed Wizard, I have decided to implement a new Wizard that generate/copies the fixed holidays and recomputes the floating holidays for the given year. (#476 ) It is more convenient for the users. But every country has different fixed and floating holidays. So every country should have its specific implementation. The German implementation was already finished. I will create a new PR in "OCA/l10n-germany" as soon as possible.

gurneyalex (approved):

alexey-pelykh (approved):
Code LGTM

Issue #520, opened 2018-12-02
hr_employee_id Error


mnaphade:
I have hr_employee_id module installed on 11.0-20181121 CE. Everything seems alright for the privileged users. However, when employees login with their user credentials, they see an error upon clicking "Employees" module in their login.
```
Error:
Uncaught Error: QWeb2 - template['kanban-box']: Runtime Error: TypeError: Cannot read property 'raw_value' of undefined

https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3689
Traceback:
Error: QWeb2 - template['kanban-box']: Runtime Error: TypeError: Cannot read property 'raw_value' of undefined
    at Object.exception (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3689:7)
    at Engine.eval (eval at _render (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3731:73), <anonymous>:73:29)
    at Engine._render (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3730:296)
    at Engine.render (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3730:151)
    at Engine._render (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3734:57)
    at Engine.render (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:3730:151)
    at Class._render (https://portal.company.com/web/content/516-50c604e/web.assets_backend.js:1534:118)
    at https://portal.company.com/web/content/509-691ab44/web.assets_common.js:802:681
    at fire (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:796:299)
    at Object.add [as done] (https://portal.company.com/web/content/509-691ab44/web.assets_common.js:797:467)
```

If I uninstall hr_employee_id module everything seems to work alright. Any ideas?

pedrobaeza:
Hi, thanks for reporting. Are you able to reproduce the problem in runbot infrastructure (https://runbot.odoo-community.org/runbot/116/11.0 - admin/admin)?

If so, please explain with all the possible detailed steps to get to the error. Not having them, it's very difficult to hunt down the problem. Please don't be shy of adding details, and even better if you provide some screenshots or videos.

If not, this can be due to a configuration change, a custom/community module you have deployed in your DB or an outdated code. Please report back if it's the case.

For keeping a clean house in this issue tracker, if you don't answer in several days, this will be closed assuming it was one of the stated things.

[...]

PR #521, opened 2018-12-04
[FIX]hr_employee_id:Fixed the kanban view and form view load issue wh…

Trivedi-Vacha-SerpentCS:
Solution against issue no #520


alexey-pelykh:
In 12.0 it was addressed a bit differently, and it would be great to sync up on approach on fix https://github.com/OCA/hr/blob/12.0/hr_employee_id/views/hr_employee_views.xml
by adding
```
<field name="groups_id" eval="[(6, 0, [ref('hr.group_hr_user') ])]"/>
```

PR #522, opened 2018-12-05
hr_expense_analytic_tag: New module

ernestotejeda:
This module provides 'Analytic Tags' on Expense. These tags are
propagated to Analytic Entries

Cc @Tecnativa


hr_expense_analytic_tag/README.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to expenses form in *Expenses -> My Expenses -> Expenses to Submit* and you will find the new *Analytic Tags* field below *Analytic Account* field
+#. Create a new Expense record with an *Analytic Account* and one or more *Analytic Tags*
+#. Click on *Submit to Manager* button
+#. Click on *Approve* button
```
rafaelbn, line 44: Please @ernestotejeda add field analytic tag in the expense lines of the expense sheet. Thanks

rafaelbn (approved):
Tested functionally :+1: just one comment.

cc @carlosdauden @pedrobaeza

hr_expense_analytic_tag/models/hr_expense.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import fields, models
+
+
+class HrExpense(models.Model):
+    _inherit = "hr.expense"
+
+    analytic_tag_ids = fields.Many2many(
+        comodel_name='account.analytic.tag',
+        string='Analytic Tags',
+        copy=True
```
pedrobaeza: Comma at the end for being friendly with new diffs

hr_expense_analytic_tag/models/hr_expense.py
```diff
@@ -0,0 +1,21 @@
...
+    )
+
+    def _prepare_move_line(self, line):
+        line_values = super(HrExpense, self)._prepare_move_line(line)
+        if line_values['analytic_account_id']:
```
pedrobaeza: For avoiding errors on missing dictionary value
```suggestion
        if line_values.get('analytic_account_id'):
```

hr_expense_analytic_tag/models/hr_expense.py
```diff
@@ -0,0 +1,21 @@
...
+            tag_values = [[4, tag.id] for tag in self.analytic_tag_ids]
```
pedrobaeza, line 19: Shouldn't this be `line.analytic_tag_ids`?
```suggestion
            tag_values = [[4, tag.id] for tag in line.analytic_tag_ids]
```

ernestotejeda: line is a dictionary, so it doesn't have analytic_tag_ids attribute, we just want the expense tags and we access this doing that (self.analytic_tag_ids), just like in this method in the super class it get the partner of the expense (partner_id = self.employee_id.address_home_id.commercial_partner_id.id)

pedrobaeza: OK

pedrobaeza (approved):

PR #527, opened 2018-12-11
[10.0][FIX]hr_period

AaronHForgeFlow:
Actions done:
- Add menu to edit date_range_type for hr periods and hr fiscal years
- Add security rules to allow to access only periods from your companies
- Create fiscal years according your company
- Creating periods according the company of the type of the fiscal year
- Minor fixes on test

This fixes also travis in OCA/hr 10.0 branch

IMO ready to review now.

Regards.

### DIFF
--- a/hr_period/__manifest__.py
+++ b/hr_period/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': 'HR Payroll Period',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.1.0',
     'license': 'AGPL-3',
     'category': 'Generic Modules/Human Resources',
     'summary': "Add payroll periods",
@@ -14,6 +14,7 @@
     'website': 'https://github.com/OCA/hr',
     'depends': [
         'hr_payroll',
+        'hr_payroll_account',
         'date_range',
     ],
     'data': [
@@ -22,6 +23,7 @@
         'security/ir.model.access.csv',
         'views/hr_period_view.xml',
         'views/hr_fiscalyear_view.xml',
+        'views/date_range_type_view.xml',
         'views/hr_payslip_view.xml',
         'views/hr_payslip_run_view.xml',
         'views/hr_payslip_employee_view.xml',
--- a/hr_period/__manifest__.py
+++ b/hr_period/__manifest__.py
@@ -14,7 +14,6 @@
     'website': 'https://github.com/OCA/hr',
     'depends': [
         'hr_payroll',
-        'hr_payroll_account',
         'date_range',
     ],
     'data': [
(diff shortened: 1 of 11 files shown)


AaronHForgeFlow:
OCA/hr 10.0 branch is red because of this. Ping @OCA/human-resources-maintainers

hr_period/__manifest__.py
```diff
@@ -14,6 +14,7 @@
     'website': 'https://github.com/OCA/hr',
     'depends': [
         'hr_payroll',
+        'hr_payroll_account',
```
pedrobaeza: Why are you adding this dependency?

pedrobaeza (dismissed):
This is not the best way to fix the tests, as this can lead to other side effects with demo data being manipulated by other modules through XML files with data. Please fix company creation instead. One possibility is to add `post_install = True` and `at_install = False`.

hr_period/__manifest__.py
AaronHForgeFlow: Because of the method `on_change_contract` in `'hr_payslip.py'` it calls super and super is in `hr_payroll_account`

AaronHForgeFlow:
Ready to review now

AaronHForgeFlow:
@pedrobaeza can you update your review plz

hr_period/__manifest__.py
pedrobaeza: Then this is not the solution. Just use other name for the onchange_method and don't call super. It should be `_onchange_contract_hr_period`.

pedrobaeza:
I dismiss the review for now, but one extra comment.

hr_period/security/hr_period_security.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
nikul-serpentcs: @aheficent `<odoo noupdate="1">`

nikul-serpentcs (commented):
Minor Change

hr_period/security/hr_period_security.xml
AaronHForgeFlow: ok!

hr_period/__manifest__.py
pedrobaeza: This is still pending

AaronHForgeFlow: Yes. I will do that asap.

pedrobaeza:
Please check Travis status

pedrobaeza (approved):
I'm going to fast-track it for greenifying the branch.

AaronHForgeFlow:
Thank you!

MiquelRForgeFlow:
Squash! 💩

PR #531, opened 2018-12-21
[12.0][ADD] hr_holidays_credit: allow employees to take leave credit

alexey-pelykh:
With this module installed, HR Officer can configure specific leave types to
allow negative balance, or allow employees take leave credit.

NOTE: This is sort of getting "limit" feature from Odoo 11 back https://github.com/odoo/odoo/blob/696269bbc7781dfe320c62708cfafa8e71e69de0/addons/hr_holidays/models/hr_holidays.py#L46


alexey-pelykh:
@nikul-serpentcs @aheficent if you don't mind, please review this one

hr_holidays_credit/models/hr_leave_type.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import models, fields
+
+
+class HrLeaveType(models.Model):
+    _inherit = 'hr.leave.type'
+
+    allow_credit = fields.Boolean(
+        string='Allow Credit',
+        default=False,
```
nikul-serpentcs: @alexey-pelykh `Boolean` always default False, so no need to `default=False,`

nikul-serpentcs (commented):
Improve Code

hr_holidays_credit/models/hr_leave_type.py
alexey-pelykh: Thanks, done!

alexey-pelykh:
@nikul-serpentcs changes made, please re-review

nikul-serpentcs (approved):
Code Review LGTM :+1:

alexey-pelykh:
@sswapnesh since you're actively working with leaves in 12.0, this OCA module may be of interest for you

alexey-pelykh:
@aheficent please review this one if you're aware of the topic

sswapnesh:
We can also use (Restrict) this functionality Limited to Specific Group of Employees.(May be in new module) 

(As we are doing this for Mass allocation or Leave request using this field `holiday_type` https://github.com/odoo/odoo/blob/12.0/addons/hr_holidays/models/hr_leave.py#L151)

Limit this feature to (Selection field - use current functionality if not set )

- **Employees** (m2m of employees - Allowed employees can take leave credit)
- **Company** (m2m of company - Related user belongs to Allowed companies can take leave credit)
- **Department**  (m2m of department - If Employee's department is in Allowed department than employee can take leave credit)
- **Employee Tag**  (m2m of employee tag - If Employee's tag (at leats one) belongs to one of the allowed Tags than employee take credit leave)

alexey-pelykh:
@sswapnesh the limitation should be working according to `holiday_type`, which would change the described behavior of employee tag restriction to be active only for employee-tag leaves.

sswapnesh:
@alexey-pelykh I think I didn't understand your comment correctly.
Could you please explain little more ?

Thanks

alexey-pelykh:
Limitation by making leave request only applies if `holiday_type == 'employee'` (see `_check_holidays`), thus:

> Employees (m2m of employees - Allowed employees can take leave credit)

Would work, good point.

> Company (m2m of company - Related user belongs to Allowed companies can take leave credit)

IMO, not applicable, since user can be logged in into single company at the same time only.

> Department (m2m of department - If Employee's department is in Allowed department than employee can take leave credit)

May be confusing with `holiday_type == 'department'`

> Employee Tag (m2m of employee tag - If Employee's tag (at leats one) belongs to one of the allowed Tags than employee take credit leave)

May be confusing with `holiday_type == 'category'`

sswapnesh:
Yes that is for Leave request, Here I am talking about adding this configuration in Leave type (When allow_credit is Checked and than it should consider the selection field if set)

sswapnesh:
@alexey-pelykh 
I didn't see your comments as it was in code format earlier (Sorry, I was fast there)

> Company -- IMO, not applicable, since user can be logged in into single company at the same time only.

That's True (We can check allowed companies too but I am not sure about multi-company flow here)


> Department -- May be confusing with holiday_type == 'department'
Employee Tag (m2m of employee tag -May be confusing with holiday_type == 'category'



We don't have holiday_type in leave type so we can check based on selection option
(holiday_type is used for Leave request/Allocation)

alexey-pelykh:
@sswapnesh enhanced for filtering by employees, categories, and departments, yet w/o companies

alexey-pelykh:
@sswapnesh what are your thoughts on the latest version?

hr_holidays_credit/models/hr_leave.py
```diff
@@ -0,0 +1,48 @@
...
+    def _is_holiday_credit_allowed(self):
+        self.ensure_one()
+
+        HrEmployee = self.env['hr.employee']
+        leave_type = self.holiday_status_id
+
+        if not leave_type.allow_credit:
+            return False
+
+        employee_ids = HrEmployee
```
sswapnesh: Can we do something like this:
```
if self.employee_id.id in leave_type.creditable_employee_ids:
    return True   # we dont need to check in next step if Employee is already added in  `creditable_employee_ids`
```

Same way for creditable_department_ids (`if self.employee_id.department_id in leave_type.creditable_employee_category_ids`)

and `return False` in last

IMO this will be optimised as well.
Ps: I have not pull and test this yet but just quick overview
I will test in this weekend add update you.

alexey-pelykh: `return False` in the last won't do the job (since False has to be returned only if any filter was set), but I got your point and updated the module

sswapnesh: Now this seems much better and Effective as well.👍

alexey-pelykh: Awesome! Don't forget to approve the PR if it works good ;)

sswapnesh:
@alexey-pelykh  you should also consider to alter name_get method to Remove string  `X remaining out of Y` if employee is eligible for leave credit

screenshot

alexey-pelykh:
@sswapnesh thanks for a hint, I've updated the PR

sswapnesh (approved):

alexey-pelykh:
@pedrobaeza any other actions needed here?

PR #532, opened 2018-12-21
[12.0][ADD] hr_holidays_security: adjusted leave security

alexey-pelykh:
There's no real value of hiding fact that someone is absent from other employees, yet details of why/who approved/etc should be visible to Officers only

Features:
 - [x] _Leaves > Overview_ should be visible to all employees (https://github.com/odoo/odoo/pull/30882)
 - [x] `hr_leave.name` should be accessible only by Officer (`hr_holidays.group_hr_holidays_user`) and above (shown as ****** to others - see https://github.com/odoo/odoo/blob/8ad348bf969896ac792ba9f4d7886dd6598ee4ff/addons/hr_holidays/models/hr_leave.py#L509) in `hr_leave_view_kanban`, `hr_leave_view_form`
 - [x] `hr_leave.report_note` should be accessible only by Manager (`hr_holidays.group_hr_holidays_manager`) and above, modify `report_note` field (https://github.com/odoo/odoo/pull/30884)


alexey-pelykh:
@jbeficent @pedrobaeza it would be great to hear input from you.

pedrobaeza:
No special use case by my part.

alexey-pelykh:
@nikul-serpentcs @sswapnesh it would be great to get your feedback on this

AaronHForgeFlow:
Is it not possible to filter the records by using groups and record rules?

alexey-pelykh:
@aheficent no, these are group-related changes and field-level security

AaronHForgeFlow:
Yes but anyway it is possible to use a record rule and a domain_force using the user. But at first sight it looks complex. Not sure, I'll let experts to share their opinions.

alexey-pelykh:
That would be an overkill, that would apply only to https://github.com/odoo/odoo/pull/30884 if I got you correctly

PR #534, opened 2018-12-22
[12.0][MIG] hr_holidays_public

alexey-pelykh:


alexey-pelykh:
@pedrobaeza ^ seems that @feketemihai used other email here, is there anything I should do about that?

alexey-pelykh:
Note to myself: in "Working Hours", there are "Global Leaves" tab - maybe it's the same thing?

alexey-pelykh:
It seems that these global leaves are counted properly as "everyone is out" over here https://github.com/odoo/odoo/blob/88242d6a15a03955b6f219c8330420454e571e6c/addons/resource/models/resource.py#L239

alexey-pelykh:
After diving deeper into the code, it seems that this module has to be refactored to some extent to be an extension of "Global Leaves", or in other words - become a wizard that creates Global Leaves according to more advanced configuration. Linking to state is not that needed anymore, since linking to Calendar would be more than enough to identify the set of public holidays.

As-is usage of Global Leaves should be good except for having to manually duplicate holidays for every upcoming season

pedrobaeza:
Not exactly. Global leaves are... so global. You can't have the granularity of country, state or city for holidays. This is why this exists.

alexey-pelykh:
Are we on the same page that they are applied at resource calendar level? So not that global?

pedrobaeza:
Yeah, but forcing to use a different calendar for each possible combination is very hard IMO, thus when you deal with city/state holidays like in Spain.

alexey-pelykh:
Absolutely no argue here - doing that manually (use a different calendar for each possible combination) is nonsense. In addition to that, to maintain better compatibility (due to how formulas changed) - I think it would be an option for `hr_holidays_public` to become a manager of "global leaves" and create/unlink proper derivative calendars, and assign proper one (from a combination) for a specific tag/department/etc.

For better understanding, please check https://github.com/odoo/odoo/blob/88242d6a15a03955b6f219c8330420454e571e6c/addons/resource/models/resource.py#L254

pedrobaeza:
What do you propose then?

alexey-pelykh:
I think that solution is to create a `resource.calendar.leaves` with `resource_id = False` via o2o for each line (which would be a global leave, technically)

alexey-pelykh:
That would allow to remove the custom calculation, yet it would be needed to port the exclude feature

pedrobaeza:
But again, what do you do with the local/state holidays?

alexey-pelykh:
Create a `resource.calendar` for each tuple? I got your point, I'll think more about this - as this point I see no elegant way to port this, yet keeping things as-is in Odoo is as well not acceptable

pedrobaeza:
I think it's not too much problem to leave the module as is, as it injects in the proper method the leaves when suitable, so it's not a big deal.

alexey-pelykh:
`_leave_intervals` expects tuple with (from, to, 'resource.calendar.leaves' ID), thus some leave ID has to be specified, preferably unique per year/holiday, IMHO

pedrobaeza:
Uhm, that's new in v12. In v11 there were no `ID`... Any way, that ID is used later?

alexey-pelykh:
v12 vs v11 difference is exactly why I've started the discussion
Re: actual usage - it's not used, as far as I can see, and in loops it's a stub `meta` variable, so maybe we can put any other meta thing there and let it be like that.

pedrobaeza:
Yes, I think so for now, mentioning this in the "Known issues", and for sure in v12 there won't be problem. On v13, maybe Odoo will use that info for something, and then we will see...

alexey-pelykh:
This migration has passed the WIP stage

alexey-pelykh:
@pedrobaeza @aheficent this module is ready for review

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -0,0 +1,201 @@
...
+    )
+
+    @api.multi
+    @api.constrains('year', 'country_id')
+    def _check_year(self):
+        for line in self:
+            line._check_year_one()
+
+    def _check_year_one(self):
+        if self.country_id:
```
ernestotejeda: Maybe the condition is not necessary. I think you could replace

	...
	    if self.country_id:
            domain = [('year', '=', self.year),
                      ('country_id', '=', self.country_id.id),
                      ('id', '!=', self.id)]
        else:
            domain = [('year', '=', self.year),
                      ('country_id', '=', False),
                      ('id', '!=', self.id)]
	...

with::

	...
		domain = [
			('year', '=', self.year),
			('country_id', '=', self.country_id.id),
			('id', '!=', self.id)
		]
	...

hr_holidays_public/tests/test_holidays_public.py
```diff
@@ -0,0 +1,215 @@
...
+                'name': 'holiday x',
+                'date': '1994-11-14',
+                'year_id': holiday4.id,
+                'state_ids': [(6, 0, [self.env.ref('base.state_us_35').id])]
+            })
+
+    def test_isnot_holiday(self):
+        # ensures that if given a date that is not an holiday it returns none
+        self.assertFalse(self.holiday_model.is_public_holiday(
+            fields.Date.from_string('1995-12-10')
```
ernestotejeda: Here you can use directly

  date (1995, 12, 10)

instead of

  fields.Date.from_string ('1995-12-10').

In this way we guarantee that we will not have problems with the date format of the string passed by parameter, neither now nor in the future.

I suggest the same for all places in this same file where fields.Date.from_string () is used, passing a static string parameter. I think it should only be used when you want to convert to date a date obtained from database

ernestotejeda (changes requested):

ernestotejeda (commented):
For the rest, I think everything is fine

hr_holidays_public/models/hr_holidays_public.py
alexey-pelykh: 👍

hr_holidays_public/tests/test_holidays_public.py
alexey-pelykh: 👍

alexey-pelykh:
@ernestotejeda done

ernestotejeda (approved):

hr_holidays_public/README.rst
```diff
@@ -63,13 +63,27 @@ For using public holidays on leaves:
 #. If no employee is yet selected, only global holidays will be taken into
    account.
 
+Known issues / Roadmap
+======================
+
+Known Issues
```
pedrobaeza: No need to put the subtitle in the file

hr_holidays_public/models/hr_leave.py
```diff
@@ -0,0 +1,32 @@
...
+            instance = self
+        return super(HrLeave, instance)._get_number_of_days(
+            date_from,
+            date_to,
+            employee_id,
+        )
+
+    def _onchange_employee_id(self):
+        super()._onchange_employee_id()
+        self._onchange_leave_dates()
```
pedrobaeza: Why is this needed?

Is this a workaround for a known problem in Odoo? If so, put the link to the Odoo issue/PR for reference, but I think this has been solved meanwhile. Any way, very ugly to have method overridden without docstring about what is the reason for overriding it...

hr_holidays_public/models/hr_leave.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @api.onchange('holiday_status_id')
+    def _onchange_holiday_status_id(self):
+        self._onchange_leave_dates()
```
pedrobaeza: Why is this needed?

hr_holidays_public/models/resource_calendar.py
```diff
@@ -0,0 +1,71 @@
...
+from pytz import timezone
+from datetime import datetime, time
+from dateutil import rrule
+
+
+class ResourceCalendar(models.Model):
+    _inherit = 'resource.calendar'
+
+    def _public_holidays_leave_intervals(
+        self,
```
pedrobaeza: I have to say I don't like too much this style for declaring methods, as it makes very long the signature without clear distinction over the rest of the code. I prefer this style:
```python
def _public_holidays_leave_intervals(self, start_dt, end_dt, employe_id, tz,
                                     extra_argument_1, extra_argument_2):
```
I even prefer it for the diff, because although it requires 2 diff lines, you see clearly that is a change on the arguments instead other kind of change.

hr_holidays_public/readme/CONTRIBUTORS.rst
```diff
@@ -8,3 +8,6 @@
 * `Tecnativa <https://www.tecnativa.com>`__:
 
   * Pedro M. Baeza
+* `Brainbean Apps <https://brainbeanapps.com>`__:
```
pedrobaeza, line 12: Put extra line above for proper RST syntax

hr_holidays_public/views/hr_leave_type.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!--
+        Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
```
pedrobaeza: No need of the `(c)`

pedrobaeza (changes requested):

hr_holidays_public/models/hr_leave.py
alexey-pelykh: Port of https://github.com/OCA/hr/blob/0350f5d905e7c26d10f963d0457fd0d5a80fc7a4/hr_holidays_public/models/hr_holidays.py#L23

hr_holidays_public/models/hr_leave.py
alexey-pelykh: Port of https://github.com/OCA/hr/blob/0350f5d905e7c26d10f963d0457fd0d5a80fc7a4/hr_holidays_public/models/hr_holidays.py#L23

alexey-pelykh: Added lost docstring.

hr_holidays_public/models/hr_leave.py
alexey-pelykh: Added lost docstring.

hr_holidays_public/models/resource_calendar.py
alexey-pelykh: Here we have different vision, yet since this module is more of your interest than of mine - as you wish :)

alexey-pelykh (commented):
.

hr_holidays_public/models/hr_leave.py
pedrobaeza: But is this still needed in v12? You have forgotten to add the decorator `@api.onchange` in this one.

alexey-pelykh: super `_onchange_employee_id` already has the decorator, as far as I remember, I don't need to copy it
Regarding if it's needed - looks like no, I'm not 100% sure, yet I think you're correct

hr_holidays_public/models/hr_leave.py
pedrobaeza: And this is also needed or it is included upstream already?

alexey-pelykh: Indirectly, https://github.com/odoo/odoo/blob/bdfef60d60d78c5e7360fdabe80e70ea208e6078/addons/hr_holidays/models/hr_leave.py#L256
indirectly provokes `_onchange_request_parameters` via `_onchange_request_*`

What do yo think?

pedrobaeza (approved):

vhladiienko:
I'm sorry, but when this module is going to be available in the store? It's been a while since it was merged but no 12.0 version yet, which kinda blocks other dependent modules to release their 12.0 version.

pedrobaeza:
Well, the branch is published, but it doesn't appear and I don't know why. I'll ask and come back later.

vhladiienko:
You should add this branch to the **My repos** list like [this](https://i.imgur.com/mRGGsOa.png) manually if it was not done by the module owner.

pedrobaeza:
As said, it's already done. Asking to Odoo Apss support...

pedrobaeza:
OK, I have publish/unpublish several times and now it seems it is there... https://apps.odoo.com/apps/modules/12.0/hr_holidays_public/

vhladiienko:
Looks good, I could publish our module's 12.0 version successfully too now.

PR #546, opened 2019-01-26
[IMP] description

osi-scampbell:
Added a description field to hr_skill


max3903 (approved):

smangukiya (approved):

bodedra (approved):
None blocking comment.

`description` field should be text type instead of char type.

PR #551, opened 2019-02-07
[12.0][ADD] hr_holidays_validate_days: check for outdated number of days requested

alexey-pelykh:
This module allows leave's number of days validation to handle cases when
changes to Resource Calendar (e.g. Global Leaves or Timezone) may change
computed number of days.

To validate quantities of requested days:
#. Go to *Leaves > Managers > All > Leaves*.
#. Select leaves of interest and perform *Action > Validate Number of Days*.


hr_holidays_validate_days/models/hr_leave.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+_logger = logging.getLogger(__name__)
+
+
+class HrLeave(models.Model):
+    _inherit = 'hr.leave'
+
+    @api.multi
+    def action_validate_number_of_days(self):
+        def validate_number_of_days(leave):
```
tarteo: Can you put this method or the domain into class level so it can be inherited.

tarteo (changes requested):
I don't know if it's too much work but it would be nice if the result is put into a wizard. So now or later it can be extended with a button like 'Fix outdated leaves'

hr_holidays_validate_days/models/hr_leave.py
alexey-pelykh: Sure, done

alexey-pelykh:
@tarteo fixing is complicated - it requires unvalidate=>reset to draft=>....=>validate done by bot

tarteo (approved):
Thanks!

yweng8111 (approved):
Nice Job, Thanks for your contribution.

alexey-pelykh:
@pedrobaeza do you have any comments on the subject?

pedrobaeza:
By my part, we shouldn't add more plurals in the module names if avoidable. `hr_holidays` prefix is imposed by Odoo, but `days` can be replaced by the singular. Seeing the goal of the module, I would call it `hr_holidays_length_verification`, and more if you request hours, not days, in your leave :wink:

alexey-pelykh:
@pedrobaeza agree on plurals, `validation` vs `verification` vs `check`?

alexey-pelykh:
I'm leaning towards `validate_length`, only `length_check` is shorter

pedrobaeza:
`length_validation` then. It's preferable to use nouns instead of verbs. Don't worry about the length of the module, that it's not too much yet.

alexey-pelykh:
@pedrobaeza done

pedrobaeza:
README doesn't reflect the question about the days/hours duality, but if you don't mind, me neither.

alexey-pelykh:
@pedrobaeza you mentioned, I changed 👍

pedrobaeza:
Hehe, very compliant!

PR #553, opened 2019-02-08
[FIX] hr_employee_social_media: _compute_name

alexey-pelykh:


pedrobaeza (approved):
Fast-tracking this obvious bugfix

PR #554, opened 2019-02-12
[8.0][MIG] Backport of hr_attendance_rfid

fkantelberg:
Hello,

this is the backport of the module from odoo 10.0.


hr_attendance_rfid/security/ir.model.access.csv
```diff
@@ -0,0 +1,4 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_hr_attendance_rfid,access.hr.attendance.rfid,hr_attendance.model_hr_attendance,hr_attendance_rfid.group_hr_attendance_rfid,1,1,1,0
+access_hr_employee_rfid,access.hr.employee.rfid,model_hr_employee,hr_attendance_rfid.group_hr_attendance_rfid,1,0,0,0
```
yweng8111, line 3: In function attendance_action_change() will state of employee be changed. It seems that the device user just only in group_hr_attendance can either write on model hr.employee nor create data for model hr_attendance. I haven't tested it. Maybe it's also wrong in v12

yweng8111 (commented):
Nice job. Thanks for your contribution.
I have just finished my review and I didn't find some unit test for the access right. So I was confused with the group definition.

yucer:
Thank you for the port.

I find that the LINT errors have nothing to do with your code. I'll try to fix some of them in another PRs so that future error reports are more clear.

hr_attendance_rfid/security/ir.model.access.csv
fkantelberg: You are never writing to hr.employee. The state is a function field in 8.0 which uses the values from hr.attendance where you have the rights.

fkantelberg:
@yucer @yweng8111 
Apparently the unittests were inactive because the import was still using odoo instead of openerp.

HviorForgeFlow (approved):
Functional review LGTM :+1:
No code review

fkantelberg:
@yucer @yweng8111 @hveficent 
Thanks for the review. I finally fixed the tests besides the lint problems of the other modules.

JordiBForgeFlow:
@fkantelberg please propose here the required fixes for pylint, or in a new PR, so we can merge this. Please focus only on the errors in red.

hr_expense_sequence/__openerp__.py
```diff
@@ -27,6 +27,7 @@
     'author': "Serv. Tecnol. Avanzados - Pedro M. Baeza,"
               "Odoo Community Association (OCA)",
     'website': 'http://www.serviciosbaeza.com',
```
nikul-serpentcs, line 29: URL should be `https://github.com/OCA/hr`

HviorForgeFlow: This module is not related with this PR

fkantelberg:
@jbeficent I fixed all the critical pylint errors but I don't know why the runbot isn't starting and the database in the unittests can't be accessed. Do you have an idea?

hr_expense_sequence/__openerp__.py
nikul-serpentcs: > This module is not related with this PR

Ok, Noted!!

PR #556, opened 2019-02-13
[12.0][MIG] hr_attendance_theoretical_time_report: Migration to 12.0

pedrobaeza:
* Standard procedure
* Adapt data model
* Adapt views
* Adapt code for date handling
* Adapt tests

cc @Tecnativa


shartenfels:
Dear @pedrobaeza ,
Thanks a lot for the migration of this module to V12. @jgrandguillaume asked me to do some testings on this and we would like to share our feedback with you.

Feedback:
1. Computation works well from our point of view and depending on attendances and leaves considered
2. Would it be possible to flip the axis so that you have employees as lines and period as columns? This would be more user friendly and also similar with the standard attendance report. Especially also if you have more then 120 employees it would be more visible.
3. Referring to point 2 - we flipped the axis and then we searched for employees on the report. As the filter on "Previous Month or Current Month" is still active, the employee was found, but the information about the considered period disappeared. Would it be possible for you to have a look on this?
Attached a screencast for you about point 3: 
[vokoscreen-2019-02-14_11-43-25.mkv.zip](https://github.com/OCA/hr/files/2864658/vokoscreen-2019-02-14_11-43-25.mkv.zip)

Thanks a lot for your feedback in advance.
Have a nice day.
Cheers,
Stefan

pedrobaeza:
Hi @shartenfels,

About point 2, I have flipped the axis already on the code, as it's a good suggestion due to the default filter. You will have the same problem if removing the filters when you have a lot of months, but it's less critical.

About point 3, I'm afraid that's not something on the module itself, but the Odoo's pivot view handling, which can be considered a bug that you can report to them.

jgrandguillaume (approved):
Hi,


Thanks for the job. Works well at my end too. Good to hear you flipped the axis already ! I understand the point 3, I think you're right, probably something to report to Odoo.

Regards,

Joël

shartenfels:
Dear @pedrobaeza ,
Thanks for your feedback and the flip of the axis.

Regarding point 3 I will do some further investigations.

Best regards,
Stefan

PS: Seems Odoo has an issue on the Pivot as soon as you flip the axis, set a filter and then remove the filter. Expected result having same information as before the filter - so I report a bug.

tschanzt (approved):
Looks good to me

pedrobaeza:
Thanks to all reviewers. Merging.

pedrobaeza:
@shartenfels this commit (from yesterday) might fix the problem reported in point 3, but it can be another similar thing not yet fixed.

pedrobaeza:
Sorry, the commit is:

https://github.com/odoo/odoo/commit/de8cefcef4d456398e5028c52e3bd9db807e5f8a

shartenfels:
Hola @pedrobaeza , Thanks a lot for the hint. Cheers, stefan

PR #559, opened 2019-02-25
[MIG] 11.0 Backport of hr_attendance_report_theoretical_time

jarroyomorales:
Backport of `hr_attendance_report_theoretical_time` to 11.0.
Due to big changes in Odoo 12 regarding hr modules, i have needed to do a hook with resource.calendar.leaves in order to make the domain work properly. I don't know if this is the proper solution...
@jbeficent


hr_attendance_report_theoretical_time/models/resource_calendar_leaves.py
```diff
@@ -0,0 +1,13 @@
+from odoo import api, models
+
+
+class ResourceCalendarLeaves(models.Model):
+    _inherit = 'resource.calendar.leaves'
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        new_args = args
```
pedrobaeza: No need to create a new variable. Also add a docstring about why do you overwrite this.

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py
```diff
@@ -179,25 +179,23 @@ def _theoretical_hours(self, employee, date):
         """Get theoretical working hours for the day where the check-in is
         done for that employee.
         """
+        date_date = fields.Datetime.from_string(date)
         if not employee.resource_id.calendar_id:
             return 0
-        tz = employee.resource_id.calendar_id.tz
```
pedrobaeza: Why removing timezone info?

jarroyomorales: Because 'resource.calendar' object has no attribute 'tz' in 11.0

pedrobaeza: Yeah, but the tz should reside in other part (user, employee, resource...)

hr_attendance_report_theoretical_time/models/resource_calendar_leaves.py
```diff
@@ -6,6 +6,9 @@ class ResourceCalendarLeaves(models.Model):
 
     @api.model
     def search(self, args, offset=0, limit=None, order=None, count=False):
+        """
+        This function is overrided in order to add the domain into searches
```
pedrobaeza: A docstring should explain for what is used the code, no what code does, which is self-explanatory. Try with something like:
```
Inject extra domain when being called from _theoretical_hours method.
WARNING: This method should be called only once with that context
from a restricted piece of code, or can have side effects.
```

hr_attendance_report_theoretical_time/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,28 @@
+You need to be at least "Attendance / Manual Attendance" for being able to see
```
HviorForgeFlow, line 1: At least you have to be Officer because [parent menu](https://github.com/odoo/odoo/blob/b08aa26d8fd453f94b41307ac442489cf71e949a/addons/hr_attendance/views/hr_attendance_view.xml#L187) requires it.

HviorForgeFlow (changes requested):
So employees with manual attendance can't see their own report from a menu, but they can access to the view and see all employee attendances.

Would be nice that every employee could access to this menu an see only their own report.

hr_attendance_report_theoretical_time/readme/CONFIGURE.rst
pedrobaeza: This wasn't this way in v10, but on v12 this menu was accessible by all, so you should turn to the changes I was doing in v10 for this. There's a record rule for making that users with manual attendance only see their attendances any way.

JordiBForgeFlow (approved):
:+1:  Technical and functional review.

AdriaGForgeFlow (approved):
Code and functional review on runbot. LGTM :+1:

hr_attendance_report_theoretical_time/readme/CONFIGURE.rst
HviorForgeFlow: Still being an inconsistency between readme and module functionality at least in v11.
If you want to keep as this way please change readme

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
```diff
@@ -0,0 +1,68 @@
...
+        <field name="model">hr.attendance.theoretical.time.report</field>
+        <field name="arch" type="xml">
+            <search>
+                <field name="employee_id"/>
+                <filter name="today" string="Today" domain="[('date', '=', datetime.date.today().strftime('%Y-%m-%d'))]" />
+                <filter name="previous_month" string="Previous Month" domain="[('date', '&gt;=', (datetime.date.today() - relativedelta(months=1)).strftime('%Y-%m-01')), ('date', '&lt;', datetime.date.today().strftime('%Y-%m-01'))]" />
+                <filter name="current_month" string="Current Month" domain="[('date', '&gt;=', datetime.date.today().strftime('%Y-%m-01'))]" />
+                <filter name="current_year" string="Current Year" domain="[('date', '&gt;=', datetime.date.today().strftime('%Y-01-01'))]" />
+                <separator/>
+                <filter name="my" string="My Attendances" domain="[('employee_id.user_id.id', '=', uid)]" />
```
nikul-serpentcs, line 15: @jarroyomorales `name="my_attendances"`

yajo:
Is this still WIP?

nikul-serpentcs:
> Is this still WIP?

@Yajo I think here access for `PR Labels` like `WIP`, Because Every time reviewer asking.
@Yajo @pedrobaeza What about that??

yajo:
I don't understand what you said, sorry. The PR title starts with WIP, that's why I asked. Is it WIP then?

nikul-serpentcs:
> I don't understand what you said, sorry. The PR title starts with WIP, that's why I asked. Is it WIP then?

I asking to here we not access `PR Labels`, So if a `developer or PR author` access then added `WIP` stage. 
So, No need to asking the Every time.

aaaaaaaaaaaaa

hr_attendance_report_theoretical_time/readme/CONFIGURE.rst
pedrobaeza: Did you attend this comment?

jarroyomorales: Sorry, i forgot, just changed it in #576

PR #560, opened 2019-02-26
[12.0] hr_attendance_report_theoretical_time: My Attendance as default

tschanzt:
Should make it nicer to use and make it faster if you can see a lot of employees.

### DIFF
--- a/hr_attendance_report_theoretical_time/__manifest__.py
+++ b/hr_attendance_report_theoretical_time/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Theoretical vs Attended Time Analysis",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.0.1",
     "category": "Human Resources",
     "website": "https://github.com/OCA/hr",
     "author": "Tecnativa, "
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I agree it's faster only with "My attendances", but not with "Current year". Imagine someone removes the filter "My attendances", on December and it's attendance officer. Then, all the attendances will be computed for the whole year for all employees. That's too much.

I prefer to keep the current month + previous month filter.

tschanzt:
Thats true my description wasn't precise enough there. 
Its true that the computation will be too long if you remove the My Attendances filter, but it will be the same if someone removes both filters if we have current and previous month. (Although it might save you if you remove the filter by accident). But for most cases i think its more interesting to see the whole year.

pedrobaeza:
As I put these filters in the original version, I still consider both as the best ones, but OK if you decide to change them. This is easily switchable to other through favorite put as default for all users.

tschanzt:
We discussed it internally again and will customize the filters in the customer project. So this pr only adds the my attendance filter as a default.

pedrobaeza (approved):

AaronHForgeFlow:
This report is meant to be used normally by hr managers, isn't it?

pedrobaeza:
It can be used by any employee with manual attendance.

AaronHForgeFlow (approved):
But normally it is used by the manager for reporting purposes, isn't it? Well, anyway it is not that costly to remove a filter so it is ok to me.

PR #566, opened 2019-03-13
[FIX] hr_public_holidays: Wrong variable usage

yajo:
Without this patch, opening the holidays app fails with:

    TypeError: publicHolidays.map is not a function

    http://localhost/hr_public_holidays/static/src/js/holidays_highlighter.js:25

Now it should be OK. Promises were being used as Array or String wrongly.

@Tecnativa


rafaelbn (approved):
:+1: thanks

pedrobaeza (approved):

PR #571, opened 2019-03-21
[11.0] [MIG] hr_employee_calendar_planning

jarroyomorales:
Migration of  hr_employee_calendar_planning and added overlapping of lines constrains


hr_employee_calendar_planning/hooks.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, SUPERUSER_ID
+from collections import defaultdict
+
+
+def post_init_hook(cr, registry, employees=None):
```
JordiBForgeFlow, line 8: @pedrobaeza during the migration to v11 we realized that the previous init hook was not working properly, and was producing overlapping calendars, and so it would crash. We have simplified, as it can be really complex to migrate without overlaps. I think that each company should consider installing this module and then deciding how to best convert their existing calendars.

pedrobaeza: Not really. On "fragmented" calendars, there's no overlap. The overlap on employees might happen if your original calendar had it, but simply deactivate the constraint through a context key on that init hook, but please, don't remove this as good initial setup.

jarroyomorales:
@pedrobaeza We were having problems with the post init hook and thought that simply copying the calendar that the employee had onto the new hr.calendar model is much easier and is not a bad initial setup either. What do you think?

pedrobaeza:
What are the problems? This method is intended for splitting current calendar by dates. Doing other thing will mix periods and generate duplicate calendars.

jarroyomorales:
@pedrobaeza You were right with the hook, sorry for adding and removing it all the time. The problem we were having was that in 11.0 resource.calendar has default attendance lines and the hook was causing the calendar to have all the lines doubled. I think that with the small fixes we added this is solved.

hr_employee_calendar_planning/hooks.py
etobella: Now the old logic has been maintained.

etobella (approved):
LGTM :+1:

hr_employee_calendar_planning/models/hr_employee_calendar.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
```
pedrobaeza: Why splitting this in a new file? It was put on purpose in the previous one. Check contributing guidelines where this is specified that a submodel should go in the same file of the main model:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#file-naming

jarroyomorales: Okay! Fixed.

hr_employee_calendar_planning/models/resource_resource.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResourceResource(models.Model):
+    _inherit = 'resource.resource'
+
+    calendar_id = fields.Many2one(required=False)
```
pedrobaeza: Why are you changing this as not required?

jarroyomorales: You need it in order to be able to execute employee.resource_calendar_id = False at the end of the post init hook since it is a related field linked to calendar_id, which in 11.0 is required.

pedrobaeza: That was for regenerating the calendar, but I added at the end a check also on active field, so you can deactivate the calendar instead of this.

pedrobaeza:
Travis is red

hr_employee_calendar_planning/models/resource_resource.py
jarroyomorales: Okay, so can i just delete employee.resource_calendar_id = False and leave calendar_id as it was?

pedrobaeza: You have to replace it by `employee.resource_calendar_id.active = False`

jarroyomorales: Doing so causes two tests to fail because we dont enter this if in regenerate_calendar:
```
        if not self.resource_calendar_id or self.resource_calendar_id.active:
            self.resource_calendar_id = self.env['resource.calendar'].create({
                'active': False,
                'name': _(
                    'Auto generated calendar for employee'
                ) + ' %s' % self.name,
                'attendance_ids': [],
            }).id
```

I think that removing `employee.resource_calendar_id = False` and keeping `active` equal to True is what we need and makes all tests pass.

pedrobaeza: Right, I have looked incorrectly at the regenerate method. Please make sure that after this, you switch again the calendar to active=False.

jarroyomorales: Yes, on the creation dictionary there is already `'active': False`.

pedrobaeza: Yeah, but that are the split ones. I'm talking about the original one.

jarroyomorales: okay

jarroyomorales:
Added recomputation when unlinking a `hr.employee.calendar` from an employee.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -65,13 +63,23 @@ class HrEmployeeCalendar(models.Model):
...
         for employee in self.mapped('employee_id'):
             employee._regenerate_calendar()
         return res
+
+    @api.multi
+    def unlink(self):
+        employees = self.mapped('employee_id')
+        res = super(HrEmployeeCalendar, self).unlink()
+        for employee in employees:
+            employee._regenerate_calendar()
```
pedrobaeza, line 87: This doesn't serve and can lead to errors. You only need to remove manually the linked specific calendar

hr_employee_calendar_planning/views/hr_employee_views.xml
```diff
@@ -6,10 +6,10 @@
         <field name="model">hr.employee</field>
         <field name="inherit_id" ref="hr.view_employee_form"/>
         <field name="arch" type="xml">
-            <field name="calendar_id" position="attributes">
-                <attribute name="invisible">1</attribute>
+            <field name="resource_calendar_id" position="attributes">
+                <attribute name="invisible">0</attribute>
```
pedrobaeza: This is invisible on purpose. Don't remove it.

pedrobaeza (changes requested):

hr_employee_calendar_planning/views/hr_employee_views.xml
jarroyomorales: Oh yes sorry, I put it visible to do some tests and forgot to change it back.

hr_employee_calendar_planning/models/hr_employee.py
jarroyomorales: Without this unlink the user can delete `hr.employee.calendar` lines from an employee (in case he did some mistakes) and the resource calendar behind will not be updated, which in my opinion is not correct.

pedrobaeza: Sorry, I have mixed models. I was talking about unlink in `hr.employee`, which should remove the linked private calendar. Yes, this is OK.

jarroyomorales: Another case where it is not regenerated is when there are changes in attendance lines of a calendar. Again, I know these changes should only be done when the user made a mistake configuring the calendar, but do you think we should take them into account just in case?

pedrobaeza: In this case I see a possible infinite loop problem and a lot of code for controlling that, so I prefer to not put it. You can put a button next to the calendar mapping table to manually regenerate the calendar as a "workaround".

jarroyomorales: Putting a button could lead to confusion, the user may not now what 'Regenerate' means or can be hard to know which users you have to regenerate and when. 
I tried doing what I told you before and the infinite loop could happen theorically if the resource calendar of an employee has `active` as True. In practice, this is not possible since the only ways of modifiying the resource calendar of an employee are with the hook, which sets active to False, and regenerate, which also sets it to False.

pedrobaeza: OK, try it then

jarroyomorales: After trying it a lot I haven't had any problem with the infinite loop, so I think we can keep this feature to make sure that the calendar planning and the resource_calendar are always consistent.

pedrobaeza (approved):

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def regenerate_calendar(self):
+        self._regenerate_calendar()
+
+
+class HrEmployeeCalendar(models.Model):
+    _name = 'hr.employee.calendar'
+
+    date_start = fields.Date(
+        string="Start",
```
nikul-serpentcs: @jarroyomorales  string="Start", --> `string="Start Date",`

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    date_end = fields.Date(
+        string="End",
```
nikul-serpentcs: @jarroyomorales string="End", --> `string="End Date",`

hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
```diff
@@ -0,0 +1,113 @@
...
+
+from odoo.tests import common
+from ..hooks import post_init_hook
+
+
+class TestHrEmployeeCalendarPlanning(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestHrEmployeeCalendarPlanning, cls).setUpClass()
+        cls.calendar1 = cls.env['resource.calendar'].create({
```
nikul-serpentcs: `cls.env['resource.calendar']` use multiple time so use like a global variable

nikul-serpentcs (commented):
nitpicking

jarroyomorales:
@nikul-serpentcs Changes done!

nikul-serpentcs (approved):
Only Code Review LGTM :+1:

nikul-serpentcs:
@pedrobaeza  @jarroyomorales here need squash 2 Commits/bot commits?
zzzzzzzzzz

pedrobaeza:
Well, not really, as they come from different bots and this is an intermediate version. On upper ones, we can squash that with the migration commit itself.

PR #574, opened 2019-03-22
[ADD] hr_calendar_multiweek

etobella:
This addon allows to use resource calendars that works differently according to the week. For example, on the first week I work on mornings, on the second week on the afternoons and so on.


pedrobaeza:
Isn't this overlapping with #571?

etobella:
No, this allows to use different weekly configuration and the other works to define changes o employee calendars.
Also, this module works with theoretical time and calendar_planning as expected

pedrobaeza:
OK, I see. You can get the same result with the other module, but it's a pain to define in a fixed way the dates.

etobella:
Well, you should define every week of the employee... with this change you only define 1 calendar

pedrobaeza:
Yeah, indeed. It will be a compatibility problem with `hr_employee_calendar_planning` any way.

etobella:
No with the 11.0 version. We have made a change in order to make it compatible :wink: 
https://github.com/OCA/hr/pull/571/files#diff-5b4aa33afaecf895285b9aed57eb4739R30

pedrobaeza:
OK, great. I suppose you have added code for copying all fields except the blacklisted ones (start and end date).

etobella:
Exactly! :smile:

jarroyomorales (approved):

hr_calendar_multiweek/models/resource_calendar.py
```diff
@@ -0,0 +1,55 @@
...
+    def _get_day_attendances(self, day_date, start_time, end_time):
+        res = super()._get_day_attendances(day_date, start_time, end_time)
+        return res.filtered(lambda r: r._check_week(day_date))
+
+
+class ResourceCalendarAttendance(models.Model):
+    _inherit = "resource.calendar.attendance"
+
+    calendar_week_number = fields.Integer(default=1)
+    week_number = fields.Integer(default=1)
```
jarroyomorales: @etobella Could you please add a `help` to this fields or explain their difference better in the readme? I think it is not clear at first sight what they are meant to be used for and it may help other people review this module and get it merged. :wink: :wink:

Abranes (approved):

PR #575, opened 2019-03-22
[10.0][FIX] hr_period

AaronHForgeFlow:
Since we have hierarchy in date_ranges (https://github.com/OCA/server-tools/pull/1431) this modules fails when creating new hr periods from scratch. 

For now I disable the feature for those date ranges types, until better solution is found.


DucTruongKomit (approved):
LGTM overall

AaronHForgeFlow:
I close this as long as the issue was fixed in a different way here: https://github.com/OCA/hr/pull/883

PR #577, opened 2019-03-28
[12.0][MIG] hr_employee_firstname

astirpe:
Includes fix for https://github.com/OCA/hr/issues/580


luismontalba:
This is not related with this migration, but, shouldn't be the firstname mandatory as explained in readme? If you try to create an employee with only the lastname an error occurs:

DeepinScreenshot_Seleccionar área_20190406140358

astirpe:
@luismontalba yes indeed, thank you!
The issue you mentioned also occurs in previous versions of the module. Unfortunately for the previous versions, if I change the field Firstname to mandatory, it could create problems on existing installations in production.

I just committed a quick fix for the issue you mentioned. This fix does not set the field Firstname to mandatory, but just allows to create an employee with Lastname only. I think this is a safer fix for the moment. Anyway I think is better to open a dedicated issue for the previous versions, what do you think?

luismontalba:
I guess it's the best way. ĺ can write an issue to document it

luismontalba (approved):
Tested on runbot

pedrobaeza:
Why not putting it as required only in UI?

astirpe:
@pedrobaeza yes, is also possible but it will change the actual functional behavior. Do we really need the two fields both mandatory?

pedrobaeza:
In `partner_firstname` is solved being required attribute dynamic. Can't this mimic that one?

astirpe:
Yes, that's what I wanted to try to achieve, but I'm busy with other stuff now, and at the moment is not a priority for me. If it is possible I would like to keep this PR as pure porting of the module and work on that issue on a separated dedicated PR.

pedrobaeza:
OK, for me it's not blocking nor I'm using this module right now. It was just a comment as general observer of the PRs :smiley:

hr_employee_firstname/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# ©  2010 - 2014 Savoir-faire Linux (<http://www.savoirfairelinux.com>)
```
alexey-pelykh: "Copyright" instead of "©"

hr_employee_firstname/models/hr_employee.py
```diff
@@ -0,0 +1,79 @@
...
+    def _get_name(self, lastname, firstname):
+        return self.env['res.partner']._get_computed_name(lastname, firstname)
+
+    @api.onchange('firstname', 'lastname')
+    def _onchange_firstname_lastname(self):
+        if self.firstname and self.lastname:
+            self.name = self._get_name(self.lastname, self.firstname)
+
+    def _firstname_default(self):
+        return ' ' if self.env.context.get('module') else False
```
alexey-pelykh: Better extract default firstname to a variable and reuse it (at least 2 places of magic constant use)

hr_employee_firstname/views/hr_view.xml
```diff
@@ -0,0 +1,25 @@
...
+<odoo>
+
+    <record id="view_employee_form" model="ir.ui.view">
+        <field name="model">hr.employee</field>
+        <field name="inherit_id" ref="hr.view_employee_form"/>
+        <field name="arch" type="xml">
+            <xpath expr="//label[@for='name']" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='name']" position="attributes">
```
alexey-pelykh, line 11: can be written via `<field name="name">` instead of xpath

alexey-pelykh (changes requested):

astirpe:
Last commit implements https://github.com/OCA/hr/pull/577#issuecomment-480791628 and fixes some issues when changing the first and last name.

alexey-pelykh (approved):

pedrobaeza:
I have squashed commit history a bit and merging when CIs finish.

Issue #578, opened 2019-03-29
[11.0] hr_holidays_compute_days error


glauco70:
There is a problem in hr_holidays_compute_days module, if I check the "Full day" before inserting a date I have this error in _inverse_date_to_full (and in  _inverse_date_from_full also) : 
**dt = fields.Datetime.from_string(record.date_to_full).replace(
AttributeError: 'NoneType' object has no attribute 'replace'**

I resolve inserting this if : 
_if record.date_to_full:_
before the line indicated

pedrobaeza:
How do you get that situation? Can you do a PR with the solution?

glauco70:
> How do you get that situation?

I create a hr.holidays request and I check the "Full Day" check box before inserting the date_from or date_to.

> Can you do a PR with the solution ?

Yes

hugho-ad:
this can be closed

moylop260:
fixed from https://github.com/OCA/hr/pull/579

PR #579, opened 2019-03-29
[11.0] FIX hr_holidays_compute_days error when check Full Day check

glauco70:
Fix error when check "Full Day" check box with empty date fields

https://github.com/OCA/hr/issues/578#issuecomment-478037490

### DIFF
--- a/hr_holidays_compute_days/models/hr_holidays.py
+++ b/hr_holidays_compute_days/models/hr_holidays.py
@@ -99,6 +99,8 @@ class HrHolidays(models.Model):
         as fallback.
         """
         for record in self.filtered('from_full_day'):
+            if not record.date_from_full:
+                continue
             tz_name = record.employee_id.user_id.tz or record.env.user.tz
             dt = fields.Datetime.from_string(record.date_from_full).replace(
                 hour=0, minute=0, second=0, microsecond=0,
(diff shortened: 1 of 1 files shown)


hr_holidays_compute_days/models/hr_holidays.py
```diff
@@ -99,24 +99,26 @@ def _inverse_date_from_full(self):
         as fallback.
         """
         for record in self.filtered('from_full_day'):
-            tz_name = record.employee_id.user_id.tz or record.env.user.tz
-            dt = fields.Datetime.from_string(record.date_from_full).replace(
-                hour=0, minute=0, second=0, microsecond=0,
-                tzinfo=tz.gettz(tz_name),
-            ).astimezone(tz.tzutc())
-            record.date_from = fields.Datetime.to_string(dt)
+            if record.date_from_full:
```
pedrobaeza: Instead of doing this big diff, you can just:
```python
if not record.date_from_full:
    continue
```

pedrobaeza (changes requested):

hr_holidays_compute_days/models/hr_holidays.py
glauco70: I pushed the change like you suggested

pedrobaeza (approved):
Thanks!

labaggio (approved):

PR #581, opened 2019-04-16
[12.0][ADD] hr_holidays_leave_repeated

astirpe:
This module allows to create periodical leaves.

cc @CasVissers


hr_holidays_leave_repeated/models/hr_leave.py
```diff
@@ -0,0 +1,154 @@
...
+                'user_error_msg': _(
+                    '''The repetition is every week:
+                     the duration of the leave request
+                     must not exceed 1 week.
+                    ''')
+            },
+            'biweek': {
+                'days': 14,
+                'user_error_msg': _(
+                    '''The repetition is based on 2 weeks:
```
CasVissers-360ERP: @astirpe 
Every 2 weeks?

hr_holidays_leave_repeated/models/hr_leave.py
```diff
@@ -0,0 +1,154 @@
...
+        ('month', 'Every four weeks')
+    ])
+    repeat_mode = fields.Selection([
+        ('times', 'Number of Times'),
+        ('date', 'End Date'),
+    ], default='times')
+    holiday_type_repeat = fields.Boolean(related='holiday_status_id.repeat')
+    repeat_limit = fields.Integer(default=1, string='Repeat # times')
+    repeat_end_date = fields.Date(
+        'Repeat Until', default=lambda self: fields.Date.today())
```
CasVissers-360ERP: @astirpe 
Whe should choose for end date or repeat until throughout the module. I would make it easy and use end date.

hr_holidays_leave_repeated/models/hr_leave.py
```diff
@@ -0,0 +1,154 @@
...
+        if not skip_create_handler and all_vals_set:
+            self.create_handler(vals)
+        return res
+
+    @api.constrains('repeat_limit', 'repeat_end_date')
+    def _check_repeat_limit(self):
+        for record in self:
+            if record.repeat_mode == 'times' and record.repeat_limit < 0:
+                raise UserError(
+                    _('Please set a positive repeating # of times limit.'))
```
CasVissers-360ERP: @astirpe 
"Please set a positive amount of repetitions.

hr_holidays_leave_repeated/models/hr_leave.py
```diff
@@ -0,0 +1,154 @@
...
+            if record.repeat_mode == 'date' \
+                    and record.repeat_end_date < fields.Date.today():
+                raise UserError(
+                    _('The repeat until date cannot be in the past'))
```
CasVissers-360ERP: @astirpe 
Change to end date, or until date.

hr_holidays_leave_repeated/models/hr_leave.py
astirpe: Just changing it to _End Date_ causes the following warning:
`Two fields (repeat_end_date, date_to) of hr.leave() have the same label: End Date.`

So I'm going to change it to: _Repeat End Date_.

CasVissers-360ERP (approved):
Functional test done.

tbaden:
@astirpe  how about using relativedelta(months=1) instead of 28 days?
you just need a switch between days and months, this way you could even add quarter or half year interval for your repeat.

astirpe:
@tbaden initially the module used relativedelta(months=1) instead of 28 days, but it was causing troubles when the repeated leave happened in the weekend or in a day with some different working schedule. Same discourse for quarter or half year periods. To avoid such confusing situations, the monthly period is assumed as 28 days (a multiple of one week).

tbaden (approved):
ok

Issue #573, opened 2019-03-21
hr_expense_invoice V.11


ghost:
When I try to pay the employee's expense this message appears:
![1](https://user-images.githubusercontent.com/25029267/54755258-df739a00-4be5-11e9-98cd-2b2e3b5ba84b.png)
But I have checked the accounting account and it's not reconcilied.
![2](https://user-images.githubusercontent.com/25029267/54755520-60329600-4be6-11e9-8bed-a546f7fb423a.png)
And when I validated the expense this jounal entry was made. It is correct and the invoice is paid.
![3](https://user-images.githubusercontent.com/25029267/54755374-1ea1eb00-4be6-11e9-9ec0-b362b5f94510.png)
But I can't pay to my employee.

kittiu:
```
diff --git a/addons/hr_expense/wizard/hr_expense_sheet_register_payment.py b/addons/hr_expense/wizard/hr_expense_sheet_register_payment.py
index f77f013d..66d6c116 100644
--- a/addons/hr_expense/wizard/hr_expense_sheet_register_payment.py
+++ b/addons/hr_expense/wizard/hr_expense_sheet_register_payment.py
@@ -106,7 +106,7 @@ class HrExpenseSheetRegisterPaymentWizard(models.TransientModel):
 
         # Reconcile the payment and the expense, i.e. lookup on the payable account move lines
         account_move_lines_to_reconcile = self.env['account.move.line']
-        for line in payment.move_line_ids + expense_sheet.account_move_id.line_ids:
+        for line in payment.move_line_ids + expense_sheet.account_move_id.line_ids.filtered(lambda l: not l.reconciled):
             if line.account_id.internal_type == 'payable':
                 account_move_lines_to_reconcile |= line
         account_move_lines_to_reconcile.reconcile()
```

Hi, I found that, because this module will create 2 payable line for Expense. 1 payable line will already reconcile with invoice and make invoice paid. 1 payable line is left to be reconciled when doing register payment on expense.

But from Odoo core, it is getting both line (1 reconciled and 1 is not) and try to reconcile with payment entry. So, it shows error.

I see that, to fix it, we should only select unreconciled payable `expense_sheet.account_move_id.line_ids.filtered(lambda l: not l.reconciled)`

But this has to do in core Odoo.

kittiu:
Without fixing in core Odoo, I think there are 2 ugly way.
1) Do a hook addon, just to fix this
[...]

PR #584, opened 2019-04-23
[12.0][MIG] hr_expense_invoice

kittiu:
Done Migration


pedrobaeza:
Please check Travis

kittiu:
> Please check Travis

Still WIP.

kittiu:
@pedrobaeza during learning it to migration, I also have some opinion here,
https://github.com/OCA/hr/issues/585

kittiu:
Done. Only part of the code that can not happen in test.
image

kittiu:
> Done. Only part of the code that can not happen in test.
> image

Q: when we face the part of code that can't write test to cover like this, normally, how do we go about?

pedrobaeza:
You have to write use case that triggers that part of code, or let it uncovered with the codecov warning.

kittiu:
> You have to write use case that triggers that part of code, or let it uncovered with the codecov warning.

Yah :)

tbaden:
@kittiu ready for review or still wip?

kittiu:
Yes ready. Test is green now.

hr_expense_invoice/views/hr_expense_views.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="hr_expense_view_view" model="ir.ui.view">
```
tbaden: "ugly" id (view_view)
please use hr_expense_view_form (optional)

tbaden (approved):
code review: lgtm
just a minor detail, but it's optional

hr_expense_invoice/views/hr_expense_views.xml
kittiu: Done, thank you.

hr_expense_invoice/models/hr_expense.py
```diff
@@ -0,0 +1,59 @@
...
+
+from odoo import api, fields, models
+
+
+class HrExpense(models.Model):
+    _inherit = 'hr.expense'
+
+    invoice_id = fields.Many2one(
+        comodel_name="account.invoice",
+        string='Invoice',
```
luistorresm: The string param is not necessary on v12 (in this case)

hr_expense_invoice/models/hr_expense.py
```diff
@@ -0,0 +1,59 @@
...
+    @api.multi
+    def write(self, vals):
+        self._check_vals(vals)
+        return super(HrExpense, self).write(vals)
+
+    @api.multi
+    def _get_account_move_line_values(self):
+        move_line_values_by_expense = super()._get_account_move_line_values()
+        for expense_id, move_lines in move_line_values_by_expense.items():
+            expense = self.browse(expense_id)
```
luistorresm, line 53:
```suggestion
            expense = self.browse(expense_id)
            if not expense. invoice_id:
                return move_line_values_by_expense
```

To avoid the next iteration, that is not necessary when the expense does not have an invoice.

With this remove the check `expense.invoice_id` in the next `if`

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,45 @@
...
+from odoo.tools import float_compare
+
+
+class HrExpenseSheet(models.Model):
+    _inherit = "hr.expense.sheet"
+
+    @api.multi
+    def action_sheet_move_create(self):
+        precision = self.env['decimal.precision'].precision_get('Account')
+        expense_line_ids = self.mapped('expense_line_ids').filtered(
```
luistorresm:
```suggestion
        expense_line_ids = self.mapped('expense_line_ids').filtered('invoice_id')
```

hr_expense_invoice/models/hr_expense.py
kittiu: Sorry, but why? invoice_id is a new field for hr.expense also, why no need `string='Invoice'`

hr_expense_invoice/models/hr_expense.py
kittiu: Done

hr_expense_invoice/models/hr_expense_sheet.py
kittiu: Done

hr_expense_invoice/models/hr_expense.py
kittiu: By the way, I also see `oldname='invoice'`. Do we still need it for migration to v12?
What is it actually used for?

kittiu:
We also still have same problem as in v11 according to this issue
https://github.com/OCA/hr/issues/573

hr_expense_invoice/models/hr_expense.py
hugho-ad: because the string is computed using the field name, so there is any need to explicitly set the string param

kittiu: @hugho-ad, you mean "invoice_id" will be display as string "Invoice" anyway, right?

luistorresm: Yes @kittiu

kittiu: Note. But in this case, may I also change it to "Vendor Bill", and it is domain to in_invoice + open status. I think Vendor Bill is also more meaningful. :)

luistorresm: Yes @kittiu

When the label is different, the string parameter is necessary.

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,44 @@
...
+from odoo.exceptions import UserError
+from odoo.tools import float_compare
+
+
+class HrExpenseSheet(models.Model):
+    _inherit = "hr.expense.sheet"
+
+    @api.multi
+    def action_sheet_move_create(self):
+        precision = self.env['decimal.precision'].precision_get('Account')
```
luistorresm: Hello @kittiu 

Now this `decimal.precision` do no exists

kittiu: Fixed to 'Product Price'

luistorresm (approved):
Technical Review

kittiu:
@pedrobaeza we will have this issue pending - https://github.com/OCA/hr/issues/573
But it shouldn't affect this PR directly.

Thank you.

ghost (commented):
When I try to pay the expense to the employee appears the message:

kittiu:
I think this is related to account type or reconcile flag used in account of advance.

You can try different one?

ghost:
@kittiu I don't understand what you want to say. Can you explain?

ghost:
@kittiu I didn't understand what you wanted to say. Can you explain it better again, please?

kittiu:
@gelo-landoo
We do not have this problem. I think it is about accounting code config.
Or would you send me URL and step to reproduce the problem?

Issue #569, opened 2019-03-19
Validation Error in hr_public_holidays module v11


burhan-vakharia:
Odoo Version - v11 Community

1) Create a new Public Holiday.
2) Add a new Public holiday line.
3) Select the date, add name and untick the Date may change Boolean.

Now try to save the Public Holiday, it will give a Validation Error - Screenshot Link - https://prnt.sc/mzviim .

Video URL - https://nimb.ws/BGXD7x

nikul-serpentcs:
@pedrobaeza @mreficent Can you please check it?

SSanfilippo:
Hi, have any feedback about this error? Also I cannot use the module for odoo 11 for same problem of burhan.

Best regards,
Simone

moylop260:
@SSanfilippo @nikul-serpentcs @burhan-vakharia 
Could you check the following fix https://github.com/OCA/hr/pull/590 from @luistorresm, please?

SSanfilippo:
Hi, I update the module on my enviroment, but I have always the error.

The reference of my error is: [oggetto con referenza: date - date]

luistorresm:
Hello @SSanfilippo 

You have updated the module? In your Odoo instance?

SSanfilippo:
Yes @luistorresm, I had downloaded the source from github, then I extract in my addons folder the hr_public_holidays folder.
I check also if the two modified files are presents, and yes are modified.

luistorresm:
Maybe is missing make -u hr_public_holidays or in the apps update manually the module, because was changed the view.

Regards

SSanfilippo:
I make the update manually, better I restart odoo service then as developer mode update all apps.
I use windows version, probably forgot a step?

luistorresm:
Yes @SSanfilippo 

Please open the apps view and search `hr_public_holidays`, open  it and press `Update`

SSanfilippo:
Hi, good news (and sorry for my forgotten operation).
I update the module in apps and later public holiday work fine.

luistorresm:
Hello @moylop260 @max3903 

Could be closed this ticket?

Regards

SSanfilippo:
Hi at all,
probably in Odoo 12 there is the same problem.

![image](https://user-images.githubusercontent.com/17512851/71314703-a656a780-244d-11ea-891a-5c04855b7ca5.png)

PR #590, opened 2019-05-04
[FIX] hr_holidays_public: Traceback when change the date was fixed

luistorresm:
The field date in the public holidays cannot be changed when the field
to block the field cancellation is True.

With the attrs to block the field, if the record is not saved, the date
field is clear and do not allow to save the record because is a required
field. Then was improved with the onchange, to show an error message if
the date is changed and the check to block is True.

[Closes](https://github.com/OCA/hr/issues/569)

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -191,3 +191,14 @@ class HrHolidaysPublicLine(models.Model):
             raise UserError(_('You can\'t create duplicate public holiday '
                               'per date %s.') % self.date)
         return True
+
+    @api.onchange('date')
+    def onchange_date(self):
+        if not self.variable_date:
+            self.date = False
+            return {'warning': {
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -191,3 +191,14 @@ def _check_date_state_one(self):
...
         return True
+
+    @api.onchange('date')
+    def onchange_date(self):
+        if not self.variable_date:
+            self.date = False
+            return {'warning': {
+                'title': _('WARNING'),
+                'message': _(
+                    'The check that does not allow change the date is '
```
max3903:
```suggestion
                    'The check that does not allow to change the date is '
```

max3903 (approved):

luisiniguezh (approved):
:shipit: the change on the text @max3903 proposed helps to understand the message better

PR #595, opened 2019-05-09
[IMP] holidays_public: Return Intervals with holiday lines

etobella:
With this change, the leaves are Intervals with holidays as data. Now it can be detected the origin when processing it.
It might be interesting if a working interval should be affected by leaves but not by public holidays. like people that works on holidays (policeman, doctors, firemans ...)

### DIFF
--- a/hr_holidays_public/__manifest__.py
+++ b/hr_holidays_public/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'HR Holidays Public',
-    'version': '11.0.1.0.0',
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow (approved):
:+1: Technical review. Tested locally.

JordiBForgeFlow:
@hveficent  can you review this one? The change has no effect if you don't use the data that is introduced in the interval with this change.

jarroyomorales (approved):
:+1:

PR #596, opened 2019-05-13
[11.0][ADD] hr_holidays_allowed_period

jarroyomorales:
This module will allow the hr manager to determine in which date range a holiday type can be requested. For example Legal Leaves 2019 from 1/1/2019 to 31/12/2019. 
For now, selecting a date out of the holiday type's range will only display a red warning in the holiday request form since I wasn't sure if i should make it a constraint.


pedrobaeza:
Please, don't use plurals on module name. An a better name suggestion: `hr_holidays_allowed_period`.

jarroyomorales:
@pedrobaeza Okay, I already changed it.

JordiBForgeFlow (approved):
:+1: Technical and functional review. I made a small PR to improve the display of the warning message https://github.com/tegin/hr/pull/6

AaronHForgeFlow:
Is this the migration of this one? https://github.com/OCA/hr/tree/10.0/hr_holidays_validity_date

jarroyomorales:
@aheficent I didn't realize that there was already a module in 10.0 that adds this feature. I just did a PR with the migration of that one. https://github.com/OCA/hr/pull/598
Should I close this PR?

AaronHForgeFlow:
@jarroyomorales I am afraid we have to close this one, because both modules does the same thing. Nevertheless, I like the way this one shows a warning before the user clicks on save. Maybe we can add that to the other module.

jarroyomorales:
@aheficent Yes, that's the problem. I prefer to show a warning instead of a constraint blocking the selection holidays of out range but I dont know if I should change a feature during a migration...

AaronHForgeFlow:
If the change is a clear improvement of the module then it is fine to propose that in a migration.Perhaps we can propose that and let other people to share their opinions.

PR #597, opened 2019-05-16
[11.0] hr_employee_calendar_planning - improve usability

JordiBForgeFlow:
With this PR the employee's working calendar plan uses the full witdh of the page, for better usability.

image

### DIFF
--- a/hr_employee_calendar_planning/__manifest__.py
+++ b/hr_employee_calendar_planning/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Employee Calendar Planning",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "category": "Human Resources",
     "website": "https://github.com/OCA/hr",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
I don't see this as needed. I put it on purpose in the same place of previous working calendar.

I would say that you are even changing the page in which is contained.

jarroyomorales:
I liked how it was before but because I always use the `web_responsive` module. Anyways it is true that without it the lines are too small when editing so maybe we can think of another solution. Maybe keeping it on the side but without the label would give it more space and make it look nicer.

pedrobaeza:
But removing label will make it weird with the rest of the elements, and incorrect in mobile. This is OCA, everyone has `web_responsive`!

jarroyomorales:
Then I would keep it how it was

JordiBForgeFlow:
@pedrobaeza But the current view is almost unmanageable. What about a button to redirect the user to the calendar records in a separate view?

pedrobaeza:
I don't see it unmanageable. Tell me why. I don't see any need for the button.

JordiBForgeFlow:
OK! Reverting then.

PR #600, opened 2019-05-23
[Translation] Add French Translation

kallard1:
Add french translation


jarroyomorales:
Shouldn't you do it through the OCA weblate?

PR #603, opened 2019-05-27
[12.0][MIG] hr_holidays_compute_days

kittiu:
Work in progress


pedrobaeza:
Is this one still needed? I don't think so.

kittiu:
@pedrobaeza not need because that feature are already included in v12?

pedrobaeza:
Exactly

javimurcia:
Are you sure? We are having a discussion in the Spanish group because we cant include rest days in the holidays calculation (calculate in natural days instead of working days which is odoo's default), and if i'm not mistaken this module enables that feature.

See: https://groups.google.com/forum/#!topic/openerp-spain-users/HhsHxQjgcXk (in spanish)

pedrobaeza:
But this module was doing the contrary: to not include rest days, which was the default. As I said on the mailing list, we would need to add that feature, but not in this module, as the name doesn't reflect correctly what is doing (the name is an inherited one since long).

javimurcia:
Am i missing something?

- Default odoo behavior (version 11 & 12): count only working days, so saturday and sunday are left out, and a week "costs" 5 days ¿Right?
- What this module does: Adds the checkbox "Exclude rest days" to the leave configuration, default checked (same behavior as stock odoo). When the checkbox is unchecked, the leaves include all days in the calculation, so a week is 7 days long.

 And just to confirm, i've tested it in the runbot and my instance, everything under version 11, and just updated a few days ago (in the case of my instance) in all of the test worker has the default "Standard 40 Hours/Week", and try to schedule a leave begining 14/10/2019 (monday) 27/10/2019 (sunday), 2 weeks

Results:

- ODOO 11 from odoo SA runbot: **10 days**. it is only counting working days
- ODOO 11 with this module and "Exclude public holidays" & "Exclude rest days" checked: **10 days**, same behavior as "normal" odoo.
- ODOO 11 with this module and "Exclude public holidays" & "Exclude rest days" **unchecked**: 14 days, now it's counting natural days.

The last test had to be done with my instance, since i suspect there is something wrong with the demo data, since it was counting 19 days for the 2 weeks (great for the business, but workers will complain). My instance calculates the leave correctly as 14 days when I unset the checkbox (and 10 days if i leave it checked)

pedrobaeza:
Uhm, maybe you're right. I remembered the change being in v12, not in v11. Please confirm and then the module can continue with that. I have a special problem with the module name, which is not very explaining, but we can continue with it for better migration path.

javimurcia:
@pedrobaeza you are not alone, and the negative logic of "exclude rest days" doesn't make thing easier.

The name of the module, i suppose, came from its main function: compute the leave duration in full days, and this functions has been implemented in core. And somehow the module ended having also the function of computing the leave in natural days.

If i repeat the test in odoo's runbot for versions 10.0, 12.0 and 13.0 they all yield the same result: 10 days.

Maybe we can rename the module during migration (or maybe not, i don't know which impact would have this in openupgrade) since it's main function has changed.

While the module is being ported, if someone need this functionality urgently, they can edit the view for hr.leave and make the field "duration" writable again (is writable in every version except for 12) and they can calculate the leave duration by hand and put in that field.

pedrobaeza:
@kittiu are you going to handle this?

kittiu:
> @kittiu are you going to handle this?

Sure.

kittiu:
Moved to https://github.com/OCA/hr/pull/698

PR #606, opened 2019-05-28
[12.0][MIG] hr_employee_seniority

jarroyomorales:
Migration to 12.0 of hr_employee_seniority


JordiBForgeFlow (approved):
:+1: Functional and technical review.

nikul-serpentcs:
@jarroyomorales  Fix Transifex commits `OCA Transbot updated translations from Transifex`

jarroyomorales:
@nikul-serpentcs Done!

nikul-serpentcs (approved):
only Code Revie LGTM :+1:

etobella (changes requested):
There is a module on 12.0 that has a similiar functionality:
* `hr_employee_service` + `hr_employee_service_contract`

Maybe we should think to not migrate this module and use the new ones on 12.0 (maybe a migration script? or transform this module on a fake module (no views, no models) that migrates the dat?)
The only difference is that on seniority, the start_date can be defined (but is constrained by contracts), but hr_employee_service_contract defines the date as computed. Maybe a PR is necessary in order to make the same functionality than before without creating several modules.

What do you think? @alexey-pelykh @jbeficent

alexey-pelykh:
@etobella I guess that's a valid point, I may also add that `hr_holidays_accrual_advanced` also relies on `hr_employee_service` for accurate leaves allocation. As for contracts - if there's already a contract information, why to enter start date manually? I guess it can be made semi-computed but I see no usecase of that

etobella:
It's interesting for employees that started working a long time ago. 25 years ago, Odoo was not here, but some employees where (our oldest employee started working at 1982, so 37 years...)  Probably, we will not have all the contracts for him, but we have the date where he started. However, this is not blocking, we could live without this implementation.

alexey-pelykh:
@etobella sure, but what stops from inputting contracts from that time in Odoo? To keep the data integrity?

etobella:
I said this is not blocking because  we could do that :smile:

etobella:
As talked with @jbeficent we will add the change of name of module on OpenUpgrade
https://github.com/OCA/OpenUpgrade/blob/12.0/odoo/addons/openupgrade_records/lib/apriori.py#L5
And then, we will add a migration script on the `hr_employee_service (if necessary).
We will try to test a migration.

Tardo:
Hi, any progress here?

jarroyomorales:
@Tardo I'm working on it but I have some doubts regarding the migration with OpenUpgrade. The idea was to write the name change in the `apriory.py` file and then add the migration script in  `hr_employee_service_contract`. If i do so, will openupgrade install the new module and run its migration script?

jarroyomorales:
I think this PR can be closed and keep the discussion in #625

PR #609, opened 2019-05-28
Employee's anniverasy module

MarkingSC:
This module allows to use employment anniversary inside work information tab.

Note: The field is not computed by contract start date. It is just for informative purposes.


hr_employee_anniversary/models/hr_employee.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class HrEmployee(models.Model):
+
+    _inherit = "hr.employee"
+
+    # Anniversaries fields
```
eilst: unnecessary comment, the field name is self-explanatory

hr_employee_anniversary/__init__.py
```diff
@@ -0,0 +1,2 @@
+
+from . import models
```
eilst:
```suggestion
from . import models

```

hr_employee_anniversary/models/__init__.py
```diff
@@ -0,0 +1,2 @@
+
+from . import hr_employee
```
eilst:
```suggestion
from . import hr_employee

```

eilst (approved):

hr_employee_anniversary/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+        Adds a employment anniversary field next to birthday field.
+    """,
+
+    'author': "BerrySoft MX, Odoo Community Association (OCA)",
+    'website': "https://github.com/OCA/hr",
+
+    # Categories can be used to filter modules in modules listing
+    # for the full list
+    'category': 'Generic Modules/Human Resources',
+    'version': '0.1',
```
nikul-serpentcs: @MarkingSC IMO `'version': '0.1'`, -->` 'version': '12.0.1.0.0',` or If new module then ` 'version': '12.0.0.0.0',`

hr_employee_anniversary/views/hr_employee_views.xml
```diff
@@ -0,0 +1,18 @@
+
+<odoo>
+
+  <data>
```
nikul-serpentcs: no need to `data` tag

nikul-serpentcs (commented):
nitpicking

hr_employee_anniversary/__manifest__.py
MarkingSC: Thanks, I will solve that.

gurneyalex:
CLA is OK, I think. Problem could be comming from the commits on behalf of @eilst but he seems to have an ECLA so I'm not sure...

hr_employee_anniversary/views/hr_employee_views.xml
MarkingSC: Done, thanks.

jarroyomorales:
I think `hr_employee_seniority` already adds this feature. It allows you to set an initial employment date and also computes how long you have been working with the company. The module is on 11.0 but I already did the migration PR: https://github.com/OCA/hr/pull/606 to 12.0. If we are not talking about the same feature please tell me.

MarkingSC:
@jarroyomorales you are right. It's already satisfied with that module. Thanks.

PR #613, opened 2019-06-05
[FIX] Added general permission on hr_employee_skill and fixed service_duration error, field does not exist on view

mvasquez13:
When the logged user only has the field type_user configured as internal type, the employee form can not be opened because does not find the field service_duration on view.


pedrobaeza:
Thanks for your contribution, but please do it in English for this international repositories. Change the commit message for being in English, and only one module per PR unless related. Lastly, please sign the CLA. Thanks.

PR #617, opened 2019-06-10
[12.0][MIG] hr_expense_sequence (8.0->12.0)

kittiu:
WIP


pedrobaeza:
Please squash together adjacent "OCA Transbot.." (and the "[FIX] remove en.po that was..." one) commits using https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

LoisRForgeFlow:
Thanks for the work, I also think this could be an useful module.

Did you realize that it is actually available in v10? https://github.com/OCA/hr/commits/10.0/hr_expense_sequence

kittiu:
Oh thanks I didn't know, my bad. I will do both 11 and 12 then.
Thanks @lreficent

kittiu:
Will redo again in 10->11 and 10->12

PR #623, opened 2019-06-13
[FIX] hr_expense: Allow a user to attach an invoice in draft

gustavovalverde:
If the accounting team is the one in charge of validating invoices (users can't validate), then the users won't be able to attach invoices made by them from the expense app.

This also fixes a weird behavior as you can can hit the create and edit button; complete the invoice and save it; but then it won't be shown in the selection, but you can't neither search it.

### DIFF
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -11,7 +11,7 @@ class HrExpense(models.Model):
     invoice_id = fields.Many2one(
         comodel_name='account.invoice',
         string='Vendor Bill',
-        domain="[('type', '=', 'in_invoice'), ('state', '=', 'open')]",
+        domain="[('type', '=', 'in_invoice'), ('state', 'in', ('draft', 'open'))]",
         oldname='invoice',
     )
 


gaikaz (approved):
LGTM
@pedrobaeza Seems like you are the original contributor of the module and this domain attribute comes all the way from your original commit https://github.com/OCA/hr/commit/88bdb29fce2940f17c2ed3f8561c6e6846a62e14#diff-0a613f418c3ba0a249dc5adf298aa05aR69
Could you please review this too? Maybe there was a good reason for draft invoices to be filtered out?

pedrobaeza:
This is incorrect, as draft invoice doesn't have still any journal entry, so you can't finish the flow in that state. That's why I limited this.

:-1: in this condition.

gaikaz:
But then it would be nice to find a solution for the described odd behavior:
expenses_draf_invoice_problem

pedrobaeza:
It's not odd, it's because draft invoices are not allowed, and Odoo empty the field when you try to pass one through that "trick", which is good. Any way, the solution for not being counter intuitive is to not allow to create invoices from that many2one (add `create="0"`).

Issue #629, opened 2019-06-25
[12.0] Travis failing


jarroyomorales:
I don't know if there has been a recent odoo change but some tests are failing. They were working fine for me until I updated odoo.

The error is always the same, requesting holidays before the `validity_start` field and the solution is simple, setting it to false on the tests.
For now I have detected it in 'hr_holidays_accrual_advanced` and `hr_holidays_leave_repeated` but it may happen in other modules.

I'm already solving it but I don't know if it is the best solution, thank you!

pedrobaeza:
Don't open issue if you are going to open a PR afterwards.

PR #630, opened 2019-06-25
[12.0][FIX] Travis Failing Holidays Leave Type Validity

jarroyomorales:
This PR intends to solve issue #629


alexey-pelykh (approved):

etobella (approved):
This issue is a consequence of odoo/odoo#34225 merged 7 days ago.

LGTM, but it is an issue to modify several addons on the same PR?

pedrobaeza:
Good to know the source of the problem. It's not a problem to modify several modules, as the source problem is the same. I'll add to the commit message that link of the source.

PR #634, opened 2019-07-01
[12.0][FIX] hr_holidays_accrual_advanced: tests and more logs

alexey-pelykh:


alexey-pelykh:
@pedrobaeza please fast-track if possible. Changes made:
 * one test input params changed since if 1 day delta is likely to be on weekend, so tests are failing 3 days in a week out of 7
 * Added more logs about how calculation is happening
 * Moved one code block a bit lower to have data to put to log - does not affect the calculation

pedrobaeza (changes requested):
I prefer to not have that logs always hammering while only needed by debugging purposes. Put them under `debug` level or remove them at all.

alexey-pelykh:
@pedrobaeza replaced all except one with `debug`

pedrobaeza (approved):

pedrobaeza:
I still think it's too much code added for only maintaining some kind of debug information...

PR #636, opened 2019-07-02
[IMP] add attribute name to page to improve the search in inheritance

mvasquez13:


etobella:
Why do you change permissions?

mvasquez13:
> Why do you change permissions?

Hi, because a person with a basic permission should not have access to this kind of personal information.

etobella:
Well, the commit is referring to the change of the name page, not the permissions... Also, the permissions of the module allows all users to access this information. I think that you should split it in two PRs, as the change of permissions should be validated for more contributors than this.

astirpe (changes requested):
@limfit changing the permissions the way you did (in the view) is not enough, since you can read those sensitive information in other ways, eg.: filtering, searches, etc...
Please only add the attribute name within this PR, as @etobella said. Thanks!

PR #637, opened 2019-07-02
[IMP] add permission to model hr_employee_skill and attribute name to…

mvasquez13:
… page to improve the search in inheritance

### DIFF
--- a/hr_skill/views/hr_employee.xml
+++ b/hr_skill/views/hr_employee.xml
@@ -12,7 +12,7 @@
         <field name="inherit_id" ref="hr.view_employee_form"/>
         <field name="arch" type="xml">
             <notebook position="inside">
-                <page string="Skills">
+                <page string="Skills" name="employee_skills" groups="hr.group_hr_user">
                     <group>
                         <field name="employee_skill_ids" nolabel="1">
                             <tree editable="bottom">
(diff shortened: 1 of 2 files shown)


hr_skill/views/hr_employee.xml
```diff
@@ -12,7 +12,7 @@
         <field name="inherit_id" ref="hr.view_employee_form"/>
         <field name="arch" type="xml">
             <notebook position="inside">
-                <page string="Skills">
+                <page string="Skills" name="employee_skills" groups="hr.group_hr_user">
                     <group>
```
nikul-serpentcs, line 16: @limfit if you don't mind, here Add Group name also.

hr_skill/views/hr_employee.xml
```diff
@@ -12,7 +12,7 @@
         <field name="inherit_id" ref="hr.view_employee_form"/>
         <field name="arch" type="xml">
             <notebook position="inside">
-                <page string="Skills">
+                <page string="Skills" name="employee_skills" groups="hr.group_hr_user">
```
etobella, line 15: Why don't you add the permissions directly to the view in order to not load the fields?

mvasquez13: Because there are some public information outside this page should have visible to a person with a basic permission

etobella: I know, but your code only adds this part. It is a right approach also.

With my comment, the previous view would work perfectly for users outside the group and will be visible only for this one. The modification of the view will be applied only for the users of the group

PR #639, opened 2019-07-16
[IMP] hr_holiday_compute_days: Remove unneeded field

pedrobaeza:
Making use of force_save new attribute, we can save a new field + make smaller view definition.

cc @Tecnativa


pedrobaeza:
Working as expected on runbot, so merging

/ocabot merge minor

PR #640, opened 2019-07-20
[12.0][FIX] hr_holidays_settings: duplicate Leave Types menuitem

alexey-pelykh:
image

### DIFF
--- a/hr_holidays_settings/views/res_config_settings.xml
+++ b/hr_holidays_settings/views/res_config_settings.xml
@@ -30,9 +30,4 @@
           action="hr_holidays_config_settings_action"
           groups="base.group_system"/>
 
-    <menuitem id="menu_hr_holidays_status" name="Leave Types"
-          parent="hr_holidays.menu_hr_holidays_configuration" sequence="10"
-          action="hr_holidays.open_view_holiday_status"
-          groups="base.group_system"/>
-
 </odoo>
--- a/hr_holidays_settings/__manifest__.py
+++ b/hr_holidays_settings/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "HR Holidays Settings",
     "summary": "Enables Settings Form for HR Holidays.",
-    "version": "12.0.1.0.0",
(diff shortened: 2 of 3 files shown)


alexey-pelykh:
@tarteo you're most involved with the lines changed, it would be great to get your feedback, if you don't mind

hr_holidays_settings/views/res_config_settings.xml
```diff
@@ -30,9 +30,8 @@
           action="hr_holidays_config_settings_action"
           groups="base.group_system"/>
 
-    <menuitem id="menu_hr_holidays_status" name="Leave Types"
-          parent="hr_holidays.menu_hr_holidays_configuration" sequence="10"
-          action="hr_holidays.open_view_holiday_status"
-          groups="base.group_system"/>
+    <record model="ir.ui.menu" id="hr_holidays.hr_holidays_status_menu_configuration">
```
astirpe: I would rather remove completely this "Leave Types" code from here and keep only the standard behavior.

astirpe (changes requested):

hr_holidays_settings/views/res_config_settings.xml
alexey-pelykh: That would be behaviour change and frankly speaking I have no preference over this. Yet not a fix totally, that should be addressed as separate PR

astirpe: In V11 the "Configuration" menuitem was opening the "Leave Types" directly (the explicit "Leave Types" menuitem was missing). That's why this module was overriding the two menuitems.

But now in V12, a dedicated "Leave Types" menuitem is present and the "Configuration" menuitem is just a parent of it. So this block of code is now obsolete. Is just a leftover from the migration. The correct fix is to remove it, not changing its group (there's no need to change its group). Keeping the standard behavior is fine.

alexey-pelykh: If you say so :) Thanks for details :)

astirpe (approved):

PR #643, opened 2019-07-23
[12.0][IMP] hr_employee_service: support _get_date_start_work

alexey-pelykh:
To support https://github.com/odoo/odoo/commit/90731ad170c503cdfe89a9998fa1d1e2a5035c86

### DIFF
--- a/hr_employee_service/models/hr_employee.py
+++ b/hr_employee_service/models/hr_employee.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+# Copyright (C) 2018-2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import models, fields, api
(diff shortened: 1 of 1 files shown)


jarroyomorales (approved):
Nice, I dont like the odoo's function since the date an employee is created and the real service start day arent always the same and this solves it.

hr_employee_service/models/hr_employee.py
```diff
@@ -107,3 +107,7 @@ def _compute_service_duration_display(self):
     def _onchange_service_hire_date(self):
         if not self.service_start_date:
             self.service_start_date = self.service_hire_date
+
+    # NOTE: Support odoo/odoo@90731ad170c503cdfe89a9998fa1d1e2a5035c86
+    def _get_date_start_work(self):  # pragma: no cover
```
pedrobaeza: Don't trick coverage on this

alexey-pelykh: Removed

pedrobaeza (approved):

Issue #646, opened 2019-07-24
[12.0][BUG] hr_holidays_public: error when inputting Date of Leave Request is Public Holiday have existed on the other Leave Request


cuongpk:
![image](https://user-images.githubusercontent.com/23528443/61784854-ad44a400-ae34-11e9-889d-810057bbc970.png)

Please help me to check this issue.
I think we need to call the function check the overlapping date before `_onchange_leave_dates` called.

pedrobaeza:
Please indicate clear steps to reproduce the problem

cuongpk:
> Please indicate clear steps to reproduce the problem
![image](https://user-images.githubusercontent.com/23528443/61782499-55a43980-ae30-11e9-83b0-edb46aab2143.png)
This is my Public holiday.
![image](https://user-images.githubusercontent.com/23528443/61782516-5b9a1a80-ae30-11e9-9e59-0c2cd7c1d7bd.png)
And I have created a new Leave Request, it's still good this step.
![image](https://user-images.githubusercontent.com/23528443/61784873-ba619300-ae34-11e9-9ac0-b17a78edb9d8.png)
Then, I have tried to create a new Leave Request with the same Public Holiday and I get this error.

jarroyomorales:
Well, by looking at the logs you can clearly see that the problem comes from trying to create an interval from an interval that comes from a `resource.calendar.leave` and another one that comes from a public holiday. This didn't happen in 11.0 because intervals were appended using + instead of | , which is redefined by the `Intervals` class.
By changing https://github.com/OCA/hr/blob/12.0/hr_holidays_public/models/resource_calendar.py#L44
 to an empty resource.calendar.leave (`self.env['resource.calendar.leaves']`) you fix the bug, but you lose the reference to the public holiday and it may have side effects.
On the migration to 12.0 I have seen that @alexey-pelykh and @pedrobaeza talked about assigning a `resource.calendar.leave` to every public holiday, which may be the best solution here.
What do you think?

meigallodixital:
I made a leave (holiday) from July 12 to 31 and 25 as a holiday and I confirm the calculation fails in theoretical too.

Traceback: https://pastebin.com/aYENSuFq

pedrobaeza:
@alexey-pelykh can you please check this?

ivantodorovich:
Following up on the discussion in https://github.com/OCA/hr/pull/534#issuecomment-449573880

[...]

PR #649, opened 2019-07-30
[12.0][IMP] hr_holidays_public: Create global calendar.event

ivantodorovich:
Notice there are 2 commits:

- The first one is just making the code a little bit more readable. No big changes there, but many lines are affected.

- The second one is the actual IMP

With this PR, public holidays will create events on the global calendar.
This is aligned with Odoo's creation of calendar events for hr.leaves, when validated.

Note that:
- The name of the event will be the name of the holiday.
- There's a "Holidays" tag applied to the event.
- The event will appear under the global calendar (Everybody's calendars)
- Since the event is created by the system, the user responsible is 'OdooBot' (System user)
- The event will be updated if the public holiday is updated
- The event will be deleted if the public holiday is deleted

image

Maybe a known issue would be that users can edit the calendar event and the changes will not be blocked nor synced to the public holiday. I didn't want to get into this because that's the same behaviour we have with the events created for hr.leaves.

### DIFF
--- a/hr_holidays_public/models/hr_holidays_public.py
+++ b/hr_holidays_public/models/hr_holidays_public.py
@@ -4,6 +4,7 @@
 from datetime import date
 
 from odoo import api, fields, models, _
+from odoo import SUPERUSER_ID
 from odoo.exceptions import ValidationError
 
 
@@ -158,6 +159,8 @@ class HrHolidaysPublicLine(models.Model):
         'Related States'
     )
 
+    meeting_id = fields.Many2one('calendar.event', string='Meeting')
+
     @api.multi
     @api.constrains('date', 'state_ids')
     def _check_date_state(self):
@@ -195,3 +198,44 @@ class HrHolidaysPublicLine(models.Model):
                 'You can\'t create duplicate public holiday per date %s.'
             ) % self.date)
         return True
+
+    @api.multi
+    def _prepare_holidays_meeting_values(self):
+        self.ensure_one()
+        categ_id = self.env.ref('hr_holidays_public.event_type_holiday', False)
+        meeting_values = {
+            'name': (
+                '%s (%s)' % (
+                    self.name,
+                    self.year_id.country_id.name
+                ) if self.year_id.country_id else self.name),
+            'description': ', '.join(self.state_ids.mapped('name')),
+            'categ_ids': [(6, 0, categ_id.ids if categ_id else [])],
+            'start': self.date,
+            'stop': self.date,
+            'allday': True,
+            'partner_ids': False,
+            'user_id': SUPERUSER_ID,
+            'state': 'open',
+            'privacy': 'confidential',
+            'show_as': 'busy',
+        }
+        return meeting_values
+
+    @api.constrains('date', 'name', 'year_id', 'state_ids')
+    def _update_calendar_event(self):
+        for rec in self:
+            if rec.meeting_id:
+                rec.meeting_id.write(rec._prepare_holidays_meeting_values())
+
+    @api.model
+    def create(self, values):
+        res = super().create(values)
+        res.meeting_id = self.env['calendar.event'].create(
+            res._prepare_holidays_meeting_values())
+        return res
+
+    @api.multi
+    def unlink(self):
+        self.mapped('meeting_id').unlink()
+        return super().unlink()
--- a/hr_holidays_public/README.rst
+++ b/hr_holidays_public/README.rst
@@ -74,6 +74,15 @@ This may or may not be a problem, yet since this component is also being set to
 `resource.calendar.attendance` records in `_attendance_intervals()`, seems it
 should be ok.
 
+There are no restrictions to block users from modifying or removing calendar
+events linked to public holidays. There's a suggestion to overload `write` and
+`unlink` methods of `calendar.event`, but it might have other impacts like
+users not being able to edit event tags, or even custom fields.
+
+Regional public holidays are shown in the public calendar. The regions will be
+noted in the description of the event, but it'll be shown to all users. It'd
+be good to have it show only for users in these regions.
+
 Bug Tracker
 ===========
 
@@ -93,6 +102,7 @@ Authors
 * Michael Telahun Makonnen
 * Tecnativa
 * Fekete Mihai (Forest and Biomass Services Romania)
+* Druidoo
 
 Contributors
 ~~~~~~~~~~~~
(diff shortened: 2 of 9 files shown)


hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -135,26 +140,30 @@ class HrHolidaysPublicLine(models.Model):
         'Name',
         required=True,
     )
+
```
pedrobaeza: Please don't add useless empty lines. It's discouraged by PEP8 and doesn't increase readability. Indentation does that.

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -57,8 +62,7 @@ def _compute_display_name(self):
             if line.country_id:
                 line.display_name = '%s (%s)' % (
                     line.year,
-                    line.country_id.name
-                )
+                    line.country_id.name)
```
pedrobaeza: Don't change ending characters deindented in other line. This is for readability, as you are doing for example in https://github.com/OCA/hr/pull/649/commits/a23796e805c8b65e228c90190bf43532263c1a69#diff-3080bf617f93d891bde828f3edd165d3R95

hr_holidays_public/models/hr_leave_type.py
```diff
@@ -10,9 +11,7 @@ class HrLeaveType(models.Model):
 
     exclude_public_holidays = fields.Boolean(
         string='Exclude Public Holidays',
+        help='If enabled, public holidays are skipped in leave days'
```
pedrobaeza: No added value

hr_holidays_public/models/hr_holidays_public.py
ivantodorovich: Noted!

hr_holidays_public/models/hr_holidays_public.py
ivantodorovich: Yeap, don't even know why I did that

hr_holidays_public/models/hr_leave_type.py
ivantodorovich: Reverted.

alexey-pelykh:
@ivantodorovich please rebase on latest 12.0, there was a fix-merge recently, and please tidy up the commit history

alexey-pelykh (commented):
please rebase on latest 12.0, there was a fix-merge recently, and please tidy up the commit history

alexey-pelykh:
> Maybe a known issue would be that users can edit the calendar event and the changes will not be blocked nor synced to the public holiday. I didn't want to get into this because that's the same behaviour we have with the events created for hr.leaves.

@ivantodorovich can we block editing of such events entirely? If event is linked to a holiday, limit its `write` method and show error. But I guess it should be done via separate PR if anyone wishes to

pedrobaeza:
@ivantodorovich thanks for this. For me, taking Odoo's example of letting users to edit the generated entries is not a good reason for doing the same here, as we are in control of all elements. What you can do is:

* Add `public_holiday_id` many2one at calendar event level.
* Fill it in these entries.
* Overwrite `write` and `unlink` methods for triggering an exception when the m2o is filled unless an specific context is passed.
* When removing public holidays, call `unlink` of the calendar event with this context.
* The same when changing the public holidays.

What do you think?

Another doubt I have. What happen with regional public holidays? Are they put in the calendar as global?

ivantodorovich:
> please rebase on latest 12.0, there was a fix-merge recently, and please tidy up the commit history

Done!

I'm not sure we should block writing on it, maybe the user wants to change things like tags, descriptions, or even custom fields.. and that should be ok. So in that case a constraint restricting specific fields like date and name makes more sense to me.

Anyhow I honestly don't have much time these days, this was requested and financed by a customer as it is. I've added a note in the ROADMAP / Known issues file, so that we don't lose track of this, and we can implement that in a another PR later.

@pedrobaeza I've also added the note regarding regional public holidays.

> There are no restrictions to block users from modifying or removing calendar
events linked to public holidays. There's a suggestion to overload `write` and
`unlink` methods of `calendar.event`, but it might have other impacts like
users not being able to edit event tags, or even custom fields.

> Regional public holidays are shown in the public calendar. The regions will be
noted in the description of the event, but it'll be shown to all users. It'd
be good to have it show only for users in these regions.

pedrobaeza:
OK, I understand if you are not able to add that withing budget. A note in roadmap is enough for now.

alexey-pelykh:
I'm looking into #646 right now, it seems that if to implement creating global leave for each public holiday, events are going to be there in any case?

ivantodorovich:
@alexey-pelykh ~~yeap, they will be generated by the `hr.leave` model.~~
Let's discuss that in #646

ivantodorovich:
Actually, just realized that @jarroyomorales is talking about the `resource.calendar.leaves` model, that won't generate any event in the calendar.

hr_holidays_public/models/hr_holidays_public.py
```diff
@@ -190,3 +193,43 @@ def _check_date_state_one(self):
...
+    @api.model
+    def create(self, values):
+        res = super().create(values)
+        res.meeting_id = self.env['calendar.event'].create(
+            res._prepare_holidays_meeting_values())
+        return res
+
+    @api.multi
+    def unlink(self):
+        self.mapped('meeting_id').unlink()
```
jarroyomorales: Here the `return super().unlink()` is missing

jarroyomorales:
@alexey-pelykh @ivantodorovich I think both things make sense and are compatible. On one hand we have the public holidays generating `resource.calendar.leaves` to solve some bugs mentioned in #646 and on the other hand we have this PR which looks absolutely fine for me since we would be doing the same than Odoo: Creating the leave -> Creating the `resource.calendar.leave` -> Creating the event.

jarroyomorales (approved):

ivantodorovich:
@jarroyomorales 
> Here the return super().unlink() is missing

Thanks!!!! Just sent the fixup

pedrobaeza (changes requested):
Please rebase squashing your fixup commit and removing that ugly merge commit that will scramble future migrations.

I see also some code that seems a bugfix for another feature (the one in hr_leave). Please put that in another PR.

ivantodorovich:
> Please rebase squashing your fixup commit and removing that ugly merge commit that will scramble future migrations.
> 
> I see also some code that seems a bugfix for another feature (the one in hr_leave). Please put that in another PR.

Upps.. sorry about that, it's done!

pedrobaeza (approved):
I consider fair to add Druidoo as co-author given the kind of contribution.

pedrobaeza:
Can you please migrate this to v13 as well? Thanks!

ivantodorovich:
> Can you please migrate this to v13 as well? Thanks!

Thanks @pedrobaeza ! :)
13.0 PR: https://github.com/OCA/hr-holidays/pull/2

PR #652, opened 2019-08-01
[11.0][ADD] New module: hr_holidays_extendable

jarroyomorales:
Hr Holidays Extendable
======================

Allows a certain type of holidays to be extended.

A much easier solution would be to just create a new leave, but this module can
be interesting if you want to keep the chatter and everything in one record.
It is also more realistic, since at the end, you are having just one leave,
not two or three.

Usage
=====

The first thing to do is select which holiday types can be extended.
To do so go the configuration menu in the Leaves app, and select the check
box *Extendable*.

Holidays can only be extended by the Human Resources Manager and when they are
in *Approved* state. A new button will appear automatically when all these
conditions are met.

Clicking on it will open a small wizard that will allow you to select the new
end date for the holiday. After clicking *Extend* the original holiday will be
updated with the new values.


jarroyomorales:
Waiting for #653

jarroyomorales:
Green :green_apple:

jarroyomorales:
ping @etobella

etobella (approved):
Interesting, I think it would be better to add the button on the "button_box", but there is no button box here. Maybe we could finish earlier the holidays. Would be possible to do this here or it is better to add it on a different addon?

jarroyomorales:
@etobella I'm trying to convert this module into `hr_holidays_variable_end_date` or something like this. The problem is that I can't find a way of computing the new number of days on the wizard without refusing and validating the holiday again (in the case where you reduce the end_date). `_get_number_of_days()` always returns 0...

PR #655, opened 2019-08-06
[12.0][FIX] hr_holidays_leave_request_wizard: used fields must be in XML

alexey-pelykh:
Workarounds https://github.com/odoo/odoo/issues/35419

### DIFF
--- a/hr_holidays_leave_request_wizard/wizards/hr_leave_wizard.py
+++ b/hr_holidays_leave_request_wizard/wizards/hr_leave_wizard.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import babel.dates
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think you are facing https://github.com/odoo/odoo/issues/20163

alexey-pelykh:
@pedrobaeza I doubt that, since it was working when it was tested and merged couple of months ago

pedrobaeza:
Well, I don't think this big patch is the solution either, and you are declaring incorrectly the fields in the one2many. You must put either `<tree>` or `<form>` in between. What is clear is that non nested views are a problem, so nest the views of children models.

alexey-pelykh:
Yeah, this fix still does not fix anything :( And nested views are not pulled at all, e.g. field of one2many won't use XML declaration at all. Right now not sure how to proceed with actual fix

Issue #653, opened 2019-08-01
[11.0] hr_holidays_compute days, fields are True by default


jarroyomorales:
The new field `compute_full_days` in this module is set as True by default, which can (and I think it is) cause problems to other modules tests since they shouldn't take in account that the computation of days has been changed by this module.

https://github.com/OCA/hr/blob/11.0/hr_holidays_compute_days/models/hr_holidays_status.py#L19

jarroyomorales:
@pedrobaeza What do you think of this? Should I create a PR?

jarroyomorales:
And I think it is the same case with the other field, exclude_rest_days`. 
A field that changes the result of the number of days computation shouldn't be default True because other modules can't do anything against it

pedrobaeza:
Well, it shouldn't be too much problem, as you normally do that for the leaves. Any way, you can add code for inhibiting that behavior on the module itself detecting if you are in test mode and not testing the module itself.

jarroyomorales:
The module itself works fine, the problem is just with the tests. You can always ask if the leave type has the field `compute_full_days` and set it to False but I dont know if thats the best solution since these two modules are independent.

pedrobaeza:
As said, you can inhibit the behavior when testing other modules:

1. Add `from odoo.tools import config`.
2. On the code that changes the behavior if `compute_full_days`, check if `config['test_enable']` and there's not a specific context key. Then ignore behavior.
3. On `hr_holidays_compute_days` module tests, add that context.

Example of this: https://github.com/OCA/bank-payment/blob/8.0/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

jarroyomorales:
Solved! At least in 11.0, should we also apply this to 12.0? @pedrobaeza

pedrobaeza:
Yes, please propose PR

jarroyomorales:
The PR of the migration https://github.com/OCA/hr/pull/603 was closed since this module is already included in odoo 12.0 core so I think we can close this issue.

PR #656, opened 2019-08-06
[11.0][IMP] hr_holidays_compute_days: Add test context

jarroyomorales:
This PR intends to solve #653 by adding a context variable that makes other modules ignore the function that is returning unexpected values.

@pedrobaeza

### DIFF
--- a/hr_holidays_compute_days/tests/test_holidays_compute_days.py
+++ b/hr_holidays_compute_days/tests/test_holidays_compute_days.py
@@ -11,7 +11,9 @@ class TestHolidaysComputeDaysBase(common.SavepointCase):
     @classmethod
     def setUpClass(cls):
         super(TestHolidaysComputeDaysBase, cls).setUpClass()
-        cls.HrHolidays = cls.env['hr.holidays']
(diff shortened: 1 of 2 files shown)


hr_holidays_compute_days/tests/test_holidays_compute_days.py
```diff
@@ -75,7 +75,7 @@ def test_number_days_excluding_full_day_employee(self):
             'date_to': '1946-12-27 11:00:00',  # Friday
             'holiday_status_id': self.holiday_type.id,
             'employee_id': self.employee.id,
-        })
+        }).with_context(test_full_days=True)
```
pedrobaeza: Instead of here, add the context in the object `self.HrHolidays` for having it everywhere, and no need to put it in any other place.

jarroyomorales: Done!

pedrobaeza (approved):

jarroyomorales:
ping @etobella

etobella (approved):

PR #666, opened 2019-08-29
[12.0][FIX] hr_employee_service: Test 3

jarroyomorales:
Test 3 created an employee with a hire date 6 months previous than today and then tested that the computation was correct and the employee had a length of service of exactly 6 months.
Today it is `2019-08-29` and 6 months ago it should be `2019-02-29` but since it was February, python's `relativedelta` returns `2019-02-28` which makes the length of service 6 months and 1 day long, which made the test fail.
I know this only happens once a year and tomorrow tests will work properly but I this this little fix is good to have anyways.

### DIFF
--- a/hr_employee_service/tests/test_hr_employee_service.py
+++ b/hr_employee_service/tests/test_hr_employee_service.py
@@ -5,6 +5,8 @@ from odoo import fields
 from odoo.tests import common
 
 from dateutil.relativedelta import relativedelta
+from datetime import date
+from mock import patch
 
 
 class TestHrEmployeeService(common.TransactionCase):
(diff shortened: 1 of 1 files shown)


alexey-pelykh:
@jarroyomorales I guess we can simply use `service_termination_date`? But that would duplicate other test, so meaningless. I'd just cut down number of tests

If we want to keep that test, just mock the today for calls to compute OR mock today for entire method?

alexey-pelykh:
test 3, PR 666 🤣

jarroyomorales:
If you prefer to reduce the number of tests it is ok for me, I dont know if the same can happen some day with other tests...
Yes hahaha, I also realized about the 666 :rofl:

alexey-pelykh:
@jarroyomorales I think that testing a single `or` is too much, I guess I was in mood of testing everything 🤷‍♂ Yet I'd do it like this now:
```
mocked_today = date(2019, 8, 27)
employee = self.SudoEmployee.create({
    'name': 'Employee #3',
    'service_hire_date': (
        mocked_today - relativedelta(months=6)
    ),
    'service_start_date': (
        mocked_today - relativedelta(months=6)
    ),
})

with patch('odoo.fields.Date.today') as today:
    today.return_value = mocked_today
    self.assertEqual(employee.service_duration_years, 0)
    self.assertEqual(employee.service_duration_months, 6)
    self.assertEqual(employee.service_duration_days, 0)

```

jarroyomorales:
It is the same I did, isnt it?

alexey-pelykh:
Almost, `create()` may also use today for whatever reason and we may have side effect. But I'm only suggesting

jarroyomorales:
Okay! We can go with your solution. Thank you!

hr_employee_service/tests/test_hr_employee_service.py
```diff
@@ -51,19 +53,22 @@ def test_2(self):
         self.assertEqual(employee.service_duration_days, 0)
 
     def test_3(self):
+        mocked_today = date(2019, 8, 27)
         employee = self.SudoEmployee.create({
             'name': 'Employee #3',
             'service_hire_date': (
-                self.today - relativedelta(months=6)
+                    mocked_today - relativedelta(months=6)
```
alexey-pelykh: Formatting off

alexey-pelykh (approved):

alexey-pelykh:
Waiting for CI to complete and will merge

PR #672, opened 2019-09-19
[12.0][ADD]hr_expense_petty_cash

ps-tubtim:
@kittiu


hr_expense_petty_cash/data/account_account.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+
+    <record id="petty_cash" model="account.account">
```
kittiu: We don't need to provide the account code. Accountant should be the person to add this manually into his/her chart of account.
This file can be removed.
Note: in the test script, you can use coding to create this account before run test.

hr_expense_petty_cash/models/account_invoice.py
```diff
@@ -0,0 +1,78 @@
...
+            balance = petty_cash.petty_cash_balance
+            limit = petty_cash.petty_cash_limit
+            max_amount = limit - balance
+            account = petty_cash.account_id
+            # Looking for petty cash line
+            inv_lines = rec.invoice_line_ids.filtered(
+                lambda l: l.account_id == account)
+            amount = sum([line.price_subtotal for line in inv_lines])
+            company_currency = rec.company_id.currency_id
+            amount_currency = rec.currency_id.compute(amount, company_currency)
```
kittiu:
```suggestion
            amount_company = rec.currency_id.compute(amount, company_currency)
```
I understand that you try to change from any currency (USD), into company currency (THB), right?
If so, amount in company currency, should named, amount_company. This is naming convention you can find in account.move.line, i.e.,
* debit, credit = company currency always, i.e., THB (else where it will be amount_company)
* amount_currency = foreign currency, i.e., USD

hr_expense_petty_cash/models/account_invoice.py
```diff
@@ -0,0 +1,78 @@
...
+    def _add_petty_cash_invoice_line(self, petty_cash):
+        # Get suggested currency amount
+        amount = petty_cash.petty_cash_limit - petty_cash.petty_cash_balance
+        company_currency = self.env.user.company_id.currency_id
+        amount_doc_currency = \
+            company_currency.compute(amount, self.currency_id)
+
+        inv_line = self.env['account.invoice.line'].new({
+            'name': petty_cash.account_id.name,
+            'invoice_id': self.id,
```
kittiu, line 56:
```suggestion
```
I think this is not necessary, can you remove this line and test whether it works?
Note: because when we add line to o2m, and save it from UI, it auto assign invoice_id.

hr_expense_petty_cash/models/hr_expense.py
```diff
@@ -0,0 +1,47 @@
...
+    def action_submit_expenses(self):
+        res = super().action_submit_expenses()
+        context = res.get('context', False)
+        context['default_expense_line_ids'] = self.ids
+        context['default_name'] = self[0].name if len(self) == 1 else ''
+        set_petty_cash_id = set()
+        for rec in self:
+            if rec.payment_mode == 'petty_cash':
+                set_petty_cash_id.add(rec.petty_cash_id.id)
+        if len(set_petty_cash_id) > 1:
```
kittiu, line 32: What if they use the same petty cash holder, but create multiple petty cash expense? I think, it should be ok, right?
This seem to test only number of petty cash expense, and not yet cover the case mentioned. Please verify.

hr_expense_petty_cash/models/account_invoice.py
```diff
@@ -0,0 +1,78 @@
...
+            'account_id': petty_cash.account_id.id,
```
kittiu, line 57: As I asked to remove the Petty Cash Account data, then, please check if petty_cash.account_id == False.
And give a very good detail on how to correctly add new Petty Cash Account in chart of account. (as guideline here)

hr_expense_petty_cash/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,17 @@
...
+class HrExpenseSheet(models.Model):
+    _inherit = 'hr.expense.sheet'
+
+    payment_mode = fields.Selection(
+        selection_add=[('petty_cash', 'Petty Cash')],
+    )
+    petty_cash_id = fields.Many2one(
+        string='Petty cash holder',
+        comodel_name='petty.cash',
+        readonly=True,
```
kittiu, line 18: add `ondelete='restrict',`

hr_expense_petty_cash/models/hr_expense.py
```diff
@@ -0,0 +1,47 @@
...
+    _inherit = 'hr.expense'
+
+    payment_mode = fields.Selection(
+        selection_add=[('petty_cash', 'Petty Cash')],
+    )
+    petty_cash_id = fields.Many2one(
+        string='Petty cash holder',
+        comodel_name='petty.cash',
+        readonly=True,
+        states={'draft': [('readonly', False)]},
```
kittiu, line 19: add `ondelete='restrict',`
Note: normally, please add this when you create any  many2one, because some time we want it differently, i.e., those line_ids, we may want `ondelete='cascade'`, etc.

kittiu:
@ps-tubtim looks good so far. Please fix per review, I will do functional test again.
Also, please add test case.

hr_expense_petty_cash/models/account_invoice.py
ps-tubtim: If remove this line, invoice line can't auto appear.

hr_expense_petty_cash/models/account_invoice.py
ps-tubtim: **Create a Petty Cash Account**

#. Go to Invoicing > Configuration > Chart of accounts
#. Create a new petty cash account
#. Type = Bank and Cash or Account type that is asset

Note:
* You will need the "Show Full Accounting Features" to see accounting data

hr_expense_petty_cash/views/petty_cash_views.xml
```diff
@@ -0,0 +1,90 @@
...
+        <field name="model">petty.cash</field>
+        <field name="arch" type="xml">
+            <form string="Petty Cash">
+               <sheet>
+                   <group>
+                       <group>
+                           <field name="partner_id"/>
+                           <field name="account_id"/>
+                           <field name="petty_cash_limit"/>
+                           <field name="petty_cash_balance" readonly="1"/>
```
kittiu:
```suggestion
                           <field name="petty_cash_balance"/>
```
readonly is not necessary, required field will be readonly anyway.

kittiu:
On "Submit to manager", "Approve" and "Post Journal Entries", the expense sheet, there should be warning message, i..e, 
```
Not enough money in petty cash holder.
You are requesting 1,000฿, but the balance is 100฿
```
Now, there is no blocking, and so, it allow to have negative balance.
image

kittiu:
In petty cash form, let's move like this,
image

kittiu:
Please add Group By: Petty Cash Holder

kittiu:
@ps-tubtim So far so good.
So, please add some test cases as final step.
Thank you!

ps-tubtim:
Please go to https://github.com/OCA/hr/pull/679

PR #686, opened 2019-10-01
[13.0][MIG][WIP] hr_expense_advance_clearing

kongrattapong:
I need help.

I stuck in condition: `if line.account_id == emp_advance.property_account_expense_id:`
In this condition it check 4 Payments and should match 2 Payments: debit & credit but match only 1 Payment (debit) How can I fix it?

Step:
- Create new expense with Employee Advance (Expense Account is 1200100 Outstanding Cheques)
- Create report > Submit > Approve > Post > Payment > Return Advance
and Amount to Clear is not clear


kittiu:
Replaced by https://github.com/OCA/hr/pull/712

PR #693, opened 2019-10-04
[MIG] hr_employee_id: Migration to 13.0

damdam-s:


MiquelRForgeFlow:
Please, include https://github.com/OCA/hr/commit/467e53037a7e0a713afe354d8a49a08b7d8b8c00#diff-3929a11375079966d8468d502f4a40f7 changes on employee file.

Saran440:
@damdam-s This module already feature in core odoo v.13
Employees > HR Settings > field `Badge ID`

We should be change this module for auto gen number field `Badge ID` or else?
Any suggestion?

alexey-pelykh:
@Saran440 v13 declared the field and yet it's completely manual? A reference to that field in Odoo code would be great (sorry, I'm in v12 deeply still)

Saran440:
@alexey-pelykh Sorry, I don't get your question.

In v12
there is field `budge id` but you need to install `hr_attendance` module
- this field will auto random number to new employee

In v13
there is field `budge id` in `hr` module
- there is `generate` button next to `budge id` for auto random number
- and this field write help = ID Used for employee identification
Selection_003

So, I think this field has the same function as this module that
should we use `Badge ID` instead of create new field?

alexey-pelykh:
@Saran440 to my understanding, Badge ID is something different that identification ID: e.g. in HRIS like BambooHR every employee gets unique ID for internal document references (agreements, payroll, etc.), which is not related to badge ID that's printed if form of QR or other code on access badge.

Saran440:
@alexey-pelykh Okay, At first I thought badge ID can instead of identification ID because it's similar function and unique.
Thank you for your advice. :)

alexey-pelykh:
Badges can be lost and reissued with new ID, while employee ID never changes, even after hire-fire-hire cycle :)

alexey-pelykh (changes requested):
Please do a migration according to https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 (see the precommit section)

astirpe:
I'm continuing the migration here: https://github.com/OCA/hr/pull/797.

PR #694, opened 2019-10-04
[12.0][FIX] hr_holiday_public: Fix method return for concatenation

p-tombez:
Prevent error like `TypeError: can only concatenate list (not "hr.holidays.public.line") to list` when trying to concatenate the result of multiple calls to `get_holidays_list` if any of them is empty


pedrobaeza (approved):
Fast-tracking this obvious one

PR #696, opened 2019-10-08
[12.0][ADD] hr_expense_tier_validation

ps-tubtim:


hr_expense_tier_validation/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,20 @@
...
+    _state_from = ['draft']
+    _state_to = ['submit', 'approve']
+
+    @api.multi
+    def action_submit_sheet(self):
+        if self.need_validation:
+            if self.validated:
+                super().action_submit_sheet()
+            raise ValidationError(_('Please request a validation.'))
+        super().action_submit_sheet()
```
LoisRForgeFlow: I think there is no need for this method, just defining `_state_from` `_state_to` should be enough. Or am I missing something special here?

LoisRForgeFlow (changes requested):

hr_expense_tier_validation/models/hr_expense_sheet.py
ps-tubtim: I need to raise if '**Submit to manager**' before '**Request Validation**'. You can suggest to me.

ps-tubtim: Sorry, I misunderstand. This code is no need. Thank you

LoisRForgeFlow (approved):
Tested on runbot LGTM

kittiu:
@lreficent @ps-tubtim 
I want to discuss with you. Currently in Tier Validation, user will need to click on "Request Validation" to start the approval. Which is good for generic case.
In real project, we may want the Request Validation to be triggered automatically. How you normally that? Or, should this become a new feature in base_tier_validation?

kittiu (approved):

ps-tubtim:
@kittiu 
I think it can be a new feature but it should choose to use or not use.

LoisRForgeFlow:
@kittiu it will depend on how this automatic request should be triggered, what do you have in mind? just after creation? when some condition is met?

kittiu:
@lreficent to ensure flexibility, I think we can do it based on condition, i.e., when state is set or some value is filled. But how this could be configured, I still wonder.
Anyway, I better create this as new issue, we can finish this PR first.

Thanks!

LoisRForgeFlow:
@kittiu yes, I agree better to discuss it in a new issue at server-ux. Regarding this specific PR the only think left is someone who merges it :smile:

kittiu:
Sorry for the out of scope discussion.
Ready to merged.

HviorForgeFlow (approved):
Code review :+1:

PR #697, opened 2019-10-09
[FIX] hr_holiday_public: Delete holidays public

OpenCode:
**Before PR:**

Open a Public Holiday and try to delete it. An error will be raised:

```
The operation cannot be completed:
- Create/update: a mandatory field is not set.
- Delete: another model requires the record being deleted. If possible, archive it instead.

Model: Public Holidays Lines (hr.holidays.public.line), Field: Calendar Year (year_id) 
```

**After PR:**

No error. Public Holidays and all the lines will be removed.

### DIFF
--- a/hr_holidays_public/__manifest__.py
+++ b/hr_holidays_public/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'HR Holidays Public',
-    'version': '12.0.1.0.2',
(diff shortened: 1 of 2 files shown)


jarroyomorales (approved):
True, it was very annoying. Thank you! :+1:

pedrobaeza (approved):
FYI, now we can change automatically module version on merge

PR #698, opened 2019-10-10
[12.0][MIG] hr_holidays_compute_days [WIP]

kittiu:
Standard migration, work in progress


kittiu:
@pedrobaeza Odoo 12 do not have function `to_naive_utc()`
Do you know what can be used instead?

pedrobaeza:
I suppose that standard datetime manipulation: https://stackoverflow.com/a/7065242

kittiu:
@pedrobaeza I want to withdraw myself from migrating this module with following reasons,

- I think there are many changed introduced in v12. I..e., split models, rename models, Half Day, date from/to no longer have time anyway (with different name). To the point it is not a standard migration.
- I myself, not very familiar with hr_holidays nor hr_holidays_compute_days, so might not do things right.

I have push code to the point of attempt to do standard migration and will leave it here. Sorry...may be someone else want to continue?

pedrobaeza:
OK, thanks any way for trying

shabeervpk:
Hi,
Can we achieve the 'Rest day or Weekend Exclude' Option in Leave as the same way we are excluding public holidays ([hr_holidays_public](https://github.com/OCA/hr/tree/12.0/hr_holidays_public))

Am testing the code, changes made:

- Leave Type : exclude_weekends (Boolean Field)

`exclude_weekends = fields.Boolean(
        string='Exclude Weekend',
        default=False,
        help=(
            'If enabled, weekends (weekend is considered from employee working hours) are skipped in leave days'
            ' calculation.'
        )
    )`

- In `hr.leave` model: pass exclude_weekend in the context
def _get_number_of_days(self, date_from, date_to, employee_id):
        instance = self
        if (self.holiday_status_id.exclude_public_holidays or
                not self.holiday_status_id):
            instance = instance.with_context(
                employee_id=employee_id,
                exclude_public_holidays=True,
            )
        if self.holiday_status_id.exclude_weekends:
            instance = instance.with_context(
                exclude_weekends=True
            )

- In `resource.calendar` model: If exclude weekend, then make resource = False

@api.multi
    def _leave_intervals(self, start_dt, end_dt, resource=None, domain=None):
        if self.env.context.get('exclude_weekends'):
            res = super()._leave_intervals(
                start_dt=start_dt,
                end_dt=end_dt,
                resource=False,
                domain=domain,
            )
        else:
            res = super()._leave_intervals(
                start_dt=start_dt,
                end_dt=end_dt,
                resource=resource,
                domain=domain,
            )
        if self.env.context.get('exclude_public_holidays'):
            tz = timezone((resource or self).tz)
            public_holidays = self._public_holidays_leave_intervals(
                start_dt,
                end_dt,
                self.env.context.get('employee_id', False),
                tz
            )
            res = res | public_holidays
        return res

AaronHForgeFlow:
Is this PR superseded anywhere? Thanks in advance.

pedrobaeza:
This one is not needed anymore, as Odoo does it in core.

AaronHForgeFlow:
Yes, kind of, the most thing you can do if taking half a day. I guess that is enough. Thank you!

PR #701, opened 2019-10-16
[12.0][MIG] hr_employee_medical_examination

jarroyomorales:
Standard migration. no changes.


Saran440:
@jarroyomorales Functional Test.
- Field year should be not show thousands separator.
- Tree view in menu View Medical Examinations should be show employee field for case create Medical Examinations from employee.

jarroyomorales:
@Saran440 Attended you second comment. Regarding the first one, should I convert the field to char? The format is because of being an integer field, I dont know if there is an easy way to remove the formatting. Maybe its not that bad to keep it like this.

Saran440:
@jarroyomorales IMO, if you used char field. It will look good and convert to integer when you want to calculate something in the future but I don't mind :)

Saran440 (approved):
Functional Test, LGTM :+1:

jarroyomorales:
Okay, I put it as char @Saran440

etobella (approved):
LGTM

PR #709, opened 2019-10-22
[12.0][MIG] hr_calendar_rest_time

jarroyomorales:
Migration to 12.0 of `hr_calendar_rest_time`and it's hook module `resource_hook`


jarroyomorales:
@etobella @alexey-pelykh @JordiBForgeFlow @Saran440 Could you review please? Thank you!

hr_calendar_rest_time/readme/HISTORY.rst
```diff
@@ -0,0 +1,5 @@
+11.0.1.0.0 (2019-05-28)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* Define rest time in the resource calendar to adjust the number of worked
+  hours during the day taking this time into consideration.
```
Saran440, line 5: please, add history version 12.0.1.0.0

resource_hook/readme/HISTORY.rst
```diff
@@ -0,0 +1,7 @@
+11.0.1.0.0 (2019-05-28)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* Introduce hook in method '_get_work_days_data' of ResourceMixing class in
+  order to allow changes in the logic to compute working hours within an
+  interval. A new method '_get_work_hours' is now provided for this purpose.
+  The method can be inherit to alter the standard behaviour.
```
Saran440, line 7: please, add history version 12.0.1.0.0

Saran440 (commented):
@jarroyomorales Sorry for delay, In core odoo, you can create day period morning and afternoon in working times.
I think result is not difference.
Selection_001

but I tested this module
Selection_002
- `Average hour per day` should be 8.00 hours. exclude `Rest Time`
- `Day Period` should be add `All Day`

and code reviewed

jarroyomorales:
@Saran440 Thanks for the review. Yes, the total hours per day will be the same
The difference is that in core Odoo as you say, you define 2 lines, one for the morning, form 8-12 for example and another one for the afternoon, from 13 to 17. In this example you are indirectly saying that the employee has dinner from 12 to 13 but in a company which dinner time is flexible, it is very hard to manage it like this.
The most realistic way is to say that the employee works from 8 to 17 but takes 1 hour to have dinner, which is what this module is for.

Saran440:
Okay, I see.
I will approve when you fixed.

jarroyomorales:
@Saran440 Well, it is the version and the date when it was done, why would I change it to 12.0 and today's date?

Saran440:
@jarroyomorales Sorry, I mean add history v.12 in newline but not sure that necessary.
https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/HISTORY.rst

jarroyomorales:
@Saran440 Oh sorry, I misunderstood you. There haven't been major changes so I think we can keep it like this.

Saran440:
@jarroyomorales Okay. by the way, I'm not sure that you see this sentence.
Selection_001

jarroyomorales:
@Saran440 Comments attended! Added a new `day_period` and rewrote `_onchange_hours_per_day` to compute the average hours properly.

hr_calendar_rest_time/models/resource_calendar.py
```diff
@@ -0,0 +1,26 @@
...
+
+class ResourceCalendar(models.Model):
+
+    _inherit = 'resource.calendar'
+
+    @api.onchange('attendance_ids')
+    def _onchange_hours_per_day(self):
+        if self.env.context.get('use_old_onchange_hours_per_day'):
+            return super()._onchange_hours_per_day()
+        attendances = self.attendance_ids.filtered(
```
Saran440, line 16: Why are you filtered attendance_ids not `date_from` and `date_to`?
I tested create `Starting Date` and `End Date`. 
It not used `Rest Time` to calculate `Average hour per day`

jarroyomorales: Yes, I know this is an issue but it is the same way odoo does it in the core `resource` module.
At first I implemented the right solution taking into account the start and end date but it is not that easy. You would need a cron to check every day which lines are active are need to be used for the calculation and then update the value, since it is computed in an onchange.
Maybe we can add this behavior but it should probably be in another module.

Saran440 (approved):
Functional Test. LGTM :+1:

etobella (approved):
LGTM :+1:

etobella:
@OCA/human-resources-maintainers Can this be merged?

PR #660, opened 2019-08-22
[12.0] [MIG] hr_holidays_imposed_days

emagdalenaC2i:

PR #713, opened 2019-10-24
[12.0][MIG] hr_holidays_imposed_days

WhiteTiger08:
Fixing issues with 12.0:
- Changing hr.holidays to hr.leaves
- Changing hr.status.id to hr.leave.type


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0

pedrobaeza:
And there's already other migration in #660. Please check that one instead.

PR #717, opened 2019-11-05
[FIX] hr_employee_calendar_planning: link leaves

chienandalu:
- On the module init, existing employee calendar leaves should be linked
to the employee's new autocalendar.

cc @Tecnativa TT20229

### DIFF
--- a/hr_employee_calendar_planning/__init__.py
+++ b/hr_employee_calendar_planning/__init__.py
@@ -2,4 +2,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from . import models
+from . import wizards
 from .hooks import post_init_hook
--- a/hr_employee_calendar_planning/__manifest__.py
+++ b/hr_employee_calendar_planning/__manifest__.py
@@ -16,6 +16,7 @@
     "data": [
         "security/ir.model.access.csv",
         "views/hr_employee_views.xml",
+        "wizards/wizard_employee_calendar_views.xml",
     ],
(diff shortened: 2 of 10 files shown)


pedrobaeza (approved):

pedrobaeza:
I didn't add the smart-button on purpose: that must be handled only in backend, not by user.

chienandalu:
> I didn't add the smart-button on purpose: that must be handled only in backend, not by user.

I encountered some situations in wich the user would need to get rid of old calendars that could cause troubles, like the ones with lots of date_ranges on them, wich lead to multiple attendance_ids for the days issued.

Another use I think of is to leave on the user behalf to clean up the automatic calendars. When lots of them are present it can be quite a mess (when maybe just a handful of them could be really needed).

pedrobaeza:
There's only one automatic calendar per employee. I insist this shouldn't be available at user level.

chienandalu:
There's one atomatic calendar per user, but many calendars could be created in the module init to fit different date ranges or simply exist before and no be usable anymore.

pedrobaeza:
> but many calendars could be created in the module init to fit different date ranges or simply exist before and no be usable anymore.

Yeah, but they are active=True, so no need to show the "Deactivate" button: remove them or adjust them.

chienandalu:
Ok, I dropped the commit

pedrobaeza (approved):

etobella (approved):

pedrobaeza:
/ocabot merge minor

To be migrated to 11.0/12.0

PR #718, opened 2019-11-05
[IMP] hr_attendance_report_theoretical_time: recompute attendances

chienandalu:
- We add a wizard so the already existing attendances can be recomputed
according to an employee calendar change, like a different for a set of
dates that weren't took into consideration in advance, so the wrong
theoretical hours are computed.

cc @Tecnativa TT20229

### DIFF
--- a/hr_attendance_report_theoretical_time/wizards/recompute_theoretical_attendance.py
+++ b/hr_attendance_report_theoretical_time/wizards/recompute_theoretical_attendance.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class RecomputeTheoreticalAttendance(models.TransientModel):
+    _name = 'recompute.theoretical.attendance'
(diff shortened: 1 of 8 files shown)


hr_attendance_report_theoretical_time/wizards/recompute_theoretical_attendance.py
```diff
@@ -0,0 +1,37 @@
...
+    @api.multi
+    def action_recompute(self):
+        self.ensure_one()
+        attendances = self.env['hr.attendance'].search([
+            ('employee_id', 'in', self.employee_ids.ids),
+            ('check_in', '>=', self.date_from),
+            ('check_out', '<=', self.date_to),
+        ])
+        attendances._compute_theoretical_hours()
+        return {'type': 'ir.actions.act_window_close'}
```
pedrobaeza, line 37: I think there's no need of the action. Simply `return`

pedrobaeza (approved):

jarroyomorales (approved):
Only code review, but LGTM! I think it is a necessary feature for this module.

pedrobaeza:
Who wants to migrate it to 11.0/12.0? :pray:

jarroyomorales:
@pedrobaeza I can do it on monday

pedrobaeza:
Great, thanks!

PR #719, opened 2019-11-05
Added Theoretical time difference to employee form. Fixed inconsistency

wpichler:


pedrobaeza:
Please split both changes in correspondent PRs if you want the fix arrives sooner than the rest. For the main change, please explain rationale.

wpichler:
About the inconsistency fix: The Theoretical time report menu is titled "All Employees" - but if you open it - then you get it filtered for your current user - and not for "All Employees". This fix does change the default filter to be for "All Employees"

wpichler:
About the new theoretical difference field: Users did expected to see the overtime (= theoretical difference) for the employees directly on the employee form. For this i did added the computed (and not stored) field theoretical_hours_difference to hr.employee and did added a smartbutton to the form view. So the overtime is visible and when you click it - you will get the pivot table with the employee's theoretical time analysis.

pedrobaeza:
Thanks for the explanations. I proposed the 2 PRs for allowing the fix to be easily merged while the new feature is discussed. Any way, this is all red. Can you check?

wpichler:
As requested - created #720 and #721 - will close this one

PR #720, opened 2019-11-06
[FIX] Fixed inconsistency - Menu action is called report for all employees - default was only for the current user.

wpichler:
The Theoretical time report menu is titled "All Employees" - but if you open it - then you get it filtered for your current user - and not for "All Employees". This fix does change the default filter to be for "All Employees"

Part of #719

### DIFF
--- a/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
+++ b/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
@@ -45,7 +45,7 @@
         <field name="name">Theoretical vs Attended Time Analysis</field>
         <field name="res_model">hr.attendance.theoretical.time.report</field>
         <field name="view_type">form</field>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
This was introduced as something "covered" in https://github.com/OCA/hr/commit/3ec402526d617b14321a1e5c83ef0cfb8911d99c#diff-b7196d0e6241925a57cc8af2ceb7e4ccR68.

In its moment, the filter `my` was activated for performance reasons, so first view is quick, but you always can remove it for seeing all (if you have permissions). The logic is similar to "Sales Orders", where you can see all, but by default they are filtered by only yours, with a removable filter.

@jarroyomorales you were that did the change, what do you think?

jarroyomorales:
@pedrobaeza @wpichler In my opinion it is not an inconsistency to have a menu that shows you information about all employees but by default it only shows yours. That menu is limited to people with at least Officer permission, so people that can only see their own attendances cant access the menu (that would be an inconsistency with the name).
I put that name to difference between the menu that lets you filter employees previously and the menu that shows you all the employees..

PR #721, opened 2019-11-06
[ADD] Added computed theoretical difference field on employee with smart button to report

wpichler:
Users did expected to see the overtime (= theoretical difference) for the employees directly on the employee form. For this i did added the computed (and not stored) field theoretical_hours_difference to hr.employee and did added a smartbutton to the form view. So the overtime is visible and when you click it - you will get the pivot table with the employee's theoretical time analysis.


hr_attendance_report_theoretical_time/models/hr_employee.py
```diff
@@ -1,15 +1,25 @@
...
-from odoo import fields, models
+from odoo import fields, models, api
 
 
 class HrEmployee(models.Model):
     _inherit = 'hr.employee'
 
+    @api.multi
+    def _compute_theoretical_hours_difference(self):
+        data = self.env['hr.attendance.theoretical.time.report'].read_group([('employee_id', 'in', self.ids)],
```
pedrobaeza: This doesn't conform flake8 80 col limit

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
```diff
@@ -49,6 +49,37 @@
         <field name="view_mode">pivot,graph</field>
     </record>
 
+    <record id="hr_attendance_theoretical_view_employee_pivot" model="ir.ui.view">
```
pedrobaeza, line 52: Why do you need an extra view? This one is the same as `hr_attendance_theoretical_view_pivot`

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
```diff
@@ -49,6 +49,37 @@
...
+                <field name="date" type="row"/>
+                <field name="employee_id" type="col"/>
+                <field name="worked_hours" type="measure" widget="float_time"/>
+                <field name="theoretical_hours" type="measure" widget="float_time"/>
+                <field name="difference" type="measure" widget="float_time"/>
+            </pivot>
+        </field>
+    </record>
+
+    <record id="hr_attendance_theoretical_employee_action" model="ir.actions.act_window">
```
pedrobaeza, line 66: The same for the action. If you need something different (like the context), use a method in the model reading the act_window and modifying it accordingly

pedrobaeza (changes requested):

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
wpichler: Its not the same view - the new view does have the employee_id in the col - and the date in the row. So you get a nice single employee view. This is not possible by using the same view.

pedrobaeza: Yeah, and I don't think that is a good change... I prefer consistency in the way of showing data always in the same layout.

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
wpichler: I've used a new action to be able to set my special view for the action - also i did have to remove the default context.
It would be possible to do this in a model function, but because it does also modify the used view - i think it is better to create a new action

pedrobaeza: Yeah, but following my previous comment logic, I still don't think it's needed.

hr_attendance_report_theoretical_time/models/hr_employee.py
wpichler: fixed

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
wpichler: The current layout is perfect when you look at more than 1 employee. But as soon as you have only 1 employee - and you have the complete history enabled (instead of the last two months) - then the horizontal view will become a pain for users. The vertical layout is much better suited for this - and is also what user do expect here. imho

pedrobaeza: OK, let's see what others think about this, or if we maybe switch both views with this logic

jarroyomorales:
@pedrobaeza @wpichler We should discuss if employees should see their analysis, since when this module was created it was decided that only HR officers could, IMHO it is okay if everyone can see their own report.
Also, I dont understand either why we need a second view, I would just reuse the one we already have.

hr_attendance_report_theoretical_time/views/hr_employee_views.xml
```diff
@@ -7,6 +7,15 @@
             <field name="manual_attendance" position="after">
                 <field name="theoretical_hours_start_date"/>
             </field>
+            <button name="toggle_active" position="after">
+                <button name="%(hr_attendance_theoretical_employee_action)d"
+                        context="{'search_default_employee_id': active_id}"
+                        class="oe_stat_button"
+                        icon="fa-calendar"
```
jarroyomorales: If you could choose another icon it would be great since it is the same as leaves' smart button.

wpichler: sure - which one ?

pedrobaeza:
@jarroyomorales thanks for the insights. Are you sure employees didn't see the report? At least it was that way in v10, that it was originally where I developed the module.

wpichler:
I also was missing the functionality for an employee to see his own attendance records - with this small addons also this problem is solved...

pedrobaeza:
What I did in v10 was to tweak menu permissions for being accessible to employees (in fact it was a new parent menu).

wpichler:
It seems like your changes got lost - in v10 the following part is in the report xml

`
    <record id="hr_attendance.menu_hr_attendance_report" model="ir.ui.menu">
        <field name="action"/>
        <field name="groups_id" eval="[(4, ref('hr_attendance.group_hr_attendance_user'))]"/>
    </record>
`

this is missing in branch 12.

Should i add it to this PR ?

hr_attendance_report_theoretical_time/views/hr_employee_views.xml
jarroyomorales: 'clock-o' for example?

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
meigallodixital: I change axis when get 1-3 Employees data. But its true that with a lot of them the original layout is better.

pedrobaeza: Having the direct button for flipping axis makes no pain to have the same layout for both for avoiding extra XML data

pedrobaeza:
> this is missing in branch 12. Should i add it to this PR ?

Well, if we add this, it's less needed, but now I have another doubt: as employee directory is public, any employee can see through this smart-button the report for any employee?

wpichler:
> > this is missing in branch 12. Should i add it to this PR ?
> 
> Well, if we add this, it's less needed, but now I have another doubt: as employee directory is public, any employee can see through this smart-button the report for any employee?

I can see a security rule which should prevent this - but this security rule would also affect the compute function - and will throw an exception at it. I will test this now

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
wpichler: A direct button for flipping axis ? How to get this button ?

pedrobaeza:
I think you shouldn't show the smart-button at all if you don't have permission for that, but not sure about the implementation: maybe a computed field that checks this...

hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
pedrobaeza: Peek 06-11-2019 16-25

wpichler: working now around 6 years with odoo - and have never used that button until now ;-)

PR #722, opened 2019-11-11
[11.0][IMP] hr_attendance_report_theoretical_time: recompute attendances

jarroyomorales:
Same as https://github.com/OCA/hr/pull/718 but for 11.0

PR #724, opened 2019-11-11
[IMP] hr_attendance_report_theoretical_time: recompute attendances

chienandalu:
- We add a wizard so the already existing attendances can be recomputed
according to an employee calendar change, like a different for a set of
dates that weren't took into consideration in advance, so the wrong
theoretical hours are computed.

FW port from #722 and #718

cc @Tecnativa TT20229


jarroyomorales (approved):

pedrobaeza (approved):
Thanks for the quick reaction

PR #729, opened 2019-11-22
[12.0][ADD] hr_attendance_geolocation

AdriaGForgeFlow:
With this module the geolocation of the user is tracked at the check-in/check-out step

cc ~ @Eficent

PR #730, opened 2019-11-22
[12.0][ADD] hr_attendance_permitted_geolocation

AdriaGForgeFlow:
With this module the user can set a maximum range within which the check-in/check-out process must be done. Registering attendance outside this range will trigger an error.

- [ ] Depends on #729 

cc ~ @Eficent


alexey-pelykh:
I assume it's WIP?

PR #736, opened 2019-11-28
[12.0][FIX] hr_attendance_autoclose: compute needs api.depends

MiquelRForgeFlow:
Fast-track.

### DIFF
--- a/hr_attendance_autoclose/models/hr_attendance.py
+++ b/hr_attendance_autoclose/models/hr_attendance.py
@@ -10,6 +10,7 @@ class HrAttendance(models.Model):
     _inherit = "hr.attendance"
 
     @api.multi
+    @api.depends('check_out', 'check_in')
     def _compute_open_worked_hours(self):
         for attendance in self:
             if attendance.check_out:


hr_attendance_autoclose/models/hr_attendance.py
```diff
@@ -10,6 +10,7 @@ class HrAttendance(models.Model):
     _inherit = "hr.attendance"
 
     @api.multi
+    @api.depends('check_out')
```
alexey-pelykh: and `check_in`?

MiquelRForgeFlow: No, it doesn't depend of `check_in`.

alexey-pelykh: `attendance.check_out - attendance.check_in`?

alexey-pelykh: @mreficent ^

AaronHForgeFlow:
Does it need api.depends even if the field is not stored?

jarroyomorales:
If you make it depend on `check_out` and the user doesn't check out, `open_worked_hours` will never be updated and the cron wont close it.

AaronHForgeFlow:
I mean, if it is not stored, it is recomputed any time a user reads the information, isn't it?

hr_attendance_autoclose/models/hr_attendance.py
etobella: I think `check_in` is necessary

alexey-pelykh:
let's follow https://www.odoo.com/documentation/12.0/reference/orm.html#computed-fields

alexey-pelykh:
We should list literally all used fields in `dependson`

AaronHForgeFlow:
Tested on runbot. Updating `check_in` field triggered the computation of the field. In any case, if we add the `api.depends` we should add also `check_in` in the `api.depends`.

hr_attendance_autoclose/models/hr_attendance.py
MiquelRForgeFlow: all of you are right, the check_in is necessary.

MiquelRForgeFlow:
hey people, `check_in` added. please update review and merge :)

alexey-pelykh:
but have you pushed? :)

alexey-pelykh (approved):

alexey-pelykh:
fast-tracking as trivial

MiquelRForgeFlow:
> but have you pushed? :)

there was some lag haha

etobella (approved):

AaronHForgeFlow (approved):
:+1:

PR #739, opened 2019-12-01
[12.0][IMP] hr_expense_cancel : Cancel vendor bills from hr_expense

Saran440:
Step to error:
1. create expense normal process
2. Submit to Manager > Approve > Create Vendor Bill
3. Post Journal Entries and click cancel
4. when you click cancel it will showed message error.
`You cannot do this modification on a reconciled entry. You can just change some non legal fields or you must unreconcile first.`

IMP : Collaborative support module hr_expense_cancel and hr_expense_invoice.
you can cancel when create invoice from hr_expense.

### DIFF
--- a/hr_expense_cancel/README.rst
+++ b/hr_expense_cancel/README.rst
@@ -88,6 +88,7 @@ Contributors
   * Ernesto Tejeda
 
 * Danh Vo  <https://github.com/danhvophuong>
+* Saran Lim. <saranl@ecosoft.co.th>
 
 Maintainers
 ~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


Saran440:
@ps-tubtim @danhvophuong @jarroyomorales Could you please help review.

ps-tubtim (approved):
Code and Functional review: OK.

AaronHForgeFlow (commented):
Does this work if hr_expense_invoice is not installed?

Saran440:
@AaronHForgeFlow yes, it's work. In account move line will not has invoice_id if you not install hr_expense_invoice

kittiu:
@AaronHForgeFlow your request is addressed?

newtratip (approved):
:+1: LGTM

pedrobaeza (changes requested):
I don't see correct that cancelling an expense, should cancel possible linked invoices, as they can be different documents, and although linked, it belongs to different persons to process them.

PR #741, opened 2019-12-05
[12.0][FIX] hr_attendance_report_theoretical_time: menu permissions

chienandalu:
- Although the minor permission to access the theoretical time report is
"Manual Attendance", the menu wasn't accessible due to permissions.
- Record rules also prevent such users from seeing other employees
attendances.

cc @Tecnativa TT20784


rafaelbn (approved):
:+1:

pedrobaeza (approved):

PR #742, opened 2019-12-06
[13.0][MIG] hr_employee_firstname

astirpe:


ps-tubtim (approved):
Code LGTM

lfreeke (approved):
Functional test :+1:

astirpe:
Is it ok to merge this?

astirpe:
@alexey-pelykh would you merge this?

hr_employee_firstname/tests/test_hr_employee_firstname.py
```diff
@@ -10,24 +10,23 @@
 class TestEmployeeFirstname(TransactionCase):
     def setUp(self):
         super().setUp()
-        self.employee_model = self.env['hr.employee']
```
alexey-pelykh: Why this is needed?

oca_dependencies.txt
```diff
@@ -0,0 +1 @@
+partner-contact
```
alexey-pelykh, line 1: Why that's needed?

alexey-pelykh (approved):
Overall good, but I have some questions

oca_dependencies.txt
astirpe: Module depends on `partner_firstname`

alexey-pelykh: Totally makes sense, sorry, missed that :D

hr_employee_firstname/tests/test_hr_employee_firstname.py
astirpe: Why not? I think is a matter of taste...

alexey-pelykh: More copypaste of a model name string? + a lookup every time?

astirpe: You like the old style for tests.. I see.
Do you prefer I revert such changes?

alexey-pelykh: If to take about personal preferences, I do prefer to reduce copy-paste, yet my personal preference is not a blocking one :) In this particular case I'll just need a second opinion on this, that's all - because truly I'm not sure how it's better for OCA

alexey-pelykh: I mean, on one hand, I also did tidy-up things in migration, yet on other hand it's better to keep migration as small as possible 🤷‍♂

astirpe: Reverting the change will speed up the approval process? This PR is pending for a long time, I don't want to wait more.

alexey-pelykh: I've asked @pedrobaeza to share his opinion on this, probably will take couple of hours? Yet answering to your question specifically: yes, but I don't want to push for that change

alexey-pelykh: If to talk about style preferences, I'd wrote that as `self.HrEmployee` :)))) Yet that is completely irrelevant

pedrobaeza: To cache the model references is a way to not query each time the environment, and to have shorter names for invocating them, so I see this as something correct, but author preference, so I wouldn't argue if I find it or not.

alexey-pelykh: @astirpe so if you think that is reasonable, please revert that change, otherwise I'll just merge it as is.

astirpe: Is it okay to merge as it is? Thanks.

PR #752, opened 2019-12-20
[12.0][REF][WIP] hr_holidays_public

jarroyomorales:
The purpose of this PR is to break the ice regarding the conversation in #646. 
This is a first approach on how the module would be refactored so it manages a new type of leave `by region`, the same way Odoo can manage leaves by category, by department or by company.
My idea is that the definition of the public holidays calendar remains the same, but a new button `Generate Leaves` is added, which will automatically create the leave type and all the `hr.leaves` related to that.

Known Issues:
- [ ] Public holidays should be able to overlap with other kind of leaves
- [ ] Can't modify the calendar once leaves are generated

### DIFF
--- a/hr_holidays_public/__manifest__.py
+++ b/hr_holidays_public/__manifest__.py
@@ -18,6 +18,7 @@
     'data': [
         'security/ir.model.access.csv',
         'views/hr_holidays_public_view.xml',
+        'views/hr_leave.xml',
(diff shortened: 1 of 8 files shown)


jarroyomorales:
ping @pedrobaeza @alexey-pelykh  @ivantodorovich @etobella

PR #761, opened 2020-01-03
[ADD]maximum_hours_per_day in resource.calendar model

chusamo:
This Pull request adds the maxium hours per day in the resource.calendar model instead of the res.company in order to have more flexibility with diferents workdays.


jarroyomorales (commented):
Yes, it makes more sense. LGTM :+1:
The only thing I'm not sure is how this will work with the `hr_employee_calendar_planning` module, which `max_time` will the generated calendar get?

chusamo:
Close by time

Issue #755, opened 2019-12-21
Validation Error in hr_public_holidays module v12


SSanfilippo:
Hi at all,
I think that the module had same issue that was fixed in v11: https://github.com/OCA/hr/issues/569

Odoo Version - v12 Community

1. Create a new Public Holiday.
2. Add a new Public holiday line.
3. Select the date, add name and untick the Date may change Boolean.

On save get error on Date not filled.

jarroyomorales:
At least for me it is not clear what's the purpose of `variable_date` . I thought it was used to indicate if that holiday is on the same day every year or not. (Christmas is always the 25th of December but Easter changes every year for example, at least in Spain).
If the only purpose of that variable is to set the `date` field as readonly I think we can just get rid of it....

SSanfilippo:
Yes, I confirm that I set a fixed holiday or variable holiday I get always same error.

In Odoo 11 was fixed with a few line of code, I try to reapply the fix (the files are as previous of this fix), but the error remain.

jarroyomorales:
The function in https://github.com/OCA/hr/blob/11.0/hr_holidays_public/models/hr_holidays_public.py#L196 that was made to solve the error sets the date to False and raises a warning. It sets date to False since it has lost the reference to its previous value, which doesnt make any sense, since at the end you are also changing the date and not solving anything.
We should first figure out the meaning of `variable_date` and then do something according to that.

@feketemihai @mtelahun @alexis-via @saltonmassally @yelizariev @bassn @pedrobaeza @alexey-pelykh, you are contributors/authors of this module, any answer to this?

kensshiro:
The problem is caused by an attribute in the file hr_holidays_public_view.xml. Set Date as readonly and in the create or write, the field does not take it by having this attribute. 
This can be fixed by adding the field to the values ​​variable for saving. 
The use of the variable_date field that I would give would be the following:
Add to the wizard the creation of a new Plan, the detection of those records with the field activated for modification with the new dates before the creation of the new Plan. So we can solve Easter, Easter Week, etc. in Spain, since they vary every year. And the variable_date field by default false.

jarroyomorales:
@kensshiro I completely agree with you. In fact I already had an implementation of what you say in the last paragraph, I'm going to create a PR and see what people thinks

kensshiro:
Thanks!! It's good idea!!

PR #763, opened 2020-01-08
[12.0][IMP] hr_holidays_public: improve wizard

jarroyomorales:
Refactor of the public holidays generator wizard to take into account that public holidays are not always the same day every year. It follows the discussion in #755 

Before this PR the wizard took the latest template of every country and generated the new calendars. It was very nice because you could generate all the different countries calendars with just one click, but it wasn't useful at all if after this you had to go one by one correcting the dates that changed (Easter is the most common example). Also the field `variable_date` wasn't serving any purpose.

Now, when opening the wizard you have to select a template and which year you want to generate. It will display all the lines that were marked as `variable_date` where you will have the chance of changing the proposed date in case it is necessary. After doing this, a new calendar will be created for the country of the template with the lines where `variable_date` is false with the updated year, and with the lines you just solved.
It is true that you now have to create calendars one by one, but at least you create them correctly, which in my opinion is better that creating all at the same time and correcting them one by one later...

imagen

What do you think? @kensshiro @SSanfilippo @etobella @alexey-pelykh @pedrobaeza

### DIFF
--- a/hr_holidays_public/wizards/holidays_public_next_year_wizard.xml
+++ b/hr_holidays_public/wizards/holidays_public_next_year_wizard.xml
@@ -9,44 +9,49 @@
         <field name="model">public.holidays.next.year.wizard</field>
         <field name="arch" type="xml">
             <form string="Create Next Year Public Holidays">
+                <field name="warning_existing" invisible="1"/>
+                <div class="alert alert-warning"
+                     aria-label="Warning"
+                     title="Warning"
+                     role="alert"
+                     attrs="{'invisible': [('warning_existing', '=', False)]}"
+                     style="margin-bottom:0px;">
+                    <p>
+                        <i class="fa fa-info-circle"/>
+                        <span>Warning: A public holidays calendar for this country and year already exists!</span>
+                    </p>
+                </div>
                 <sheet>
                     <div>
-                            Use this wizard to create public holidays based on the
-                            existing ones.<br/>
-                            Only the last templates of each country
-                            will be taken into account (If you select templates
-                            from 2012 and 2015 of the same country; '
-                            only the templates from 2015 will be taken into
-                            account).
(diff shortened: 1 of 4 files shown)


hr_holidays_public/wizards/holidays_public_next_year_wizard.xml
```diff
@@ -9,44 +9,48 @@
...
-                            anniversary of the Italian unification, so you want to
-                            replicate the 2010 Italian holidays to 2012."
-                            </div>
-                            <group>
-                                <field name="template_ids" />
-                                <field name="year" />
-                            </group>
+                        <page name="pending" string="Days to Review">
+                            <field name="pending_lines">
+                                <tree editable="bottom">
```
etobella: Add `create="0"`

hr_holidays_public/wizards/holidays_public_next_year_wizard.xml
```diff
@@ -9,44 +9,48 @@
...
+                                    <field name="date"/>
+                                    <field name="name"/>
+                                    <field name="state_ids" widget="many2many_tags"/>
```
etobella, line 49: `line_id` is missing and gives an error: You shoud add `<field name="line_id" invisible="1"/>`

etobella (changes requested):
LGTM but two changes are needed

jarroyomorales:
@etobella Comments attended!

ovnicraft:
@etobella please review your request.

etobella (approved):
LGTM

jarroyomorales:
@ovnicraft Can you review please?

ovnicraft (approved):

pedrobaeza:
To be forward-ported to OCA/hr-holidays@13.0

PR #770, opened 2020-01-16
[8.0][MIG] hr_attendance_geolocation Backport 12.0 to 8.0

ps-tubtim:
Backport of hr_attendance_geolocation 12.0 to 8.0


ps-tubtim:
cc: @kittiu

kittiu:
@ps-tubtim please make Travis green. Many errors still.

hr_attendance_geolocation/models/hr_attendance.py
```diff
@@ -0,0 +1,41 @@
...
+
+UNIT = dp.get_precision("Location")
+
+
+class HrAttendance(models.Model):
+    _inherit = "hr.attendance"
+
+    latitude = fields.Float(
+        string="Latitude",
+        digits=UNIT,
```
kittiu: This digit is not working after fresh install. It is still record in 2 decimal.
I think latitude is fixed decimal (i.e., 7?). Why don't use `digits=(16, 7)` ? So it can works on first install.
note: I don't see why we need to store this in database also.

kittiu:
**Testing with desktop chrome browser**

I tested it in bangkok, but the result that I check with google map is in **Petchaboon**
image

You don't have this problem?

**Try to test with chrome on Andriod, using IP address**

* Not working, but it is because Location will not work with normal IP address.
* May be we need to test by deploy this on customer development server that has URL???

kittiu:
There is no map, but it is ok. Instead, please find the solution that can do the following,

* Get location address, i.e., 50/2 Soi..., Tumbon..., Amphur

image


* A link that can open Google Map from latitude, longitude

hr_attendance_geolocation/models/hr_attendance.py
ps-tubtim: Thank you for suggestion. Edited.

PR #774, opened 2020-01-20
[13.0][MIG] hr_employee_social_media

Saran440:
- field `Social Media Accounts` will show inside `Work Information` tab


jarroyomorales:
@Saran440 Wouldnt `Private Information`'s tab be a better place for this?

hr_employee_social_media/views/hr_employee_view.xml
```diff
@@ -5,7 +5,7 @@
         <field name="model">hr.employee</field>
         <field name="inherit_id" ref="hr.view_employee_form" />
         <field name="arch" type="xml">
-            <xpath expr="//page[@name='public']/group" position="inside">
+            <xpath expr="//div[@id='o_work_employee_main']" position="inside">
```
alexey-pelykh, line 8: The previous place is no longer there?

alexey-pelykh:
@jarroyomorales that would be considered an improvement and anyhow should be done in a separate PR

hr_employee_social_media/views/hr_employee_view.xml
Saran440: The previous place is change group to div.
I don't know which is better pattern between
`<xpath expr="//page[@name='public']/div/div" position="inside">`
and
`<xpath expr="//div[@id='o_work_employee_main']" position="inside">`

Saran440:
@jarroyomorales That's a good idea. I will PR after this pr merge :)

jarroyomorales (approved):

PR #778, opened 2020-01-22
[12.0][IMP] hr_holidays_leave_auto_approve: disable creation of activity

### LINKED ISSUE #777: [12.0] hr_holidays_leave_auto_approve activity created
In previous version the leave was created with context: 'mail_activity_automation_skip' is True. In the last version this was removed resulting in unwanted e-mails to HR managers which get an activity assigned.
@astirpe 

I think it should be:
```
    @api.model
    def create(self, values):
        auto_approve = self._get_auto_approve_on_creation(values)
        tracking_disable = self.env.context.get('tracking_disable', False)
        tracking_disable = tracking_disable or auto_approve
        **mail_activity_automation_skip = self.env.context.get('mail_activity_automation_skip', False)
        mail_activity_automation_skip = mail_activity_automation_skip or auto_approve**
        ctx = self.env.context.copy()
        ctx.update({
            'tracking_disable': tracking_disable,
            'default_company_id': self.env.user.company_id.id,
            '**mail_activity_automation_skip': mail_activity_automation_skip,**
        })
        res = super(
            HrLeave, self.with_context(ctx)
            ).create(values)
        res._apply_auto_approve_policy()
        return res
```

astirpe:
Fixes https://github.com/OCA/hr/issues/777
@CasVissers

### DIFF
--- a/hr_holidays_leave_auto_approve/models/hr_leave.py
+++ b/hr_holidays_leave_auto_approve/models/hr_leave.py
@@ -27,12 +27,14 @@ class HrLeave(models.Model):
     @api.model
     def create(self, values):
         auto_approve = self._get_auto_approve_on_creation(values)
-        tracking_disable = self.env.context.get('tracking_disable', False)
-        tracking_disable = tracking_disable or auto_approve
-        res = super(
-            HrLeave, self.with_context(
-                tracking_disable=tracking_disable)
-            ).create(values)
+        tracking_disable = self.env.context.get('tracking_disable')
+        mail_skip = self.env.context.get('mail_activity_automation_skip')
+        ctx = self.env.context.copy()
+        ctx.update({
+            'tracking_disable': tracking_disable or auto_approve,
+            'mail_activity_automation_skip': mail_skip or auto_approve,
+        })
+        res = super(HrLeave, self.with_context(ctx)).create(values)
         res._apply_auto_approve_policy()
         return res
 


CasVissers-360ERP:
@astirpe 
    'mail_activity_automation_skip': mail_skip or auto_approve,
TypeError: copy() takes no arguments (1 given)

CasVissers-360ERP (approved):
LGTM, fixes the issue! Thnx!

Saran440 (approved):
Code Reviewed. LGTM :+1:

astirpe:
Can this one be merged?

hr_holidays_leave_auto_approve/models/hr_leave.py
```diff
@@ -27,12 +27,14 @@ def _apply_auto_approve_policy(self):
     @api.model
     def create(self, values):
         auto_approve = self._get_auto_approve_on_creation(values)
-        tracking_disable = self.env.context.get('tracking_disable', False)
-        tracking_disable = tracking_disable or auto_approve
-        res = super(
-            HrLeave, self.with_context(
-                tracking_disable=tracking_disable)
-            ).create(values)
+        tracking_disable = self.env.context.get('tracking_disable')
```
tafaRU, line 30: default value is no longer needed?

hr_holidays_leave_auto_approve/models/hr_leave.py
```diff
@@ -27,12 +27,14 @@ def _apply_auto_approve_policy(self):
...
+        mail_skip = self.env.context.get('mail_activity_automation_skip')
```
tafaRU, line 31: similar here: default value is no needed?

tafaRU (changes requested):

MiquelRForgeFlow (approved):

hr_holidays_leave_auto_approve/models/hr_leave.py
astirpe: Default value is False, that's why it's omitted here

hr_holidays_leave_auto_approve/models/hr_leave.py
astirpe: Same as above

hr_holidays_leave_auto_approve/models/hr_leave.py
tafaRU: It seems to be that it is `None` (https://docs.python.org/3/library/stdtypes.html#dict.get)

astirpe: Ok default is `None`.
But the way standard Odoo works is without without specifying the default. Have a look in the entire standard code regarding how `'tracking_disable'` and `'mail_activity_automation_skip'` are checked.

For example you will find:


```python
        if self._context.get('tracking_disable'):
            return bla bla ..
```

or 

```python
        if self.env.context.get('mail_activity_automation_skip'):
            return bla bla ..
```

I'm going to keep this PR as it is.

tafaRU: Ok, thanks for the explanation!

tafaRU (approved):

astirpe:
Thanks! :+1:

PR #780, opened 2020-01-24
[11.0][IMP] hr_autoclose: max_hours_hook

jarroyomorales:
This module first got this information from employee's companies but there is a PR proposing to take this attribute from employee's calendar in https://github.com/OCA/hr/pull/761. It could also be retrieved on employee's job or directly setting the max time employee per employee.
Since there are several criteria on how to decide the maximum hours an employee should work in a day I decided to create this hook and let everyone chose how they prefer to compute this field.

@etobella @AaronHForgeFlow


etobella (approved):
LGTM :+1: Just a small hook

AaronHForgeFlow (approved):
Looks good!

PR #787, opened 2020-02-11
[12.0][FIX] hr_employee_display_own_info : can't see fields

Saran440:
This PR fixed bug user is not `hr.group_hr_user` and can't see fields in `Private Information` tab
Selection_001

In core odoo write groups `hr.group_hr_user` each field in python code.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[settings]
+known_third_party = babel,dateutil,mock,odoo,openupgradelib,psycopg2,pytz,setuptools,werkzeug
--- a/hr_employee_display_own_info/README.rst
+++ b/hr_employee_display_own_info/README.rst
@@ -55,6 +55,7 @@ Contributors
 ~~~~~~~~~~~~
 
 * Andrea Stirpe <a.stirpe@onestein.nl>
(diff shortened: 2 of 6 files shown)


astirpe:
The fix proposed in the `hr_employee.py` file are actually modifying the security in the ORM layer introduced with https://github.com/odoo/odoo/commit/508e9dd9fe79379540fb90049a70780f89c5ad43#diff-7fe5c1ac38eeb355c75fa18a98b62323.
With your proposed fix, an employee user without HR officer/manager rights can for example search on fields containing sensitive data of other users. For example an employee can filter employees in advanced searches, like employees that has `Identification No = AB12345` or `birthday >= 01/04/2000` or `bank_account_id = AB1234567890` etc... This because the sensitive fields are not protected anymore by the Officer group and are now accessible to every user.

The original idea of this module was to enable only the visibility of the fields that were allowed to be read by the ORM (no group in field definition) but that were hidden in the view by groups attribute in the form.

So if you want to propose such ORM changes, I think you should state in the README what are the consequences of this.

Saran440:
@astirpe okay, I got it.
employee user without HR officer/manager rights should read self information only.
I'll give it a shot. thank you for your suggestion :)

PR #788, opened 2020-02-12
[FIX] hr_holidays_public: avoid object mixing

chienandalu:
Whenever a holiday and a leave were overlaping we could get an object
mixing error. We solve it passing it to the class Intervals as a
resource.calendar.leaves object with its proper date info.

Solves https://github.com/OCA/hr/issues/646

cc @Tecnativa TT21352

### DIFF
--- a/hr_holidays_public/models/resource_calendar.py
+++ b/hr_holidays_public/models/resource_calendar.py
@@ -2,7 +2,7 @@
 # Copyright 2018 Brainbean Apps
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, models
+from odoo import models
 from odoo.addons.resource.models.resource import Intervals
 
 from pytz import timezone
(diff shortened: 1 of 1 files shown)


jarroyomorales:
Do you mean https://github.com/OCA/hr/issues/646 ?

chienandalu:
Sure, my bad :)

jarroyomorales (approved):
LGTM! Maybe someone that uses this last field of Intervals to get more info about where does the interval come from or something will have some problems but at least we solve the mixing error.

hr_holidays_public/models/resource_calendar.py
```diff
@@ -52,12 +52,16 @@ def _public_holidays_leave_intervals(self, start_dt, end_dt, employee_id,
                             line.date,
                             time.max
                         ).replace(tzinfo=tz),
-                        line
+                        leave_obj.new({
+                            'date_from': datetime.combine(
+                                line.date, time(0, 0, 0)),
```
pedrobaeza: This is not taking tzinfo. You can extract previous computations into variables for assigning them several times.

pedrobaeza (changes requested):
Any way, what would happen if we pass an empty record instead of a new one?

hr_holidays_public/models/resource_calendar.py
chienandalu: As you pointed. An passing an empty object does the trick as well

chienandalu:
@pedrobaeza Changes done

Tardo (approved):
:+1: Code review

PR #797, opened 2020-03-04
[13.0][MIG] hr_employee_id

astirpe:
Continue migration from https://github.com/OCA/hr/pull/693


Saran440:
@astirpe Functional Test.
Field `Identification No` shoud be below Tags
or use placeholder function instead (same as Job Position)?
Selection_001

astirpe:
@Saran440 thanks! Change done.

hr_employee_id/views/hr_employee_views.xml
```diff
@@ -6,13 +6,12 @@
...
         <field name="arch" type="xml">
-            <field name="identification_id" position="replace"/>
-            <xpath expr="//div[hasclass('oe_title')]" position="inside">
-                <label for="identification_id" class="oe_edit_only"/>
-                <h3>
-                    <field name="identification_id"/>
-                </h3>
-            </xpath>
+            <field name="category_ids" position="after">
+                <field name="identification_id" position="move"/>
```
Saran440:
```suggestion
                <h3>
                    <field name="identification_id" placeholder="Identification Number"/>
                </h3>
```
What is position move?
I found when select tags. identification_id will show next to tags.

astirpe: Right! Change done! Could you check again?
I used `position="move"` to avoid the use of the `replace` (that was used before).

Saran440: @astirpe it's okay but why are you switch position after -> before?

astirpe: Just for the look&feel, looks better to me. If you prefer to have it after, just let me know. Doesn't really matter to me.

Saran440: Kanban view show tags -> id
Form view show  id -> tags
it should be similar. but I'm okay :)

Saran440 (approved):
Functional Test. LGTM :+1:

alexey-pelykh (changes requested):
Please tidy v13 commit history

astirpe:
@alexey-pelykh done

alexey-pelykh:
Shouldn't it be "isort" commit then migration commit as per https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#technical-method-to-migrate-a-module-from-120-to-130-branch ?

astirpe:
I continued the PR from the job that was already made.

alexey-pelykh:
got it, I wonder if that was worth it 🤷‍♂

astirpe:
I could re-do the job but then someone would argue that I didn't respect the commit history and the community spirit. What about to facilitate this PR? This is a very small module

alexey-pelykh:
So it's "break the migration requirements" vs "commit history". I'd go for second one, on top of that, I think it's possible to rewrite that commit in such way that the original author will remain in git history, no?

PR #604, opened 2019-05-27
[11.0][MIG] hr_contract_hourly_rate

jarroyomorales:
Migration to 11.0 of hr_contract_hourly_rate

PR #798, opened 2020-03-08
[12.0][ADD] hr_contract_rate: flexible wage

alexey-pelykh:
Adds feature to specify contract amount not only as monthly wage, but also rate per various time period. There's a #604 PR for v11, yet it adds another depth level to contracts, while this PR uses vision that change of rate/wage should be accompanied by a new legal contract.

### DIFF
--- a/hr_contract_rate/__init__.py
+++ b/hr_contract_rate/__init__.py
@@ -1,0 +1,4 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from . import models
+from .hooks import pre_init_hook, post_init_hook
--- a/hr_contract_rate/__manifest__.py
+++ b/hr_contract_rate/__manifest__.py
@@ -1,0 +1,24 @@
+# Copyright 2020 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'HR Contract Rate',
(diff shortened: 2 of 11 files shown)


alexey-pelykh:
cc @jarroyomorales since you're the author of #604

jarroyomorales (approved):
Tested and code review and LGTM, way better than `hr_contract_hourly_rate` module.
What do we do with it then? Should I close the PR and backport this or we keep the other module on 11 and start using this in 12.0?

alexey-pelykh:
@jarroyomorales thanks for your review and feedback. If it's possible to make a migration script from `hr_contract_hourly_rate`  to `hr_contract_rate`, I'd go with backporting route, yet I don't really think it's worth it at all: the migration would have to flatten the "rate class" to contracts, yet that feels like "too much"?

jarroyomorales:
Well, the problem its the same, creating a migration script from 10.0 to 11.0 or from 11.0 to 12.0 so I would go for the backporting too.
But yes, I also agree than it would be a heavy migration script for something very simple, maybe we can survive without it and users can migrate the important columns manually?

alexey-pelykh:
I think this data migration should be part of OpenUpgrade then? Technically, it's "making a contract duplicate for every rate class in it", yet I think demand for that would be not that big.

alexey-pelykh:
@jarroyomorales feels like backporting it is and let's deal with migration later

sweenya (approved):
Functional test

PR #799, opened 2020-03-09
[12.0][FIX] hr_employee_document: google_drive_link invisible

robinkeunen:
Replacing google drive link in view by nothing, that is deleting, breaks the view inheritance.

-> Making it invisible instead.


alexey-pelykh (approved):

alexey-pelykh:
Valid and trivial fix

PR #804, opened 2020-03-24
[WIP][12.0][ADD] hr_holidays_archive: archive leave requests and allocations

alexey-pelykh:


ovnicraft (approved):

jarroyomorales:
Hi @alexey-pelykh !
I can understand archiving allocations so employees cant ask for leave in specific cases, but I cant find any situation were you would archive holidays. Which would be the case?
Also I think this should only be done by a group, at least Officer I'd say.

PR #809, opened 2020-03-31
[13.0][MIG] hr_experience

flachica:


flachica:
Excuse me. I close this PR since hr_experience should not have been migrated in this way. Odoo has already incorporated hr_skills

PR #814, opened 2020-04-07
[10.0][FIX] hr_attendance_report_theoretical_time

yostashiro:
To fix the issue of menu path 'Attendaces > Reports' not being visible
to 'Manual Attendance' group.


hr_attendance_report_theoretical_time/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Theoretical vs Attended Time Analysis",
-    "version": "10.0.2.0.0",
+    "version": "10.0.2.0.1",
```
Saran440, line 6: version is not necessary to change (bot will auto generate)

PR #817, opened 2020-04-21
[12.0][FIX] resource_hook : hook function _get_work_hours in day_total

Saran440:
This module compute days mistake when I install 'hr_calendar_rest_time' and 'payroll'

Step to error
1. Config Working Schedule by rest time = 1 hour
Selection_001
2. Generate payslip in payroll
3. Number of Days will compute mistake (exclude rest time)
Selection_002

I think day_total should be compute function similar day_hours.


jarroyomorales (approved):
Nice!

alexey-pelykh (approved):

PR #820, opened 2020-04-22
10.0 HR Quagga Attendance

letyf1988:
This module allows you to register an attendance in kiosk mode by detecting the employee's badge through the quaggajs library


letyf1988:
Hi, is there a way to disable javascript-lint in imported libraries? 
 Also I am informed that the import of the external library is missing, should I use a "require" or something else?What is the correct procedure?

Saran440:
@letyf1988 Amm, Can you try add this in `__manifest__.py`?

**Example :** 
add package python required in module.
'external_dependencies': {'python': ['PyPDF2']},

Selection_001

sherpya:
> @letyf1988 Amm, Can you try add this in `__manifest__.py`?
> 
> **Example :**
> add package python required in module.
> 'external_dependencies': {'python': ['PyPDF2']},
> 
> Selection_001

there are no external python dependencies, look at models.py

```python
from datetime import datetime, timedelta
from odoo import models, fields, api, _
```

PR #824, opened 2020-04-28
[13.0] cleanup gitignore, lib dir should not be ignored (used for javascript 3rd party libraries)

sherpya:
Most entries are only useful if developing python extensions
or using buildout
I've added also .vscode to ignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,30 +2,6 @@
 __pycache__/
 *.py[cod]
 
-# C extensions
-*.so
-
-# Distribution / packaging
-.Python
-env/
-bin/
-build/
-develop-eggs/
-dist/
-eggs/
-lib/
-lib64/
-parts/
-sdist/
-var/
-*.egg-info/
-.installed.cfg
-*.egg
-
-# Installer logs
-pip-log.txt
-pip-delete-this-directory.txt
-
 # Unit test / coverage reports
 htmlcov/
 .tox/
@@ -40,8 +16,10 @@ coverage.xml
 # Pycharm
 .idea
 
-# Mr Developer
-.mr.developer.cfg
+# Visual Studio Code
+.vscode
+
+# Eclipse
 .project
 .pydevproject
 


alexey-pelykh:
I'd suggest making PR to the https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.gitignore

simahawk:
Thanks for your proposal.

This is based on https://github.com/github/gitignore/blob/master/Python.gitignore and IMO there's no reason to drop those lines, but if we go for this change it must be done on all repos.

Anyway, editors should be ignored globally by your own config https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

If you want to propose  common change, pls open a PR following @alexey-pelykh link.

sherpya:
> I'd suggest making PR to the https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.gitignore

this one is ok, at the end has 
```
# OCA rules
!static/lib/
```

sherpya:
> Thanks for your proposal.
> 
> This is based on https://github.com/github/gitignore/blob/master/Python.gitignore and IMO there's no reason to drop those lines, but if we go for this change it must be done on all repos.
> 
> Anyway, editors should be ignored globally by your own config https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
> 
> If you want to propose common change, pls open a PR following @alexey-pelykh link.

ok adding
```
# OCA rules
!static/lib/
```

instead?
alternatively should I need to temp disable git ignore to add lib in a module?

simahawk:
I've proposed this https://github.com/OCA/maintainer-quality-tools/pull/641
Normally we should not have any ext lib on python side included in the module so we tend to support it only for static folder.
You can review it and support it or propose another approach w/ proper motivation ;)

sherpya:
> I've proposed this [OCA/maintainer-quality-tools#641](https://github.com/OCA/maintainer-quality-tools/pull/641)
> Normally we should not have any ext lib on python side included in the module so we tend to support it only for static folder.
> You can review it and support it or propose another approach w/ proper motivation ;)

I've recreated pr adding only:
```
# OCA rules
!static/lib/
```

alternatively please update the hr .gitgnore from https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.gitignore as per @alexey-pelykh suggestion.

I really don't need the cleanup but only to add files in `static/lib/`

sherpya:
> Anyway, editors should be ignored globally by your own config https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

I added only vscode, and the comment to eclipse (entries were altready in), pycharm and others were already in.

PR #831, opened 2020-05-06
[11.0] hr_attendance_lock_date

vrenaville:
* Purpose
Some customers want to prevent employee to modify or create attendance before a locking date

### DIFF
--- a/hr_attendance_lock_date/models/hr_employee.py
+++ b/hr_attendance_lock_date/models/hr_employee.py
@@ -1,9 +1,16 @@
 # Copyright 2019 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import fields, models
+from odoo import fields, models, api
 
 
 class HrEmployee(models.Model):
     _inherit = 'hr.employee'
 
     attendance_lock_date = fields.Datetime('Attendance lock date')
+
+    @api.model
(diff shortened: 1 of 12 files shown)


Tonow-c2c (commented):
Please add one readme with a description of the module usage

jarroyomorales:
Hi @vrenaville ! Wouldnt it be more interesting if it was more dynamic? Instead of setting a date, set a number of days to employees, let's say 15 for example, and lock all attendances that happened before 15 days ago.

vrenaville:
@jarroyomorales 

Thanks for your proposition, the module is more link with invoicing stuff, when the invoice day is planned, manager wants to block on a specific date to be sure attendance will not move.

hr_attendance_lock_date/models/hr_employee.py
```diff
@@ -0,0 +1,16 @@
...
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    attendance_lock_date = fields.Datetime('Attendance lock date')
+
+    @api.model
+    def _cron_lock_all_employee(self):
+        """ Must be called from ir.cron """
+        for employee in self.search([]):
```
jarroyomorales, line 14: Here wouldnt it be better just to do this? So you do just one write instead of one per employee
``` python
self.search([]).write({'attendance_lock_date': fields.Datetime.now()})
```

PR #832, opened 2020-05-06
[13.0][MIG] hr_employee_calendar_planning

jarroyomorales:
Standard migration to 13.0


tonygalmiche (approved):
I tested on my computer. It is OK for me : 
image

image

pedrobaeza:
Are you taking into account new 1 week/2 week calendar feature?

tonygalmiche (changes requested):
@pedrobaeza,
Good remark :)
It does not work :

image

tonygalmiche:
it still does not work : 
image

jarroyomorales:
Hi @pedrobaeza @tonygalmiche ! The idea I implemented for the two week calendars is the following.
- If one of the employee's calendars is a two week calendar, the resulting calendar will be a two week calendar
- Two week calendars are copied as they are onto the new calendar.
- To fit one week calendar lines into the a two week calendar, its lines will be copied twice, one for odd weeks and one for even weeks, so we cover the whole time, which is what the calendar represented being a one week calendar.

This is a test example I did:
Calendar 1: One week calendar, working monday and tuesday from 7 to 14.
image

Calendar 2: Two week calendar, working monday and tuesday even weeks and wednesday, thursdays on odd weeks from 8 to 15.
image

The planning is the following:
image

The resulting calendar is the following:
image

I think the result is as it should be, what do you think?

tonygalmiche (approved):
@jarroyomorales,
It works well as indicated
The specification suits me
It is OK for me
Thank you for this migration

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -68,20 +69,17 @@ class HrEmployeeCalendar(models.Model):
         ),
     ]
 
-    @api.model
```
pedrobaeza: `@api.model` doesn't disappear. Any way, better to decorate it with `@api.model_create_multi` and take advantage of speed optimization in it.

hr_employee_calendar_planning/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils 0.14: http://docutils.sourceforge.net/" />
```
pedrobaeza: Please update your docutils local version or manually restore this value.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -24,23 +59,47 @@ def _regenerate_calendar(self):
...
+                seq += 1
                 vals_list.append((0, 0, data))
+                if duplicate:
+                    data = attendance_line.copy_data(
+                        {
+                            "calendar_id": self.resource_calendar_id.id,
+                            "date_from": line.date_start,
+                            "date_to": line.date_end,
+                            "week_type": "1",
+                            "sequence": seq + 26,
```
pedrobaeza, line 98: Why 26? Use an even bigger hoop (100 or 1000) for being sure (and for having better numbers: 1-101, 2-102...).

pedrobaeza (dismissed):

hr_employee_calendar_planning/models/hr_employee.py
jarroyomorales: Done, I deleted it by mistake when deleting api multis.

hr_employee_calendar_planning/static/description/index.html
jarroyomorales: Updated my docutils to 0.15.1

hr_employee_calendar_planning/models/hr_employee.py
jarroyomorales: Odoo sets the sequence of the Odd Week label line to 25 so I did the same and then forced odd week lines to have a sequence greater than 25. but I can change it to 100 if you prefer.

https://github.com/odoo/odoo/blob/13.0/addons/resource/models/resource.py#L266

pedrobaeza: OK, you can put this as comment for later references and let it this way.

hr_employee_calendar_planning/README.rst
```diff
@@ -0,0 +1,119 @@
...
+During the installation of the module, current working times are split by
+start/end dates for having consistent data, and the potential new composed
+calendar planning is saved instead on the employee.
+
+Configuration
+=============
+
+#. Go to *Employees > Employees*.
+#. Open or create a new one.
+#. On the "Public Information" page, fill the section "Calendar planning" with:
```
chienandalu, line 55:
```suggestion
#. On the "Work Information" tab, fill the section "Calendar planning" with:
```

chienandalu (approved):
Tested in runbot :+1: 

Little remark:

PR #833, opened 2020-05-15
[12.0][IMP] hr_attendance_rfid: Add record rule

brian10048:
@HviorForgeFlow

### DIFF
--- a/hr_attendance_rfid/data/ir_rule.xml
+++ b/hr_attendance_rfid/data/ir_rule.xml
@@ -1,0 +1,15 @@
+<odoo noupdate="1">
+
+    <record id="hr_attendance_rfid_rule" model="ir.rule">
+        <field name="name">RFID Attendance Rule</field>
+        <field name="model_id" ref="hr_attendance.model_hr_attendance"/>
+        <field name="global" eval="False"/>
+        <field name="domain_force">[(1, '=', 1)]</field>
+        <field name="groups" eval="[(4, ref('hr_attendance_rfid.group_hr_attendance_rfid'))]"/>
+        <field name="perm_read" eval="True"/>
+        <field name="perm_write" eval="True"/>
+        <field name="perm_create" eval="True"/>
+        <field name="perm_unlink" eval="False"/>
+    </record>
+
+</odoo>
--- a/hr_attendance_rfid/__manifest__.py
+++ b/hr_attendance_rfid/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     'name': 'HR Attendance RFID',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.1.0',
     'category': 'Human Resources',
     'website': 'https://github.com/OCA/hr',
     'author': 'Comunitea,'
@@ -19,6 +19,7 @@
     'data': [
         'security/hr_attendance_rfid.xml',
         'security/ir.model.access.csv',
+        'data/ir_rule.xml',
         'views/hr_employee_view.xml',
     ],
 }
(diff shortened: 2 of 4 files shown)


brian10048:
@HviorForgeFlow @JordiBForgeFlow 
This should fix the issue when scanning RFID tags and receiving a security error when the RAS user attempts to make an attendance record from the scan

marcelsavegnago (approved):
Looks great for me

HviorForgeFlow (approved):

HviorForgeFlow:
@jarroyomorales ready to move forward

hr_attendance_rfid/data/ir_rule.xml
```diff
@@ -0,0 +1,15 @@
+<odoo>
```
Saran440:
```suggestion
<odoo noupdate="1">
```

Saran440 (commented):
Code Reviewed.

hr_attendance_rfid/data/ir_rule.xml
etobella: If we are adding the noupdate, we should need to add a migration script that will create the rule

HviorForgeFlow: Really? even if that `ir.model.data` record is not already set on the system? OUCH

etobella: Or the forcecreate, but I would opt for the migration

etobella: Something as hard as this:
https://github.com/OCA/community-data-files/blob/12.0/account_payment_unece/migrations/12.0.1.0.0/post-migration.py#L8 :wink:

Saran440: I don't know about migration script.
if you add `noupdate` in code, Will it cause problems?

jarroyomorales: > Something as hard as this:
> https://github.com/OCA/community-data-files/blob/12.0/account_payment_unece/migrations/12.0.1.0.0/post-migration.py#L8 wink

I think this is the best option, better than the force_create

jarroyomorales (commented):
Nice! Thank you!

You also need to upgrade manually the manifest version to 12.0.1.0.2 and change the name of the migration folder to the same version (12.0.1.0.2). Then this will be merged with the nobump option

brian10048:
> 
> 
> Nice! Thank you!
> 
> You also need to upgrade manually the manifest version to 12.0.2.0.1 and change the name of the migration folder to the same version (12.0.2.0.1). Then this will be merged with the nobump option

I was just getting ready to comment and ask if I should manually update the manifest or if the merge bot was going to be told what to do!  Now I have updated and manually bumped the manifest to 12.0.2.0.0 from 12.0.1.0.0

jarroyomorales:
@brian10048 Yes, now we have to agree on the version number. It is not a very huge change so maybe instead of 12.0.2.0.0 we could do 12.0.1.1.0. @Saran440 What do you think?

Saran440:
I agree with @jarroyomorales 
it should be 12.0.1.1.0
and I will merge with nobump

brian10048:
@Saran440 @jarroyomorales - Version changes made as requested

jarroyomorales (approved):

Saran440 (approved):

HviorForgeFlow (approved):

etobella (approved):

etobella:
@OCA/human-resources-maintainers I think this is ready to merge :wink:

PR #834, opened 2020-05-18
[IMP] hr_family: Groups for family now could be extended

luistorresm:
In v10> is not possible inherit a group by string, for this reason was
added a name in the groups.


hugho-ad (approved):

moylop260 (approved):
LGTM 👍

elrodo1 (approved):

PR #841, opened 2020-05-20
[12.0][Fix] Fix hr_attendance_report_theoretical_time pivot report

ghost:
This fixes an issue when printing the xls report of the pivot view from hr_attendance_report_theoretical_time

### DIFF
--- a/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
+++ b/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report_views.xml
@@ -20,7 +20,7 @@
     <record id="hr_attendance_theoretical_view_pivot" model="ir.ui.view">
         <field name="model">hr.attendance.theoretical.time.report</field>
         <field name="arch" type="xml">
-            <pivot disable_linking="1">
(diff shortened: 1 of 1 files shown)


jarroyomorales:
@busmanapps I have downloaded the pivot view as xls without this fix in version 12.0 and it has worked properly for me... Maybe it just fails in some cases. Anyways this doesn't do any bad but string should be `Theoretical vs Attended Time Analysis` instead of just `Attendances` in my opinion.

busmanapps:
> @busmanapps I have downloaded the pivot view as xls without this fix in version 12.0 and it has worked properly for me... Maybe it just fails in some cases. Anyways this doesn't do any bad but string should be `Theoretical vs Attended Time Analysis` instead of just `Attendances` in my opinion.

In past versions it was "Attendances" that's why we put that, but yes I think "Theoretical vs Attendend Time Analysis" could be good too

jarroyomorales:
@busmanapps Could you make the change? Thank you!

jarroyomorales (approved):

pedrobaeza:
Printing the XLS report? I don't get the problem...

PR #849, opened 2020-06-17
[13.0][ADD] hr_org_chart_overview

AdriaGForgeFlow:
New module that adds an Organization Chart Overview in v13.

cc ~ @ForgeFlow @LoisRForgeFlow


jcdrubay (commented):
I tested on Runbot, it is very nice.

Maybe you could add to the Readme which module to click on.

Also, if someone agrees that it might not be easy to understand, mention that the chart can be expended / collapsed by clicking on the arrows on the sides of each employee.

jcdrubay:
The pictures did not show in the org-chart

image

AdriaGForgeFlow:
> I tested on Runbot, it is very nice.
> 
> Maybe you could add to the Readme which module to click on.
> 
> Also, if someone agrees that it might not be easy to understand, mention that the chart can be expended / collapsed by clicking on the arrows on the sides of each employee.

Agreed, and done :+1:

AdriaGForgeFlow:
> The pictures did not show in the org-chart
> 

I'll look into it. It works for me on my local environment. Maybe it has something to do with runbot's attachments, not sure.

Saran440 (approved):
Functional Test. LGTM :+1: 
I tested on localhost. It can see image employee.

@AdriaGForgeFlow could you add test script in this module?

AdriaGForgeFlow:
> The pictures did not show in the org-chart
> 


Should be working now, can you check @jcdrubay ? Thanks!

mymage:
Tested on runbot and look nice.
Imho, I would set an option to the show employees image.
Question: what about configure the first employee with a checkbox [ie. head of organization] on the job position? And as consequence this job position can have only one employee.

jcdrubay:
The picture shows nicely.

I guess a configuration of hide/show select size to display the picture would be nice to have. Maybe to be added in the Roadmap.

A point  that could be mentioned in the readme is the usage or limitation for multiple trees. Indeed, the Chart does not seem to support having 2 employees which don't have managers as the head of the trees, in such case only one tree is shown. Anyway, I am not sure how many organizations are in this situation.

jcdrubay:
@mymage the employee on top is the employee without manager.

image

AaronHForgeFlow (approved):
Functional test :+1:  LGTM. Super cool BTW :)

JordiBForgeFlow (approved):

Issue #857, opened 2020-07-08
[12.0]  hr_attendance_modification_tracking: Issue on las upgrade


etobella:
With the new computed field, it takes too long to upgrade / install if there is  existing records (a large set, obviously).

I think we need to create a `pre_init_hook` and a `pre-migration` script creating the column in order to avoid this issue.

@js-landoo @jarroyomorales @alexey-pelykh @Saran440 @marcelsavegnago

jarroyomorales:
Yes, I also think we need a pre init for new instalations and a migration script for upgrades. Adding the new column will prevent the new stored computed field be calculated in old attendances, which is nice for a fast installation. This should be added to the readme so whoever wants to compute old attendances does it manually.

The other option is not doing anything but at least adding on the readme that in can take a long time for instances with a lot of attendances...

Saran440:
Fixed https://github.com/OCA/hr/pull/858

PR #858, opened 2020-07-08
[12.0][IMP] hr_attendance_modification_tracking: add manual column pre-install

jarroyomorales:
Pre init hook and migration script to tackle issue https://github.com/OCA/hr/issues/857

@etobella @js-landoo @ns-landoo @Saran440

### DIFF
--- a/hr_attendance_modification_tracking/migrations/12.0.1.2.0/pre-migration.py
+++ b/hr_attendance_modification_tracking/migrations/12.0.1.2.0/pre-migration.py
@@ -1,0 +1,14 @@
+# Copyright 2020 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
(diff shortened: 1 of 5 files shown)


etobella (approved):
LGTM :+1:

AdriaGForgeFlow (approved):
Code review :+1:

etobella:
@OCA/human-resources-maintainers Can you review please, as it is making upgrade hard

AaronHForgeFlow (approved):
:+1:  code review

hr_attendance_modification_tracking/migrations/12.0.1.2.0/pre-migration.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Creu Blanca
```
MiquelRForgeFlow, line 1: I think folder name should be '12.0.1.1.0', because the `time_changed_manually` field was added in that version.

etobella: The version should be the destination version where it was implemented, so I think this is fine

dreispt (approved):

PR #865, opened 2020-08-05
[ADD] hr_gamification_email_notification

Khalid-SerpentCS:
This module Sends Email to all the Employees with all the award given in the last week.


hr_gamification_email_notification/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+
+{
+    "name": "OSI HR Gamification Extended",
+    "version": "12.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Open Source Integrators",
+    "summary": """Send Email to all the Employees with all the award given in
+     the last week.""",
+    "category": "Helpdesk",
+    "maintainers": "Open Source Integrators",
```
pedrobaeza: This key is incorrect. It must be a GitHub user, and with a list, not a simple string.

PR #861, opened 2020-07-21
[12.0][ADD] hr_attendance_sheet

patrickrwilson:
Adds HR Attendance Sheet module.

This module expands hr_attendance to incorporate sheets. An attendance sheet has a period of time specified in the general settings and incorporates the attendance entries that fall within that period. 

This module incorporates some sheet features:
- Sheets have an approval process also defined in the general settings. Approvals can be done by HR, Managers or HR and Managers.
- Auto Lunch is an option to calculate a lunch period if the attendance exceeds a specified duration.
- [UPDATE] Removed to put into own module] Split Overnight Attendance is an option to split the attendance at midnight if it crosses overnight. This is useful to prevent an attendance from crossing into another sheet period.
- [UPDATE] Added an 'Attendance Admin' field on departments to allow an employee to be assigned to a department that would administer that departments employee timesheets. This is in addition to the manager who can already review/approve them.

Credit is due to the author/maintainers of the timesheet_sheet module as much of the code in that module was referenced to create this one since it had similar functionalities: Miquel Raïch, Andrea Stirpe, Lois Rilo, Alexey Pelykh, Dennis Sluijk

PR #875, opened 2020-08-24
[12.0][ADD] hr_attendance_split

patrickrwilson:
Related to PR #861 
This module was originally part of the attendance sheet but reviewers wished it to be in it's own standalone module, so this PR is for that purpose.

This is in alpha state as i'm sure adjustments will be needed as well as tests created in order for it to come out of alpha state.


patrickrwilson:
ping @jarroyomorales @max3903 @osi-scampbell

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2020 Pavlov Media
+# License LGPL-3 - See http://www.gnu.org/licenses/lgpl-3.0.html
+
+from odoo import api, fields, models, _
```
max3903, line 4:
```suggestion
from odoo import api, fields, models
```

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,102 @@
...
+        attendance at midnight of the employees timezone."""
+        for rec in self:
+            if 'check_out' in vals:
+                check_out = vals.get('check_out')
+            else:
+                check_out = rec.check_out
+            if 'check_in' in vals:
+                check_in = vals.get('check_in')
+            else:
+                check_in = rec.check_in
```
max3903:
```suggestion
            check_in = vals.get('check_in', False) or rec.check_in
```

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @api.multi
+    def write(self, vals):
+        """ If the user clocks out after midnight, then it will split the
+        attendance at midnight of the employees timezone."""
+        for rec in self:
+            if 'check_out' in vals:
+                check_out = vals.get('check_out')
+            else:
+                check_out = rec.check_out
```
max3903:
```suggestion
            check_out = vals.get('check_out', False) or rec.check_out
```

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,102 @@
...
+                        'check_out': new_check_out,
+                        'split_attendance': False})
+        res = super(HrAttendance, self).write(vals)
+        return res
+
+    @api.model
+    def create(self, vals):
+        """ If during creation, check-out crosses overnight, then make sure
+        the clock-out is at local midnight to prevent overnight attendances"""
+        if 'check_out' in vals and vals.get('check_out', False):
```
max3903:
```suggestion
        if vals.get('check_out', False):
```

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,102 @@
...
+        string="Company",
+        related="employee_id.company_id")
+
+    def _get_attendance_employee_tz(self, date=None):
+        """ Convert date according to timezone of user """
+        tz = False
+        if self.employee_id:
+            tz = self.employee_id.tz
+        if not date:
+            return False
```
max3903:
```suggestion
        if not date:
            return False
        tz = False
        if self.employee_id:
            tz = self.employee_id.tz
```

jarroyomorales:
@patrickrwilson Hi! I was trying to test it on runbot but I got this error.
To reproduce, I created an attendance with a check in date, saved, and the added the checkout on the following day, expecting it to split the attendance in two.

image

patrickrwilson:
Thanks @jarroyomorales I can reproduce so i'll look into it.

patrickrwilson:
@jarroyomorales I think I fixed it. After fighting with the create and write methods I decided to change it and move it into the compute method, like I originally had it in the sheets module. It seems to work better for me.

jarroyomorales:
@patrickrwilson Hi! Im sorry but I am not being able to make it work. I have checked the 'split_attendances'  option but if i create an attendance with a check in date, save, and the add the checkout on the following day, it is not splitted. Im testing on runbot so maybe the error is due to another module incompatibility?

Another thing is that i dont find the `def _compute_duration(self)` function on the odoo core, so I dont know when it should be computed, is the name correct?

patrickrwilson:
@jarroyomorales I'm sorry, I don't know what I was thinking. For some reason I was referencing the hr_attendance_sheet method for duration but we wanted to have this standalone, so i'm not sure why I did that. 

The last commit works without the need for the attendance sheet module and should now be fully standalone. I'm not sure if the way I did it was the best way but it seems to work (I think).

jarroyomorales:
@patrickrwilson I think the best place to implement this functionality is in the `write` method of `hr_attendance` since writing and creating records on a compute function hardly ever is the best option. Also I think the field `need_split` is not necessary.
In hr_attendance's write function: `if vals.get('check_out', False): Check split`

hr_attendance_split/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,27 @@
...
+                    <div class="col-xs-12 col-md-6 o_setting_box">
+                        <div class="o_setting_left_pane">
+                            <field name="split_attendance"/>
+                        </div>
+                        <div class="o_setting_right_pane">
+                            <label for="split_attendance" string="Split Overnight Attendance"/>
+                            <span class="fa fa-lg fa-building-o" title="Values set here are company-specific." aria-label="Values set here are company-specific." groups="base.group_multi_company" role="img"/>
+                            <div class="text-muted">
+                                Choose if attendances will be split if the start/end times cross over midnight.
+                                Example: User clocks in at 11pm, then clocks out 9am next day, two attendance would be created with the first clock-out at midnight and next clock-in at 12:01am.
```
jarroyomorales: This line is not correct is it? The module sets the check-out at 23:59, and the check-in at midnight (00:00).

patrickrwilson:
@jarroyomorales I originally put this in a write method however it wouldn't trigger if a user manually created an attendance that overlapped (Like a manager creating it for a missed day or something). It only worked during normal check-in then check-out. When I also tried to add it to the create method it caused errors and I think it was because there is also a create in the write method and as they looped it just didn't work right.

I'm fine with having it in the write method and just not in a create method if you are.

patrickrwilson:
@jarroyomorales Oh, I just remembered that the other problem with only having it in the write was that if the attendance was more than one day (Like check-in was 9/19 and check-out was 9/21), it wouldn't continue to split them for each day where as the compute does.

That was another reason why I abandoned the write/create methods. Ideas?

jarroyomorales:
Yes, thinking about it i think its correct if it doesnt trigger on manual creations with a check in and a check out. We could consider that case as forcing the dates.

And about the loop I think that if we make a version that splits the attendances properly this problem shouldn't appear. So tell me if you need anything!

jarroyomorales:
> @jarroyomorales Oh, I just remembered that the other problem with only having it in the write was that if the attendance was more than one day (Like check-in was 9/19 and check-out was 9/21), it wouldn't continue to split them for each day where as the compute does.
> 
> That was another reason why I abandoned the write/create methods. Ideas?

A dirty solution to this is when splitting an attendance and creating the 00:00 - XX:XX attendance, first create the attendance with the 00:00 check-in date and False checkout, and then write the XX:XX checkout date to trigger this again. I dont know if I explained myself properly.

jarroyomorales:
Also maybe it would be great to link the split attendances, to know which is the second part of an attendance.

patrickrwilson:
@jarroyomorales phew, I think I got it. This commit moves the function back to the write and your suggestion to write the check_out separately works like a charm. 

I added the create to just prevent the user from creating attendances that go overnight if they try to manually do it, which I thought was ok, since if they are manually adding attendances, then they probably should do it the manual way anyway. 

I also added the field to reference to the attendance it creates.

hr_attendance_split/views/res_config_settings_views.xml
patrickrwilson: I missed this, you are correct. I corrected the wording.

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,93 @@
...
+            if employee.company_id.split_attendance and \
+                    check_in_date != check_out_date:
+                raise ValidationError(_("Cannot create attendance that "
+                                        "starts and ends on different days."))
+        return super().create(vals)
+
+    @api.multi
+    def write(self, vals):
+        res = super().write(vals)
+        """ If the user clocks out after midnight, then it will split the
```
jarroyomorales, line 45: TravisCI is complaining about this line. I think this triple quote string should go on the first line of the function or as a comment with #

hr_attendance_split/models/hr_attendance.py
```diff
@@ -0,0 +1,93 @@
...
+        attendance_dt = datetime.strptime(str(date),
+                                          DEFAULT_SERVER_DATETIME_FORMAT)
+        attendance_tz_dt = pytz.UTC.localize(attendance_dt)
+        attendance_tz_dt = attendance_tz_dt.astimezone(time_zone)
+        attendance_tz_date_str = datetime.strftime(
+            attendance_tz_dt, DEFAULT_SERVER_DATE_FORMAT)
+        return attendance_tz_date_str
+
+    @api.model
+    def create(self, vals):
```
jarroyomorales, line 27: I am not sure on what to do here. I think we should allow a user to do this if he is manually creating it on purpose

jarroyomorales:
@patrickrwilson H! I just tested it and it works as expected! Thank you very much!

The only case where it is not working properly is with attendances implying more than 2 days. I get the error `"Check Out" time cannot be earlier than "Check In" time. `. It may be because of timezones.

hr_attendance_split/models/hr_attendance.py
patrickrwilson: I've moved the comment to the top.

hr_attendance_split/models/hr_attendance.py
patrickrwilson: Here's my opinion, if the company wants split attendance then it shouldn't allow them to manual do it either. They would get the warning, then the have to manually create multiple attendances just like it automatically does when a user is clocking in/out on different days. I think if we let them enter any in there then it goes against the purpose of this module. 
If it is decided it's best to leave this out, it's easy to remove it.

jarroyomorales: Okay, we can leave it like this for now then. Let's see if someone else has an opinion on this.

patrickrwilson:
> The only case where it is not working properly is with attendances implying more than 2 days. I get the error `"Check Out" time cannot be earlier than "Check In" time. `. It may be because of timezones.

@jarroyomorales I'm not able to reproduce this, it was able to do 5days worth of attendances. Can you provide the tz and start/end times you tried?

Issue #882, opened 2020-09-04
hr_period: date_range.py error when create period


Dariuzcm:
I'd an error when I tried to create a period,, it happend becouse when date_from or date_end are changed  try to assign hr.period object into an date.range object , on next the images, also happends into hr_fiscalyear by the way.

![Captura de pantalla de 2020-09-04 12-51-06](https://user-images.githubusercontent.com/28762220/92271479-7b15ca80-eead-11ea-9ec3-0f1b5369b9b8.png)
![Captura de pantalla de 2020-09-04 12-51-44](https://user-images.githubusercontent.com/28762220/92271486-7fda7e80-eead-11ea-9e8d-a13decb0fca0.png)

Error here:
Traceback (most recent call last):
  File "/.repo_requirements/odoo/odoo/http.py", line 642, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/.repo_requirements/odoo/odoo/http.py", line 684, in dispatch
    result = self._call_function(**self.params)
  File "/.repo_requirements/odoo/odoo/http.py", line 334, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/http.py", line 327, in checked_call
    result = self.endpoint(*a, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 942, in __call__
    return self.method(*args, **kw)
  File "/.repo_requirements/odoo/odoo/http.py", line 507, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 895, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 887, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 689, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 680, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/models.py", line 5555, in onchange
    record._onchange_eval(name, field_onchange[name], result)
[...]

PR #883, opened 2020-09-04
[FIX] hr_period: Error in onchange is fixed.

luistorresm:
Hr.period inherit from date.range model from server-tools, and, in
date_range is added an onchange to assign a possible parent when update
dates, type, or company. But an error is returned when trying to assign the
search result because try to assign:

`date.range = hr.period`

To avoid this, now the field parent_id is assigned in hr_period, and
with this avoids the error.

Fixes https://github.com/OCA/hr/issues/882

### DIFF
--- a/hr_period/models/hr_period.py
+++ b/hr_period/models/hr_period.py
@@ -81,6 +81,8 @@ class HrPeriod(models.Model):
         domain=[('hr_period', '=', True)],
         default=_default_type
     )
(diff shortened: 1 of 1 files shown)


Dariuzcm (approved):

Dariuzcm:
It works!!

hugho-ad (approved):

marcelsavegnago (approved):

PR #885, opened 2020-09-14
[IMP] hr_expense_invoice: onchange invoice

chienandalu:
Get expense amount from invoice amount. Otherwise it will do a mismatch when trying to post the account move.

cc @Tecnativa TT25442

### DIFF
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -1,5 +1,6 @@
-# Copyright 2015 Pedro M. Baeza <pedro.baeza@tecnativa.com>
-# Copyright 2017 Vicent Cubells <vicent.cubells@tecnativa.com>
+# Copyright 2015 Tecnativa - Pedro M. Baeza
+# Copyright 2017 Tecnativa - Vicent Cubells
+# Copyright 2020 Tecnativa - David Vidal
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models, _
@@ -17,7 +18,6 @@ class HrExpense(models.Model):
         copy=False,
     )
 
-    @api.multi
     @api.constrains('invoice_id')
     def _check_invoice_id(self):
         for expense in self:  # Only non binding expense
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

chienandalu:
@pedrobaeza I added a little change for defaulting to the invoice analytic account as well

hr_expense_invoice/models/hr_expense.py
```diff
@@ -38,3 +37,14 @@ def _get_account_move_line_values(self):
...
+    @api.onchange("invoice_id")
+    def _onchange_invoice_id(self):
+        """Get expense amount from invoice amount. Otherwise it will do a
+           mismatch when trying to post the account move. We also default
+           to the invoice analytic account although it can be changed if
+           needed"""
+        if self.invoice_id:
+            self.quantity = 1
+            self.unit_amount = self.invoice_id.amount_total
+            self.analytic_account_id = self.invoice_id.analytic_account_id
```
pedrobaeza: AFAIK, there's no analytic account at invoice level, but line level.

chienandalu: True. I took it for granted... changed

pedrobaeza: Anyway, it should be the same, as this module overrides account.move.line creation for linking the already created ones in invoice, so the analytic account will be used. Am I correct?

chienandalu: But we'd have the problem with multiple lines. Wich one has the correct analytic account?

pedrobaeza: It doesn't matter, as the journal items will be generated from the invoice, not from the expense. Take into account that the analytic account goes to the expense account, not to the payable.

victoralmau (approved):

rafaelbn (changes requested):
Although the analytical account is informed. And also the accounting account gives an error.

rafaelbn (approved):
Functionall review :+1: 

Previous commment was with demo data.

RTM

rafaelbn:
@kittiu can you review to get this merge?

kittiu (approved):
LGTM

marcelsavegnago (commented):
LGTM

PR #889, opened 2020-09-24
[FIX] hr_expense_advance_clearing: analytic account visibility

chienandalu:
We should prevent setting the analytic account in expenses used as
employees advance. Otherwise, it can lead to errors charging against an
analytic account that later won't be compensated when all is reconcile.
Anyway, the advance is not an expense, but money of the company moving
from one place to another, that later will be used for real expenses
that will be imputed to the corresponding analytic account

cc @Tecnativa TT25442

### DIFF
--- a/hr_expense_advance_clearing/views/hr_expense_views.xml
+++ b/hr_expense_advance_clearing/views/hr_expense_views.xml
@@ -37,6 +37,9 @@
             <h1 position="after">
                 <field name="advance"/><label for="advance"/>
             </h1>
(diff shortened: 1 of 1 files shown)


chienandalu:
@kittiu What do you think about this?

kittiu (approved):
Agreed

kittiu:
cc @Saran440

pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

PR #908, opened 2020-10-09
[MIG] hr_branch: Migration to 14.0

luistorresm:


luistorresm:
hello @moylop260 

What do you think?

Is a module that I was added on v13.

Regards

PR #909, opened 2020-10-12
[12.0][FIX] hr_expense_payment_difference, correct initial amount to pay

kittiu:
This is problem even for standard Odoo. And with this payment difference addition, it should be correct too because it will be even more confuse for payment_difference case.

By standard. I..e,
1. Expense = 200
2. Register Payment = 150
3. 2nd time, Register Payment will still be 200. This is because, Odoo use default_amount = total_amount of expense.

And that is not OK.

Fixed by adding amount residual like in invoice. So the 2nd time register payment = 50. https://github.com/OCA/hr/pull/909/commits/eaa07533d7b06d4ab076e10f689164aa0ac4933c


kittiu:
cc @Saran440

Saran440 (approved):
Functional Test :+1:

ps-tubtim (approved):

PR #914, opened 2020-10-15
[13.0] [MIG] hr_skill

JevinD:
hr_skill migration from 12.0 to 13.0


JevinD:
@bodedra Travis is green for this module. Core odoo also has a hr_skills but the functionality is different.

bodedra:
Odoo v13.0 introduce "hr_skills" module which has covered most of OCA "hr_skill" module functionality. 

So IMO, OCA "hr_skill" module can be archive.

@OCA/human-resources-maintainers @dreispt @alexey-pelykh any suggestion?

tonygalmiche (approved):
LGTM : Code review and functional review with runbot

image

tonygalmiche:
> Odoo v13.0 introduce "hr_skills" module which has covered most of OCA "hr_skill" module functionality. 
> So IMO, OCA "hr_skill" module can be archive.

I think it is good to leave the choice to the user and this also makes it possible to facilitate migrations for those who have used this module

However, this module seems incompatible with "hr_skills". I couldn't install "hr_skills"
```
  File "/.repo_requirements/odoo/odoo/addons/base/models/ir_ui_view.py", line 987, in postprocess_and_fields
    self.raise_view_error(message, view_id)
  File "/.repo_requirements/odoo/odoo/addons/base/models/ir_ui_view.py", line 614, in raise_view_error
    raise ValueError(message)
ValueError: Field `level` does not exist
Error context:
View `hr.employee.view.form.inherit.resume`
[view_id: 500, xml_id: n/a, model: hr.employee, parent_id: 411]
```
It should perhaps be specified in  README.rst ?

bodedra:
@tonygalmiche Thanks for your functional review. I agree with your point.

@JevinD Would you please add "excludes" key in manifest file. Here is example: https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L30

    "excludes": [
        "hr_skills",
    ],

Saran440:
I agree with @bodedra. 
In odoo13 has covered most of this module.
we should be used core odoo or depend it to do something.

JevinD:
> @tonygalmiche Thanks for your functional review. I agree with your point.
> 
> @JevinD Would you please add "excludes" key in manifest file. Here is example: https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L30
> 
> ```
> "excludes": [
>     "hr_skills",
> ],
> ```

I have added this to the manifest.

JevinD:
@bodedra I have added the exclude to the manifest

PR #931, opened 2020-11-11
[12.0][FIX] hr_employee_calendar_planning: generate when needed

jarroyomorales:
Underlying calendars should only be regenerated if some attendance changes on a parent calendar. Now they are getting regenerated everytime something changes in the `resource.calendar`, which I think it's not necessary.

@pedrobaeza @etobella @Saran440

### DIFF
--- a/hr_employee_calendar_planning/models/resource_calendar.py
+++ b/hr_employee_calendar_planning/models/resource_calendar.py
@@ -14,6 +14,8 @@ class ResourceCalendar(models.Model):
     @api.multi
     def write(self, vals):
         res = super(ResourceCalendar, self).write(vals)
+        if 'attendance_ids' not in vals:
+            return res
         for record in self.filtered('active'):
             calendars = self.env['hr.employee.calendar'].search([
                 ('calendar_id', '=', record.id)


pedrobaeza (commented):
For reducing diff, do an earlier return instead.

Not sure if this will cover all cases. Another thing I have in my bag is to synchronize tz between the original calendar and the generated one.

jarroyomorales:
@pedrobaeza Change done! I have tried adding, removing and editing lines and it's working as expected, for now I havent found any case where it misbehaves.
And regarding the time zones, it is true that it's something to add in the roadmap.

pedrobaeza:
If the calendar lines are changed by code in the children model, the automatic calendar is changed (this is not something of this PR, it's a general doubt)?

pedrobaeza (approved):

etobella (approved):

pedrobaeza:
To be fw-port to v13 as well

jarroyomorales:
> If the calendar lines are changed by code in the children model, the automatic calendar is changed (this is not something of this PR, it's a general doubt)?

I think it is not regenerated, since the write of the `resource.calendar` is not executed

pedrobaeza:
Yeah, that's what I though, maybe we should add it, at least mention it on roadmap.

jarroyomorales:
@pedrobaeza This was solved like this in v13.0 https://github.com/OCA/hr/blob/13.0/hr_employee_calendar_planning/models/resource_calendar.py#L14 during the migration. At the end it is exactly the same, so I think we can keep it like this and save a PR.

pedrobaeza:
Great then, thanks

PR #946, opened 2020-11-27
[13.0][IMP] hr_employee_document: Add documents button in user profile

victoralmau:
Add documents button in user profile

Please @joao-p-marques and @pedrobaeza review it.

@Tecnativa TT26974

### DIFF
--- a/hr_employee_document/models/hr_employee.py
+++ b/hr_employee_document/models/hr_employee.py
@@ -4,33 +4,23 @@
 from odoo import fields, models
 
 
-class HrEmployee(models.Model):
-    _inherit = "hr.employee"
+class HrEmployeeBase(models.AbstractModel):
+    _inherit = "hr.employee.base"
 
-    document_ids = fields.One2many(
-        "ir.attachment",
-        compute="_compute_document_ids",
-        groups="hr.group_hr_user",
-        string="Documents",
-    )
-    documents_count = fields.Integer(
-        compute="_compute_document_ids",
-        groups="hr.group_hr_user",
-        string="Document Count",
+    document_count = fields.Integer(
+        compute="_compute_document_count", string="Document Count",
     )
 
-    def _compute_document_ids(self):
-        attachments = self.env["ir.attachment"].search(
-            [("res_model", "=", self._name), ("res_id", "in", self.ids)]
+    def _compute_document_count(self):
+        self.documents_count = 0
+        attachment_groups = self.env["ir.attachment"].read_group(
+            [("res_model", "=", "hr.employee"), ("res_id", "in", self.ids)],
+            ["res_id"],
+            ["res_id"],
         )
-
-        result = dict.fromkeys(self.ids, self.env["ir.attachment"])
-        for attachment in attachments:
-            result[attachment.res_id] |= attachment
-
-        for employee in self:
-            employee.document_ids = result[employee.id]
-            employee.documents_count = len(employee.document_ids)
+        count_dict = {x["res_id"]: x["res_id_count"] for x in attachment_groups}
+        for record in self:
+            record.document_count = count_dict.get(record.id, 0)
 
     def action_get_attachment_tree_view(self):
         action = self.env.ref("base.action_attachment").read()[0]
--- a/hr_employee_document/models/hr_employee_public.py
+++ b/hr_employee_document/models/hr_employee_public.py
@@ -1,0 +1,33 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrEmployeePublic(models.Model):
+    _inherit = "hr.employee.public"
+
+    is_logged = fields.Boolean(compute="_compute_is_logged", store=False)
+
+    def _compute_is_logged(self):
+        self.is_logged = False
+        for record in self:
+            if self.env.user == record.user_id:
+                record.is_logged = True
+
+    def action_get_attachment_tree_view(self):
+        action = self.env.ref("base.action_attachment").read()[0]
+        action["context"] = {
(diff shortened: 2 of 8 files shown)


pedrobaeza:
Why in user profile instead of the employee? The idea was to have visibility profiles, as there can be documents that the employee can't even see. Only HR managers should see them.

hr_employee_document/models/hr_employee_public.py
```diff
@@ -0,0 +1,54 @@
...
+        for attachment in attachments:
+            result[attachment.res_id] |= attachment
+
+        for employee in self:
+            employee.document_ids = result[employee.id]
+            employee.documents_count = len(employee.document_ids)
+
+    def _compute_is_logged(self):
+        for item in self:
+            item.is_logged = False
```
pedrobaeza: You can do outside the loop (new ORM allows it):
```suggestion
        self.is_logged = False
```

hr_employee_document/models/hr_employee_public.py
```diff
@@ -0,0 +1,54 @@
...
+            if self.env.user.employee_id.id == item.id:
```
pedrobaeza: Why do you need to compare with `.id`?

And instead of `item`, we usually put `record` as generic word

hr_employee_document/models/hr_employee_public.py
```diff
@@ -0,0 +1,54 @@
...
+    documents_count = fields.Integer(
+        compute="_compute_document_ids", string="Document Count",
+    )
+    is_logged = fields.Boolean(compute="_compute_is_logged", store=False)
+
+    def _compute_document_ids(self):
+        attachments = self.env["ir.attachment"].search(
+            [("res_model", "=", "hr.employee"), ("res_id", "in", self.ids)]
+        )
+
```
pedrobaeza: Please don't let empty lines inside a method.

hr_employee_document/models/hr_employee_public.py
```diff
@@ -0,0 +1,54 @@
...
+        result = dict.fromkeys(self.ids, self.env["ir.attachment"])
+        for attachment in attachments:
+            result[attachment.res_id] |= attachment
+
+        for employee in self:
+            employee.document_ids = result[employee.id]
+            employee.documents_count = len(employee.document_ids)
```
pedrobaeza: This isn't very performant. You can use `read_group` instead grouping by res_id for getting the count (I know this code is copied, but you can improve both).

hr_employee_document/models/hr_employee_public.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2020 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrEmployeePublic(models.Model):
+    _inherit = "hr.employee.public"
+
+    document_ids = fields.One2many(
```
pedrobaeza: This should be `Many2many` instead of `One2many` for avoiding side effects (in both places).

pedrobaeza (commented):
I'm still missing the possibil

hr_employee_document/models/hr_employee_public.py
victoralmau: Finally remove `document_ids` field because is unnecessary.

joao-p-marques (approved):
LGTM :+1:

hr_employee_document/models/hr_employee.py
```diff
@@ -7,30 +7,23 @@
...
-
-        result = dict.fromkeys(self.ids, self.env["ir.attachment"])
-        for attachment in attachments:
-            result[attachment.res_id] |= attachment
-
-        for employee in self:
-            employee.document_ids = result[employee.id]
-            employee.documents_count = len(employee.document_ids)
+        for attachment_group in attachment_groups:
+            record = employee_obj.browse(attachment_group["res_id"])
```
pedrobaeza: Don't rebrowse, as this is breaking caches. Just create a dictionary with current groups:
```python
count_dict = {x["res_id"]: x["res_id_count"] of x in attachment_groups}
for record in self:
    record.documents_count = count_dict.get(record.id, 0)
```

hr_employee_document/models/hr_employee.py
```diff
@@ -7,30 +7,23 @@
 class HrEmployee(models.Model):
     _inherit = "hr.employee"
 
-    document_ids = fields.One2many(
-        "ir.attachment",
-        compute="_compute_document_ids",
-        groups="hr.group_hr_user",
-        string="Documents",
-    )
     documents_count = fields.Integer(
```
pedrobaeza: Please avoid this plural:
```suggestion
    document_count = fields.Integer(
```

hr_employee_document/models/hr_employee_public.py
```diff
@@ -0,0 +1,47 @@
...
+
+class HrEmployeePublic(models.Model):
+    _inherit = "hr.employee.public"
+
+    documents_count = fields.Integer(
+        compute="_compute_documents_count", string="Document Count",
+    )
+    is_logged = fields.Boolean(compute="_compute_is_logged", store=False)
+
+    def _compute_documents_count(self):
```
pedrobaeza: Do you need to redefine this compute ? Can't this be inherited somehow?

pedrobaeza (changes requested):

pedrobaeza (approved):

pedrobaeza:
@sbidoul @gurneyalex ocabot seems dead, as it's not answering commands.

sbidoul:
@pedrobaeza It was stuck waiting for git command to complete, that apparently were not timing out. It's progressing through the queue now. It will take a while to catch up.

pedrobaeza:
OK, thanks for the follow-up.

PR #947, opened 2020-12-01
[FIX] [13.0] hr_employee_calendar_planning: Defuse warning about non-matching calendars.

thomaspaulb:
Disables the warning of 'Calendar Mismatch : The employee's calendar does not match its current contract calendar. This could lead to unexpected behaviors.'.

It makes no sense as the point of this module is to replace the employee's calendar by a generated one.

### DIFF
--- a/hr_employee_calendar_planning/models/__init__.py
+++ b/hr_employee_calendar_planning/models/__init__.py
@@ -1,2 +1,3 @@
+from . import hr_contract
 from . import hr_employee
 from . import resource_calendar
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I'm not sure that `hr.contract` has anything to do with this module...

thomaspaulb:
@pedrobaeza Ah, you're right, it doesn't depend on it. However, installation of this module together with `hr_contract` does result in faulty [warnings](https://github.com/OCA/OCB/blob/13.0/addons/hr_contract/models/hr_contract.py#L65).

How to go about fixing that? An init hook that monkey-patches `hr.contract` when it exists, and otherwise not? Or a `depends_if_installed`?

pedrobaeza:
You need a glue module, or see why the warning is triggered and avoid in this code such inconsistent state, but without using introspection if possible.

thomaspaulb:
You're right, perhaps it should in fact be a glue module that would apply the new schedule to the underlying contracts as well. It will be complex to do this in the right way though, I'll return to that later. Thanks

PR #948, opened 2020-12-03
[12.0][ADD] hr_expense_sale_invoice

Khalid-SerpentCS:
HR Expense Sale Invoice
====================
This module adds smart button on invoice for expenses.


hr_expense_sale_invoice/models/account.py
```diff
@@ -0,0 +1,87 @@
...
+from odoo import api, models, _
+
+from odoo.exceptions import UserError
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    @api.multi
+    def _sale_determine_order_line(self):
```
dreispt: Please don't override, extend, ensuring you use `super()`.

dreispt: Is this just a copy of `addons/sale/models/analytic.py`? Why do we need that, AFAIK the link for Expenses to SO work out of the box.

dreispt (changes requested):
I think this should be good after you get rid of the `_sale_determine_order_line()` method.
Do that and I'll do some tests.

hr_expense_sale_invoice/models/account.py
Khalid-SerpentCS: Hi Daniel, i need this method to know for which expense, a invoice line is created.

dreispt: @Khalid-SerpentCS This can't be an override, we need to inherit from the existing method.

hr_expense_sale_invoice/models/hr_expense.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (C) 2020 Open Source Integrators
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class HrExpense(models.Model):
+    _inherit = "hr.expense"
+
+    reference = fields.Char("Vendor")
```
dreispt: Please remove this line.
```suggestion
```

hr_expense_sale_invoice/models/hr_expense.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @api.onchange("analytic_account_id")
+    def _onchange_analytic_account_id(self):
+        for rec in self:
+            if rec.analytic_account_id:
+                rec.sale_order_id = self.env["sale.order"].search(
+                    [("analytic_account_id", "=", rec.analytic_account_id.id)],
+                    limit=1,
+                    order="id desc",
+                )
```
dreispt: Please remove
```suggestion
```

hr_expense_sale_invoice/views/hr_expense_view.xml
```diff
@@ -0,0 +1,14 @@
+<odoo>
+    <record id="hr_expense_form_custom_osi" model="ir.ui.view">
+        <field name="name">hr.expense.custom.form</field>
+        <field name="model">hr.expense</field>
+        <field name="inherit_id" ref="sale_expense.hr_expense_form_view_inherit_sale_expense"/>
+        <field name="arch" type="xml">
+            <field name="sale_order_id" position="replace">
+            </field>
+            <field name="analytic_account_id" position="after">
+                <field name="sale_order_id" attrs="{'required': [('analytic_account_id','!=',False)]}"/>
```
dreispt: Please remove this: we can have non-billable Analytic Accounts or Projects

dreispt (changes requested):

PR #950, opened 2020-12-10
[13.0][IMP] hr_employee_relative: Add smart button

jarroyomorales:
Intstead of adding a new tab to the employee, I think it's better to have them in a separate view with an smart button.

What do you think? @alexey-pelykh @Saran440


Saran440 (commented):
Functional Test. LGTM.
Is it posible to add new menu Relative or else for filter, groupby something?

Maybe, It's not necessary.

Saran440 (approved):

jarroyomorales:
@Saran440 I dont know which filters could be uselful. Maybe check which employees have children or something like that but I see it very custom. If its not a problem I wouldnt add these filters to the OCA module.

etobella (approved):

Issue #955, opened 2021-01-04
V12: Bug when combining the hr_expense_advance_clearing with hr_expense_payment_difference.


hitrosol:
Hi @Saran440, sorry for late response. I try to reproduce the steps as follow:
1. Install hr_expense_advance_clearing
2. Install hr_expense_payment_difference
3. Create new Cash Advance Request Report
4. Submit
5. Approve
6. Post
![image](https://user-images.githubusercontent.com/5949936/103540098-bfe95680-4ecb-11eb-9882-beafd7c85870.png)

7. Register Payment. 
![image](https://user-images.githubusercontent.com/5949936/103540121-cd9edc00-4ecb-11eb-8bc0-963010ec1ba3.png)

8. After Register Payment, the Amount to clear becomes zero. 
![image](https://user-images.githubusercontent.com/5949936/103540144-d7284400-4ecb-11eb-8817-b83dd6f7b02b.png)

9. The Clear Advance & Return Advance are not appear in this stage as well.
9. The result, the cash advance can't be cleared.

The hr_expense_advance_clearing is working good if i don't install the hr_expense_payment_difference
![image](https://user-images.githubusercontent.com/5949936/103540209-f8893000-4ecb-11eb-98ce-50a0049549d1.png)
The Amount to clear and the Clear Advance & Return Advance buttons are there.

Hopefully this is clear enough to give the current situation.

Thanks,

Saran440:
@hitrosol Fixed. Please, Review and test this PR https://github.com/OCA/hr/pull/957
Thank you for your find bug on this module :)

PR #957, opened 2021-01-07
[12.0][FIX] name duplicate with module hr_expense_advance_clearing

Saran440:
rename field `residual` on 2 module
- hr_expense_advance_clearing - rename to `clearing_residual`
- hr_expense_payment_difference - rename to `difference_residual`

reference : https://github.com/OCA/hr/issues/955

### DIFF
--- a/hr_expense_advance_clearing/models/hr_expense.py
+++ b/hr_expense_advance_clearing/models/hr_expense.py
@@ -51,7 +51,7 @@ class HrExpense(models.Model):
                                    'product_emp_advance')
         sheets = self.mapped('sheet_id').filtered('advance_sheet_id')
         sheets_x = sheets.filtered(lambda x: x.advance_sheet_residual <= 0.0)
-        if sheets_x:  # Advance Sheets with no residual left
+        if sheets_x:  # Advance Sheets with no clearing residual left
             raise ValidationError(_('Advance: %s has no amount to clear') %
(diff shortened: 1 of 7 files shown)


jarroyomorales (approved):
LGTM :+1:  The field is not stored so there shouldn't be any problem changing its name.

marcelsavegnago (approved):

pedrobaeza:
Have both fields the same meaning in both modules?

Saran440:
@pedrobaeza yes
In hr_expense_advance_clearing module has `residual` field for clearing advance
https://github.com/OCA/hr/blob/12.0/hr_expense_advance_clearing/models/hr_expense_sheet.py#L29

In hr_expense_payment_difference module has `residual` field for keep amount residual on expense
https://github.com/OCA/hr/blob/12.0/hr_expense_payment_difference/models/hr_expense.py#L10

hitrosol:
Functionally tested locally, looks good 👍

pedrobaeza:
Then I propose to rename the hr_expense_advance_clearing to `clearing_residual` and the other to `difference_residual` and always take the guideline to not putting very generic names to fields that can collapse.

Saran440:
@pedrobaeza fixed. rename 2 module `clearing_residual` and `difference_residual`
Thank you for your advise :)

pedrobaeza (approved):

PR #960, opened 2021-01-25
[12.0][FIX] hr_attendance_report_theoretical_time: Remove search_default_my (context) when hr_attendance_theoretical_action is load from wizard

victoralmau:
Remove search_default_my (context) when hr_attendance_theoretical_action is load from wizard

wizard-error-context

@Tecnativa TT27900

### DIFF
--- a/hr_attendance_report_theoretical_time/wizards/wizard_theoretical_time.py
+++ b/hr_attendance_report_theoretical_time/wizards/wizard_theoretical_time.py
@@ -1,4 +1,5 @@
 # Copyright 2019 Creu Blanca
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models, _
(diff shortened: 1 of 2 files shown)


hr_attendance_report_theoretical_time/wizards/wizard_theoretical_time.py
```diff
@@ -58,4 +60,7 @@ def view_report(self):
             'hr_attendance_theoretical_action'
         ).read()[0]
         action['domain'] = [('employee_id', 'in', self.employee_ids.ids)]
+        context = literal_eval(action['context'])
```
pedrobaeza: Doing this will fail on some cases. You need to remove it other way, or start from a clean context. Is the previous context needed?

victoralmau: IMP previous context is not needed, although it is also true that the default filters (current month or previous month) may be useful.

pedrobaeza: Yes, they are. Then fix the context on code by a direct assignation instead of removing the expected one.

pedrobaeza (approved):

marcelsavegnago (approved):

chienandalu (approved):

PR #961, opened 2021-01-25
[12.0][IMP] hr_attendance_report_theoretical_time: Add department_id in report

victoralmau:
Add `department_id` field in report
example

@Tecnativa TT27900

### DIFF
--- a/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py
+++ b/hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py
@@ -1,4 +1,5 @@
 # Copyright 2017-2019 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models, tools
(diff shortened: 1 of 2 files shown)


hr_attendance_report_theoretical_time/reports/hr_attendance_theoretical_time_report.py
```diff
@@ -66,6 +74,7 @@ def _select_sub1(self):
     def _from_sub1(self):
         return """
             hr_attendance ha
+            LEFT JOIN hr_employee AS he ON ha.id = he.id
```
pedrobaeza: Isn't this already in https://github.com/OCA/hr/pull/961/files#diff-cf0797000a4e7329aafb13b0890ce2b5d09834ded662f2db6f898a580835f3fcR104 ?

victoralmau: It's necessary to work fine sql and view.

pedrobaeza: Then this is under-performant doing 2 times the same join. We shouldn't add this I'm afraid.

pedrobaeza:
OK, I'm now seeing what you wanted to achieve, and I think it's feasible and not very under-performant, so sorry for the noise and it's OK to revert to previous code.

pedrobaeza (approved):

marcelsavegnago (approved):

pedrobaeza:
Trying on runbot, how the same employee can be on 2 departments?

Selección_021

victoralmau:
Thanks for detect error in sql, solved now.

PR #965, opened 2021-02-03
[12.0][FIX] hr_expense_invoice: Set groups to invoice_id field in expense view

victoralmau:
Set groups to `invoice_id` field in expense view

Related to 13.0 https://github.com/OCA/hr-expense/pull/28

@Tecnativa TT28121


pedrobaeza (changes requested):
This doesn't serve, as the error is going to be raised the same if the user has no permissions. You should put the group at view level.

pedrobaeza (approved):

chienandalu (approved):

PR #966, opened 2021-02-05
[12.0][IMP] hr_attendance_report_theoretical_time: Redefine leave dates

victoralmau:
Redefine leave dates.

Please @chienandalu can you review it?

@Tecnativa TT27945

### DIFF
--- a/hr_attendance_report_theoretical_time/models/hr_leave.py
+++ b/hr_attendance_report_theoretical_time/models/hr_leave.py
@@ -1,11 +1,47 @@
 # Copyright 2017-2019 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import models
+from odoo.exceptions import ValidationError
+import logging
+
+_logger = logging.getLogger(__name__)
 
 
 class HrLeave(models.Model):
     _inherit = 'hr.leave'
+
+    def recompute_leave_dates(self):
+        """Recompute leaves if resource_calendar update after create record, although
+        it is strange, it has been possible to change the work schedule"""
+        for item in self:
+            vals = item.read()[0]
+            for key in vals:
+                if isinstance(vals[key], tuple):
+                    vals[key] = vals[key][0]
+            virtual = self.env["hr.leave"].new(vals)
+            virtual._onchange_request_parameters()
+            if (
+                virtual.date_from != item.date_from
+                or virtual.date_to != item.date_to
+            ):
+                try:
+                    item.write({
+                        "date_from": virtual.date_from,
+                        "date_to": virtual.date_to,
+                    })
+                    # change resource_calendar_leaves
+                    resource = self.env['resource.calendar.leaves'].search([
+                        ('holiday_id', '=', item.id)
+                    ])
+                    if resource:
+                        resource.write({
+                            "date_from": item.date_from,
+                            "date_to": item.date_to,
+                        })
+                except ValidationError:
+                    _logger.info("Skip: Some leaves overlap on same day")
 
     def _create_resource_leave(self):
         """On leave creation, trigger the recomputation of the involved
--- a/hr_attendance_report_theoretical_time/tests/test_hr_attendance_report_theoretical_time.py
+++ b/hr_attendance_report_theoretical_time/tests/test_hr_attendance_report_theoretical_time.py
@@ -22,12 +22,12 @@ class TestHrAttendanceReportTheoreticalTime(common.SavepointCase):
                     'name': 'Attendance',
                     'dayofweek': str(day),
                     'hour_from': '08',
-                    'hour_to': '12',
+                    'hour_to': '13',
                 }),
                 (0, 0, {
                     'name': 'Attendance',
                     'dayofweek': str(day),
-                    'hour_from': '14',
+                    'hour_from': '15',
                     'hour_to': '18',
                 }),
             ]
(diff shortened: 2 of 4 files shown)


pedrobaeza:
The method should go to execute on `hr.leave` the same methods as if you change the dates for triggering the recompute of the beginning and ending of the leave.

victoralmau:
I have changed it to use `_onchange_request_parameters() `
https://github.com/odoo/odoo/blob/12.0/addons/hr_holidays/models/hr_leave.py#L274-L319

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -7,6 +8,19 @@
 class HrLeave(models.Model):
     _inherit = 'hr.leave'
 
+    def recompute_leave_dates(self):
+        """Recompute leaves if resource_calendar update after create record, although
+        it is strange, it has been possible to change the work schedule"""
+        for item in self.filtered(lambda x: x.state == 'validate'):
+            item._onchange_request_parameters()
+            # change resource_calendar_leaves (remove + create)
+            item._remove_resource_leave()
```
pedrobaeza: This seems very invasive. Is it really needed? And you should only do it for those that are not correct. You should check the initial values and compare if there's a change.

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -1,12 +1,38 @@
...
-from odoo import models
+from odoo import models, fields
 
 
 class HrLeave(models.Model):
     _inherit = 'hr.leave'
 
+    def _prepare_resource_leave_values(self):
+        return {
+            'date_from': fields.Datetime.to_string(
```
pedrobaeza: Why do you need this?

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -1,12 +1,38 @@
...
+            ),
+            'date_to': fields.Datetime.to_string(
+                fields.Datetime.from_string(self.date_to)
+            ),
+        }
+
+    def recompute_leave_dates(self):
+        """Recompute leaves if resource_calendar update after create record, although
+        it is strange, it has been possible to change the work schedule"""
+        for item in self.filtered(lambda x: x.state == 'validate'):
```
pedrobaeza: You shouldn't filter on anything. All are applicable to be changed.

pedrobaeza (changes requested):
I still see that this doesn't get the desired effect, as you are still writing on all the possible holidays. Just those that change its starting date or ending date after applying the new calendar must be written.

hr_attendance_report_theoretical_time/models/hr_leave.py
victoralmau: It's necessary because don't exist any function to prepare dict.
Is necessary to update resource.leave record fields: `date_from` or `date_to` (datetime fields) if leave had changed and our dates (hours).

pedrobaeza: I think we are not thinking in the same things. Better to meet for joining efforts.

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -1,12 +1,48 @@
...
+            ):
+                item._onchange_request_parameters()
+                # change resource_calendar_leaves
+                resource = self.env['resource.calendar.leaves'].search([
+                    ('holiday_id', '=', item.id)
+                ])
+                if resource:
+                    resource.write(item._prepare_resource_leave_values())
+                # update meeting_id
+                if item.meeting_id:
```
pedrobaeza: Don't do this. This is probably in the past, so this doesn't matter and it has nothing to do with theoretical report.

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -1,12 +1,48 @@
...
+            for key in vals:
+                if isinstance(vals[key], tuple):
+                    vals[key] = vals[key][0]
+            virtual = self.env["hr.leave"].new(vals)
+            virtual._onchange_request_parameters()
+            if (
+                virtual.date_from != item.date_from
+                or virtual.date_to != item.date_to
+            ):
+                item._onchange_request_parameters()
```
pedrobaeza: Don't do that, but just write the changed fields.

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -1,12 +1,48 @@
...
+from odoo import models, fields
 
 
 class HrLeave(models.Model):
     _inherit = 'hr.leave'
 
+    def _prepare_resource_leave_values(self):
+        return {
+            'date_from': fields.Datetime.to_string(
+                fields.Datetime.from_string(self.date_from)
```
pedrobaeza: No need of this conversions. Odoo accepts datetime objects in the dictionary.

hr_attendance_report_theoretical_time/models/hr_leave.py
```diff
@@ -1,12 +1,48 @@
...
+                virtual.date_from != item.date_from
+                or virtual.date_to != item.date_to
+            ):
+                item._onchange_request_parameters()
+                # change resource_calendar_leaves
+                resource = self.env['resource.calendar.leaves'].search([
+                    ('holiday_id', '=', item.id)
+                ])
+                if resource:
+                    resource.write(item._prepare_resource_leave_values())
```
pedrobaeza: Isn't this synchronized with the leave?

pedrobaeza (changes requested):

hr_attendance_report_theoretical_time/models/hr_leave.py
victoralmau: It's necessary because when confirm auto-create `resource.calendar.leaves` record.

pedrobaeza:
I don't think this should land the module, as this has a very high performance cost for very edge cases that comes from the manual manipulation, and thus, they should be fixed manually as well. Closing it.

PR #972, opened 2021-02-17
[12.0][IMP] Improved the Multi Submit expense report with out attachment.

Murtaza-OSI:


hr_expense_receipt_require/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "HR Expense Receipt Require",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.0.1",
```
dreispt, line 6: For OCA modules don't bump versions.
that will be done by the merge bot.

hr_expense_receipt_require/models/hr_expense.py
```diff
@@ -9,6 +9,10 @@ class HrExpense(models.Model):
 
     @api.multi
     def action_submit_expenses(self):
+        context = self._context or {}
+        if context.get('select_multi_records') and\
```
dreispt, line 13: This is smart, but I'm not sure we need it.
It can't think of  a bad case if we don't use this context key.
Thoughts?

dreispt (commented):
Can you check my comments?
Please also fix the Travis build.
Thanks

PR #971, opened 2021-02-16
[FIX] hr_employee_firstname: Avoid access error on employee update

luistorresm:
``sudo`` is used to avoid access error when verify that module is installed.

Avoid error in the next case:
1. Create a new user without settings access (HR User)
2. Login with the user in 1, and try to edit an employee

PR #973, opened 2021-02-17
[FIX] hr_employee_firstname: Avoid access error on employee update

luistorresm:
``sudo`` is used to avoid access error when verify that module is installed.

Avoid error in the next case:
1. Create a new user without settings access (HR User)
2. Login with the user in 1, and try to edit an employee

Related PR: https://github.com/OCA/hr/pull/971


luistorresm:
Hello @moylop260 

This is a forward port from v13, could you merge it if you agree, please?

Regards

moylop260:
Merging since that it is a FW from merged:
 - https://github.com/OCA/hr/pull/971

And it is fixing a valid issue

PR #977, opened 2021-02-22
[MIG] hr_job_category: Migration to 14.0

square-ava:
Module is working fine but tests were failing because Odoo added an onchange method on the employee that is on the contract, since we were writing an employee that has no job -> the job was lost and tests were failing, added a job in test to fix it.


etobella:
You are not keeping commit history. Please review the official documentation

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

PR #978, opened 2021-02-26
[14.0][MIG] hr_contract_reference: Migration to 14.0

tobiaszehntner:
- Removed a view that sets the field to `readonly`, since the python field declaration does that already
- Avoid assigning a new value to `self.employee` in the test


Saran440 (approved):
Code Review. No major change :+1:

lfreeke (approved):
Functional test :+1:

PR #980, opened 2021-03-04
Translated using Poedit(Chinese)

bigfoot88:


pedrobaeza:
Please use Weblate instead:

https://odoo-community.org/page/translate

PR #981, opened 2021-03-05
[13.0][IMP] hr_org_chart_overview: add change orientation option 

irfani:
add change orientation option (horizontal/vertical)
image

image


hr_org_chart_overview/static/src/js/hr_org_chart_overview.js
```diff
@@ -249,6 +254,22 @@ odoo.define("hr_org_chart_overview", function(require) {
                 this._filterNodes(value);
             }
         },
+
+        _onChangeOrientation: function(ev) {
+            ev.preventDefault();
+            console.log(this.direction);
```
etobella: I think this is not needed :wink:

irfani: oops ok, i'll remove it 😄

PR #991, opened 2021-05-10
[12.0][FIX] hr_holidays_validity_date: Check for `restrict_dates` when validating an allocation

ypapouin:
In the initial behavior, the `_check_leave_type_validity` is overridden only for the `hr.leave` model but the same constraint exist in the built-in implementation of the `hr.leave.allocation` model.

This commit ensures consistent operation on both models.


albariera (approved):

olgamarcocb (approved):
LGTM 👍

PR #690, opened 2019-10-01
[MIG] hr_employee_display_own_info: Migration to 13.0

espo-tony:

PR #997, opened 2021-05-25
[MIG] hr_employee_display_own_info: Migration to 13.0

albariera:
Superseded #690 as the author seems inactive


albariera:
After reviewing this PR with @etobella and @olgamarcocb, we have seen that in odoo 13.0 we have a menu 'Employee Directory' apart from 'Employees'. At 'Employees' we can see the employee's private information but at 'Employee Directory' no. A common user has only access to 'Employee Directory', so it will not see the other employee's information. As a consequence, this module is no longer necessary, so I close this pull request as it has no sense merging it to 13.0.

PR #1002, opened 2021-06-14
[IMP]: add unit in form selection widget

maisim:
[FIX] hr_holidays_credit: plurals management in form selection widget
[IMP] hr_holidays_credit: Update french translation
[IMO] hr_holidays_credit: add messages on tests

### DIFF
--- a/hr_holidays_credit/models/hr_leave_type.py
+++ b/hr_holidays_credit/models/hr_leave_type.py
@@ -6,70 +6,95 @@ from odoo.tools.float_utils import float_round
 
 
 class HrLeaveType(models.Model):
-    _inherit = 'hr.leave.type'
+    _inherit = "hr.leave.type"
 
     allow_credit = fields.Boolean(
-        string='Allow Credit',
+        string="Allow Credit",
         help=(
-            'If set to true, employees would be able to make requests for this'
-            ' leave type even if allocated amount is insufficient.'
+            "If set to true, employees would be able to make requests for this"
+            " leave type even if allocated amount is insufficient."
         ),
     )
     creditable_employee_ids = fields.Many2many(
-        string='Creditable Employees',
-        comodel_name='hr.employee',
-        help='If set, limits credit allowance to specified employees',
+        string="Creditable Employees",
+        comodel_name="hr.employee",
+        help="If set, limits credit allowance to specified employees",
     )
     creditable_employee_category_ids = fields.Many2many(
-        string='Creditable Employee Tags',
-        comodel_name='hr.employee.category',
+        string="Creditable Employee Tags",
+        comodel_name="hr.employee.category",
         help=(
-            'If set, limits credit allowance to employees with at least one of'
-            ' specified tags'
+            "If set, limits credit allowance to employees with at least one of"
+            " specified tags"
         ),
     )
     creditable_department_ids = fields.Many2many(
-        string='Creditable Departments',
-        comodel_name='hr.department',
+        string="Creditable Departments",
+        comodel_name="hr.department",
         help=(
-            'If set, limits credit allowance to employees of specified'
-            ' departments'
+            "If set, limits credit allowance to employees of specified departments"
         ),
     )
 
     @api.multi
     def name_get(self):
-        context_employee_id = self._context.get('employee_id')
+        context_employee_id = self._context.get("employee_id")
 
         res = []
         for record in self:
             record_name = record.name
 
             extra = None
-            if record.allocation_type != 'no' and context_employee_id:
-                if record.virtual_remaining_leaves >= 0:
-                    if record.allow_credit:
-                        extra = _('%g available + credit')
-                    else:
-                        extra = _('%g available')
-                    extra = extra % (
-                        float_round(
-                            record.virtual_remaining_leaves,
-                            precision_digits=2
-                        ) or 0.0,
-                    )
-                elif record.allow_credit:
-                    extra = _('%g used in credit') % (
-                        float_round(
-                            -record.virtual_remaining_leaves,
-                            precision_digits=2
-                        ) or 0.0,
-                    )
+            if record.allocation_type != "no" and context_employee_id:
+                amount = (
+                    float_round(record.virtual_remaining_leaves,
+                                precision_digits=2)
+                    or 0.0
+                )
+                if amount >= 0:
+                    if record.request_unit == "day":
+                        if not record.allow_credit:
+                            if abs(amount) <= 1:
+                                extra = _("%g day available") % amount
+                            else:
+                                extra = _("%g days available") % amount
+                        else:
+                            if abs(amount) <= 1:
+                                extra = _("%g day available + credit") % amount
+                            else:
+                                extra = _("%g days available + credit") % amount
+
+                    elif record.request_unit == "hour":
+                        if not record.allow_credit:
+                            if abs(amount) <= 1:
+                                extra = _("%g hour available") % amount
+                            else:
+                                extra = _("%g hours available") % amount
+                        else:
+                            if abs(amount) <= 1:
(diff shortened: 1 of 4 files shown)


hr_holidays_credit/models/hr_leave_type.py
```diff
@@ -47,27 +47,28 @@ def name_get(self):
 
             extra = None
             if record.allocation_type != 'no' and context_employee_id:
-                if record.virtual_remaining_leaves >= 0:
+                amount = float_round(record.virtual_remaining_leaves, precision_digits=2) or 0.0
+                if(amount >= 0):
```
alexey-pelykh: `(` and `)` are not needed at all

hr_holidays_credit/models/hr_leave_type.py
```diff
@@ -47,27 +47,28 @@ def name_get(self):
...
+                    if record.request_unit == 'day':
+                        extra = _('%g day available') % amount if abs(amount) <= 1 else _('%g days available') % amount
+                    elif record.request_unit == 'hour':
+                        extra = _('%g hour available') % amount if abs(amount) <= 1 else _('%g hours available') % amount
+                    else:
+                        extra = _('%g %s available') % (amount, record.request_unit)
+
                     if record.allow_credit:
-                        extra = _('%g available + credit')
+                            extra += " " + _('+ credit')
```
alexey-pelykh: Indent went very bad

alexey-pelykh (changes requested):

maisim:
Thanks for your review @alexey-pelykh

hr_holidays_credit/tests/test_hr_holidays_credit.py
```diff
@@ -148,7 +148,8 @@ def test_6(self):
         name = leave_type.with_context(
             employee_id=employee.id,
         ).name_get()[0][1]
-        self.assertTrue('available + credit' in name)
+        self.assertIn('available', name, '%s not found in %s' % ('available', name))
```
alexey-pelykh: What's the point of formatting the error message?

hr_holidays_credit/models/hr_leave_type.py
```diff
@@ -47,27 +47,28 @@ def name_get(self):
...
-                        float_round(
-                            -record.virtual_remaining_leaves,
-                            precision_digits=2
-                        ) or 0.0,
-                    )
+                        extra = _('%g %s used in credit') % (amount, record.request_unit)
 
             if extra:
-                record_name = _('%(name)s (%(extra)s)') % {
+                record_name = '%(name)s (%(extra)s)' % {
```
alexey-pelykh: Why did you remove translation here? RTL languages kind-a have that backwards

hr_holidays_credit/tests/test_hr_holidays_credit.py
maisim: To know why test don't pass, So, I add messages to display compared values. I can remove it know but that's here... Let me know.
It was because my odoo test db default language was not English so name values was translated in the db default language (add -l or --load-language to command line does not help) I add to create a new one with english as default language

hr_holidays_credit/models/hr_leave_type.py
maisim: To save white bears :angel:  
To me it's useless because it mean you have strings like : "Legal Leaves 2021 (20 days available)" in your po file, wich is very unlikely

alexey-pelykh: That's not how translation works in this case. Template is getting localized, not formatted string

maisim: I don't doubt what you're saying but I would like to understand. (I probably miss something)
In this case, for example,at line 70:
name is string containing "Legal Leaves 2021"
extra is string containing " (20 days available)"

record_name is string = name + " " + extra   = a string containing  "Legal Leaves 2021 (20 days available)"

So I don't understand how it can work in another way than having the full text "Legal Leaves 2021 (20 days available)" in the po file (or in the db)

alexey-pelykh: Pay attention to what's inside the `_()` and compare this

```
_('%(name)s (%(extra)s)') % {
    'name': 'A',
    'extra': 'B',
}
```

and 

```
_('%(name)s (%(extra)s)' % {
    'name': 'A',
    'extra': 'B',
})
```

hr_holidays_credit/tests/test_hr_holidays_credit.py
alexey-pelykh: I'm not sure if that's the correct way, to include localization in testing. I'd refer this to other @OCA/human-resources-maintainers

maisim: The goal is not to test localization, it was to undesrtand why test failed has I had no tested values in logs

alexey-pelykh: Runbot tests in en_US, so probably does not matter

hr_holidays_credit/models/hr_leave_type.py
maisim: Ok, it allow to move vars in translations 
But in our case I think vars are not enough splitted to this use case.
Anyway I will undo this change.
Thanks for your time

hr_holidays_credit/tests/test_hr_holidays_credit.py
maisim: Yes but when I ran tests locally I can't see why the tests failed ( tests results didn't print variables content)
I then add messages to the tests which allowed me to see that the messages were translated (the reason why the failed tests) and to run the tests again on a basis with English as the default language.

maisim: I removed custom messages in the last commit in the same time a had updated the tests

maisim:
Can I merge commits (rebase) on my side or this repo allow squash merging ?

alexey-pelykh:
You have to tidy up commit history, no squash merging

hr_holidays_credit/__manifest__.py
```diff
@@ -4,7 +4,7 @@
 
 {
     'name': 'Leave Credit',
-    'version': '12.0.1.0.0',
+    'version': '12.0.2.0.0',
```
alexey-pelykh: This is also not needed, auto-bump is in place, and this is not a major change.

hr_holidays_credit/models/hr_leave_type.py
```diff
@@ -6,70 +6,86 @@
...
     )
     creditable_department_ids = fields.Many2many(
-        string='Creditable Departments',
-        comodel_name='hr.department',
+        string="Creditable Departments",
+        comodel_name="hr.department",
         help=(
-            'If set, limits credit allowance to employees of specified'
-            ' departments'
+            "If set, limits credit allowance to employees of specified" " departments"
```
alexey-pelykh: Clean up `" "` after autoformatting

hr_holidays_credit/models/hr_leave_type.py
```diff
@@ -6,70 +6,86 @@
...
-                        ) or 0.0,
-                    )
-                elif record.allow_credit:
-                    extra = _('%g used in credit') % (
-                        float_round(
-                            -record.virtual_remaining_leaves,
-                            precision_digits=2
-                        ) or 0.0,
-                    )
+                        extra += " " + _("+ credit")
```
alexey-pelykh: Don't do this, it's incorrect for i18n to "construct" sentences. You have to localize either entire message or major part of it.

maisim:
Hi @alexey-pelykh,
Can you have a look on this ?
Thank you in advance

PR #1003, opened 2021-06-15
[FIX] hr_employee_calendar_planning: resource calendar mismatch

chienandalu:
When we create the employee the linked resource doesn't update properly
with the autogenerated calendar and it remains with default company
calendar. If we simply do a setting on itself it links correctly again.

cc @Tecnativa TT30439

please review @pedrobaeza

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -50,8 +50,8 @@ class HrEmployee(models.Model):
         two_weeks = bool(
             self.calendar_ids.mapped("calendar_id").filtered("two_weeks_calendar")
         )
-        if not self.resource_calendar_id or self.resource_calendar_id.active:
-            self.resource_calendar_id = (
+        if not self.resource_id.calendar_id or self.resource_id.calendar_id.active:
+            self.resource_id.calendar_id = (
                 self.env["resource.calendar"]
                 .create(
                     {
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Can we avoid the re-writing and pass on `vals_list` the proper value? Or even better, just adjust the proper one in the calling method.

pedrobaeza:
What about the average hours number adjustment?

chienandalu:
Changes done

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -64,6 +64,11 @@ def _regenerate_calendar(self):
                 )
                 .id
             )
+            # We need to ensure that the related gets the proper autogenerated
+            # calendar when we create the employees. Otherwise it remains as the company
+            # default one
+            if self.resource_id.calendar_id != self.resource_calendar_id:
+                self.resource_id.calendar_id = self.resource_calendar_id
```
pedrobaeza: Now that I see the code, should we assign `self.resource_id.calendar_id` in the `create` command instead?

chienandalu: Yes, it was what I was doing in the first commit although inderectly triggering the relation setting

pedrobaeza: I meant here:

https://github.com/OCA/hr/pull/1003/files#diff-edb2d081d492df756c64e4a9cf821b691b64c7f3cf8b94b3947f151de9810a2dR54

Chat me if further clarification.

chienandalu: Understood and changed!

pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
Please fw-port it to v13

PR #1007, opened 2021-07-13
[13.0][FIX] hr_employee_calendar_planning: Prevent disable resource.calendar when is related to some employee

victoralmau:
Prevent disable `resource.calendar` when is related to some employee.

It's need to change in 14.0 too.

@Tecnativa TT31007

### DIFF
--- a/hr_employee_calendar_planning/models/resource_calendar.py
+++ b/hr_employee_calendar_planning/models/resource_calendar.py
@@ -1,13 +1,27 @@
 # Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ResourceCalendar(models.Model):
     _inherit = "resource.calendar"
 
     active = fields.Boolean(default=True)
+
+    @api.constrains("active")
+    def _check_active(self):
+        for item in self:
+            total_items = self.env["hr.employee.calendar"].search_count(
+                [("calendar_id", "=", item.id)]
+            )
+            if total_items:
+                raise ValidationError(
+                    _("%s is used in %s employee(s). You should change them first.")
+                    % (item.name, total_items)
+                )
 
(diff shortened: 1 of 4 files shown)


hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -1,14 +1,30 @@
...
+    def _check_active(self):
+        for item in self:
+            total_items = self.env["hr.employee.calendar"].search_count(
+                [("calendar_id", "=", item.id)]
+            )
+            if total_items > 0:
+                raise ValidationError(
+                    _(
+                        "%s is related with some employee in calendar planing,"
+                        "you should change it before." % item.name
```
joao-p-marques:
```suggestion
                        "%s is related to %s employee(s) in calendar planing,"
                        " you should change it first." % (item.name, total_items)
```

joao-p-marques (commented):
Code and functional review :+1:

Just a little suggestion on the error message

joao-p-marques (approved):

hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -1,14 +1,30 @@
...
 
     active = fields.Boolean(default=True)
 
+    @api.constrains("active")
+    def _check_active(self):
+        for item in self:
+            total_items = self.env["hr.employee.calendar"].search_count(
+                [("calendar_id", "=", item.id)]
+            )
+            if total_items > 0:
```
pedrobaeza:
```suggestion
            if total_items:
```

hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -1,14 +1,30 @@
...
+                raise ValidationError(
+                    _(
+                        "%s is related to %s employee(s) in calendar planing,"
+                        " you should change it first." % (item.name, total_items)
+                    )
```
pedrobaeza: Never include the placeholder elements inside the call to translate

```suggestion
                    _(
                        "%s is used in %s employee(s). You should change them first."
                    ) % (item.name, total_items)
```

pedrobaeza (changes requested):

pedrobaeza (approved):

PR #1009, opened 2021-07-15
[14.0][MIG] hr_employee_id

mtelahun:
Migration to 14.0


albariera:
Could you separate the pre-commit and migration into two different commits? Even if it is just a simple migration it makes it easier to compare versions.

mtelahun:
Sorry, I'm not sure I understand what you mean. All my changes are in 
the last three commits.

All the pre-commit stuff is in:

40c47d9  [IMP] hr_employee_id: black, isort, prettier

The migration changes are in:

5a24de1 Fix hr_employee_id: domain on non-relational field makes no sense

2e8e645 hr_employee_id: Add additional tests


If I'm misunderstanding something please let me know.

albariera:
I think it is better to fuse the last two commits into a single one called "[MIG] hr_employee_id : Migration to 14.0", as the OCA wiki recommends: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#technical-method-to-migrate-a-module-from-120-to-130-branch
It will make it easier to find the migration commit in the future :smile:

gurneyalex (approved):

hr_employee_id/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,71 @@
...
+                                <div class="text-muted col-lg-8">
+                                    Select &amp; configure default employee identifier generation scheme
+                                </div>
+                            </div>
+                            <div class="content-group">
+                                <div class="mt16">
+                                    <field
+                                        name="employee_id_gen_method"
+                                        class="o_light_label"
+                                        context="{'default_company_id': company_id}"
```
i-vyshnevska, line 35:
```suggestion
```
context also is redundant

i-vyshnevska (approved):
LG, 
also you can squash translation commits to reduce history for instance from e50aa7b to 7d35549

PR #1010, opened 2021-07-16
[FIX] hr_employee_relative: change attribute replace by invisible

albariera:
While inheriting the view_employee_form I found the problem that I could not use the field 'children' because it was replaced. To solve this problem, the 'replace' attribute of some fields has been changed to 'invisible'. The 'Dependant' group has also been made invisible.

### DIFF
--- a/hr_employee_relative/views/hr_employee.xml
+++ b/hr_employee_relative/views/hr_employee.xml
@@ -9,9 +9,9 @@
         <field name="model">hr.employee</field>
         <field name="inherit_id" ref="hr.view_employee_form" />
         <field name="arch" type="xml">
-            <field name="spouse_complete_name" position="replace" />
-            <field name="spouse_birthdate" position="replace" />
-            <field name="children" position="replace" />
+            <xpath expr="//field[@name='children']/.." position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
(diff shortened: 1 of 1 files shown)


hr_employee_relative/views/hr_employee.xml
```diff
@@ -9,9 +9,18 @@
...
+            <field name="spouse_complete_name" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </field>
+            <field name="spouse_birthdate" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </field>
+            <field name="children" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </field>
+            <xpath expr="//field[@name='children']/.." position="attributes">
```
etobella, line 12: If you hide the group it is necessary to hide all the fields to?

etobella (approved):
LGTM Not blocking, just waiting other opinions about it

hr_employee_relative/views/hr_employee.xml
albariera: You are right! Thank you :smile:

albariera:
What do you think? @Saran440

Saran440 (approved):
LGTM :+1:

PR #1012, opened 2021-07-27
[13.0][FIX] hr_employee_calendar_planning: Add check_company to calendar_id field in to show only allowed records according to company

victoralmau:
Add `check_company` to `calendar_id` field in to show only allowed records according to company.
Add constraint to prevent change avoid changing the company when the calendar is linked to an employee of another company.

@Tecnativa TT31201

### DIFF
--- a/hr_employee_calendar_planning/models/resource_calendar.py
+++ b/hr_employee_calendar_planning/models/resource_calendar.py
@@ -23,6 +23,21 @@ class ResourceCalendar(models.Model):
                     % (item.name, total_items)
                 )
 
+    @api.constrains("company_id")
+    def _check_company_id(self):
+        for item in self.filtered("company_id"):
+            total_items = self.env["hr.employee.calendar"].search_count(
+                [
+                    ("calendar_id.company_id", "=", item.company_id.id),
+                    ("employee_id.company_id", "!=", item.company_id.id),
+                ]
+            )
+            if total_items:
+                raise ValidationError(
+                    _("%s is used in %s employee(s) related to another company.")
+                    % (item.name, total_items)
+                )
+
     def write(self, vals):
         res = super(ResourceCalendar, self).write(vals)
         if "attendance_ids" in vals:
--- a/hr_employee_calendar_planning/views/hr_employee_views.xml
+++ b/hr_employee_calendar_planning/views/hr_employee_views.xml
@@ -10,8 +10,12 @@
                 <attribute name="invisible">1</attribute>
             </field>
             <field name="resource_calendar_id" position="after">
-                <field name="calendar_ids">
+                <field
+                    name="calendar_ids"
+                    context="{'default_company_id': parent.company_id}"
+                >
                     <tree editable="top">
+                        <field name="company_id" invisible="1" />
                         <field name="date_start" />
                         <field name="date_end" />
                         <field name="calendar_id" />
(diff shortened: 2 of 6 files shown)


hr_employee_calendar_planning/views/hr_employee_views.xml
```diff
@@ -14,7 +14,10 @@
                     <tree editable="top">
                         <field name="date_start" />
                         <field name="date_end" />
-                        <field name="calendar_id" />
+                        <field
+                            name="calendar_id"
+                            domain="['|', ('company_id', '=', False), ('company_id', '=', parent.company_id)]"
```
pedrobaeza: Why not doing it at field level? And what about `check_company=True`?

hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -23,6 +23,21 @@ def _check_active(self):
                     % (item.name, total_items)
                 )
 
+    @api.constrains("company_id")
+    def _check_company_id(self):
+        for item in self.filtered("company_id"):
```
pedrobaeza, line 28: Isn't this done automatically through `check_company=True` attribute on the field?

hr_employee_calendar_planning/views/hr_employee_views.xml
victoralmau: If we add check_company=True to `calendar_id` field we will need to add `company_id` field (and add these field in view as invisible) and set as related (employee_id.company_id) but don't work because is not set before record creation.

pedrobaeza: No problem on that, as AFAIK, the constraint is checked after that assignation. I think we should stick maximum to the standard multi-companies mechanisms.

hr_employee_calendar_planning/models/resource_calendar.py
victoralmau: Do you mean that it is necessary to add `check_company=True` to the `company_id` field of the `resource.calendar` model?

hr_employee_calendar_planning/views/hr_employee_views.xml
victoralmau: Ok, but if the `company_id` field is not set, no calendar will be show (because all records have some company defined).

hr_employee_calendar_planning/models/resource_calendar.py
victoralmau: `_check_company_auto=True` is not enough to prevent `company_id` change in resource calendars.

pedrobaeza (approved):

hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -23,6 +23,21 @@ def _check_active(self):
...
+            total_items = self.env["hr.employee.calendar"].search_count(
+                [
+                    ("calendar_id.company_id", "=", item.company_id.id),
+                    ("employee_id.company_id", "!=", item.company_id.id),
+                ]
+            )
+            if total_items:
+                raise ValidationError(
+                    _("%s is used in %s employee(s) related to other company.")
```
joao-p-marques: I think this would be a bit clearer:

```suggestion
                    _("%s is used in %s employee(s) related to another company.")
```

joao-p-marques (approved):
Just a small comment. However, not blocking...

LGTM :+1:

PR #1013, opened 2021-07-28
[14.0][FIX] hr_employee_calendar_planning: Add check_company to calendar_id field in to show only allowed records according to company

victoralmau:
FWP from 13.0: https://github.com/OCA/hr/pull/1012

Add `check_company` to `calendar_id` field in to show only allowed records according to company.
Add constraint to prevent change avoid changing the company when the calendar is linked to an employee of another company.

@Tecnativa TT31201


pedrobaeza (approved):

pedrobaeza:
Let's fast-track it as direct fw-port:

/ocabot merge minor

PR #1017, opened 2021-08-10
[13.0][FIX] hr_employee_document: Add security rule to allow user without employee groups can access to attachments related to own employee

victoralmau:
Add security rule to allow user without employee groups can access to attachments related to own employee.

@Tecnativa TT31359

### DIFF
--- a/hr_employee_document/README.rst
+++ b/hr_employee_document/README.rst
@@ -25,12 +25,21 @@ HR Employee Document
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module allows to attach documents to the employee profile.
+This module displays a button on the employee's profile to view linked attachments, both
+for HR officers/managers and the own employee, so you can use it as employee document storage.
 
 **Table of contents**
 
 .. contents::
    :local:
+
+Known issues / Roadmap
+======================
+
+This module adds read permission to the employee model; the rest of the model data will
+be available to the employee.
+If another module grants permissions to the employee model (Pre-officer for example),
+I will need to create a rule to extend the basic permissions.
 
 Bug Tracker
 ===========
--- a/hr_employee_document/security/security.xml
+++ b/hr_employee_document/security/security.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="access_hr_employee_system_user_override" model="ir.model.access">
+        <field name="name">hr.employee system user override</field>
+        <field name="model_id" ref="hr.model_hr_employee" />
+        <field name="group_id" ref="base.group_user" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+    </record>
+    <record id="rule_hr_employee_user" model="ir.rule">
+        <field name="name">User can only read their own employees</field>
+        <field name="model_id" ref="hr.model_hr_employee" />
+        <field name="groups" eval="[(4, ref('base.group_user'))]" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+        <field name="domain_force">[('id','in', user.employee_ids.ids)]</field>
+    </record>
+    <record id="rule_hr_employee_hr_user" model="ir.rule">
+        <field name="name">Hr user can access to all records</field>
+        <field name="model_id" ref="hr.model_hr_employee" />
+        <field name="groups" eval="[(4, ref('hr.group_hr_user'))]" />
+        <field name="perm_read" eval="1" />
+        <field name="perm_create" eval="1" />
+        <field name="perm_write" eval="1" />
+        <field name="perm_unlink" eval="1" />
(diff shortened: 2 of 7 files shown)


hr_employee_document/security/security.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).-->
+<odoo>
+    <record id="access_hr_employee_system_user_override" model="ir.model.access">
+        <field name="name">hr.employee system user override</field>
+        <field name="model_id" ref="hr.model_hr_employee" />
+        <field name="group_id" ref="base.group_user" />
+        <field name="perm_read" eval="1" />
```
chienandalu, line 7: How do model access rule prevail when they have the same group? Could `access_hr_employee_system_user` interfere with these one?

victoralmau: With this change it works as we wish by applying the rules correctly as the tests show. It is true that it is something "strange" and I don't know if it can have side effects. What do you think @pedrobaeza ?

pedrobaeza:
This is not correct. You should inherit the underlying method (similar to `dms`) when accessing attachments, and check if they belong to their employee.

hr_employee_document/models/hr_employee.py
```diff
@@ -35,3 +35,18 @@ def action_get_attachment_tree_view(self):
...
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    @api.model
+    def check_access_rights(self, operation, raise_exception=True):
+        """We try to override access_hr_employee_system_user access right
+        only to read operation because rule_hr_employee_user rule allow access
+        only own employees"""
+        if operation == "read" and not self.env.user.has_group("hr.group_hr_user"):
```
pedrobaeza: Why are you still giving access to employee? Permissions should be on documents, not on employee.

victoralmau:
> This is not correct. You should inherit the underlying method (similar to `dms`) when accessing attachments, and check if they belong to their employee.

With this idea I have made some changes that may be a better option (`dms` is much more complex and so are the attachments, because to access a record you must be able to access the linked record, `hr.employee`).

hr_employee_document/models/hr_employee.py
victoralmau: I have applied the change here, because to access a document that has a related field (with employee in this case) you must have access to that model and the linked record.

pedrobaeza: But why Odoo doesn't allow to read its own employee? This is what I don't understand, as all employees are available to be seen.

victoralmau: Odoo allows any user to access `hr.employee.public` but not `hr.employee`, and attachments are linked to `hr.employee`

hr_employee_document/README.rst
```diff
@@ -25,13 +25,21 @@ HR Employee Document
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module allows to attach documents to the employee profile.
+This module displays a button on the employee's profile to view linked attachments.
```
pedrobaeza:
```suggestion
This module displays a button on the employee's profile to view linked attachments, both
for HR officers/managers and the own employee, so you can use it as employee document storage.
```

Add in usage that you can attach the document as HR office/manager or click on the smart-button and create a new attachment.

pedrobaeza (approved):
We have seen that other solutions attacking ir.attachment will impact in the performance, as they are used everywhere in the system, so let's use this one as the less bad. ACLs are additive, so we can define an extra one for giving read permission without conflicting.

chienandalu (approved):

hr_employee_document/security/security.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+    </record>
+    <record id="rule_hr_employee_user" model="ir.rule">
+        <field name="name">User can only read their own employees</field>
+        <field name="model_id" ref="hr.model_hr_employee" />
+        <field name="groups" eval="[(4, ref('base.group_user'))]" />
+        <field name="global" eval="True" />
```
gurneyalex:
```suggestion
```

gurneyalex: `global` is computed, and it will compute to `False` because you have `groups` set.

hr_employee_document/security/security.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="perm_create" eval="0" />
+        <field name="perm_write" eval="0" />
+        <field name="perm_unlink" eval="0" />
+        <field name="domain_force">[('id','in', user.employee_ids.ids)]</field>
+    </record>
+    <record id="rule_hr_employee_hr_user" model="ir.rule">
+        <field name="name">Hr user can access to all records</field>
+        <field name="model_id" ref="hr.model_hr_employee" />
+        <field name="groups" eval="[(4, ref('hr.group_hr_user'))]" />
+        <field name="global" eval="True" />
```
gurneyalex:
```suggestion
```

pedrobaeza:
Thanks for the review, @gurneyalex 

Let's merge it once it has been changed:

/ocabot merge major

PR #1020, opened 2021-08-23
[12.0][FIX] hr_expense_invoice: Update prices correctly after invoice creation from expense with some taxes defined

victoralmau:
Steps to reproduce:

- Create an expense with defined taxes.
- Create an expense sheet.
- Validate the expense sheet.
- Create an invoice from the expense with taxes.
- The amounts (unit amount and total amount) are not modified when the invoice is created.

@Tecnativa TT31447

### DIFF
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -53,7 +53,7 @@ class HrExpense(models.Model):
             {
                 "invoice_id": invoice.id,
                 "quantity": 1,
-                "tax_ids": False,
+                "tax_ids": [(5, )],
                 "unit_amount": invoice.amount_total,
             }
         )


pedrobaeza (changes requested):
I don't get the patch, as both fields should have the same amount, which is the total. The operation should be:

- You introduce 100 € of unit price + 21% IVA. Total: 121 €
- Create invoice.
- On the expense, the unit price turns into 121 € and taxes are removed, so the total is still 121 €

Why this is not working with current code?

victoralmau:
> I don't get the patch, as both fields should have the same amount, which is the total. The operation should be:
> 
>     * You introduce 100 € of unit price + 21% IVA. Total: 121 €
> 
>     * Create invoice.
> 
>     * On the expense, the unit price turns into 121 € and taxes are removed, so the total is still 121 €
> 
> 
> Why this is not working with current code?

Because although the taxes are left blank, defining the total amount of the invoice as the unit price of the expense auto-changes the total amount of the expense by increasing the taxes. 
This does NOT happen in v13 although it is the same code.

pedrobaeza:
That can't be, so there's something incorrect, but this is definitively not the correct patch. Please dive more.

victoralmau:
> That can't be, so there's something incorrect, but this is definitively not the correct patch. Please dive more.

Perfect, what do you think of my new proposal?

pedrobaeza (approved):

pedrobaeza:
Now this has more sense. Fast-tracking the patch:

/ocabot merge patch

PR #1021, opened 2021-08-27
[13.0][ADD] hr_recruitment_notification: subscribe users to new applications automatically

joao-p-marques:
New module that provides a new message subtype to allow users to be automatically subscribed to new applications from a Job Position.

@Tecnativa
TT30413

ping @pedrobaeza @Yajo


pedrobaeza (approved):

hr_recruitment_notification/data/message_subtypes.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Tecnativa - João Marques -->
+<odoo>
+    <data noupdate="1">
```
yajo, line 4: It seems to me like this shouldn't be noupdate.

hr_recruitment_notification/data/message_subtypes.xml
```diff
@@ -0,0 +1,15 @@
...
+        <!-- Message Subtype related to Job Positions -->
+        <record id="mt_hr_applicant_new" model="mail.message.subtype">
+            <field name="name">New Applicant</field>
+            <field name="description">Applicant created</field>
+            <field name="res_model">hr.job</field>
+            <field name="default" eval="False" />
+            <field name="parent_id" eval="ref('hr_recruitment.mt_applicant_new')" />
```
yajo:
```suggestion
            <field name="parent_id" ref="hr_recruitment.mt_applicant_new" />
```

yajo (changes requested):
Please open PR for upstream odoo master about this.

hr_recruitment_notification/data/message_subtypes.xml
joao-p-marques: Well, fine by me, it just seems that that is the standard in all Odoo message subtypes. For example, in a runbot with all core modules installed:

image

joao-p-marques:
Thanks @Yajo!

Changes made and PR opened to Odoo in https://github.com/odoo/odoo/pull/75736

hr_recruitment_notification/data/message_subtypes.xml
yajo: Still missing.

yajo (changes requested):

hr_recruitment_notification/data/message_subtypes.xml
pedrobaeza: They are noupdate in core, and they should be that way for allowing any modification (for example, `default` flag).

yajo: Ah that makes sense.

yajo (approved):

PR #1022, opened 2021-09-01
[12.0][FIX] hr_employee_calendar_planning: Force to auto generate calendar even when calendar is not active.

victoralmau:

@Tecnativa TT31606

### DIFF
--- a/hr_employee_calendar_planning/migrations/12.0.1.1.0/post-migration.py
+++ b/hr_employee_calendar_planning/migrations/12.0.1.1.0/post-migration.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from openupgradelib import openupgrade
+
+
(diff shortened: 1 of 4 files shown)


hr_employee_calendar_planning/migrations/12.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE resource_calendar
+        SET auto_generate = true
+        WHERE active = false""",
```
chienandalu, line 13: Should we ensure it better filtering those with the name like   'Auto generated calendar for employee'

pedrobaeza: It can't be due to the translations.

pedrobaeza (approved):

chienandalu (approved):
Ok :+1:

PR #1023, opened 2021-09-03
[13.0][FIX] hr_employee_calendar_planning: Force to auto generate calendar even when calendar is not active

victoralmau:
FWP from 12.0: https://github.com/OCA/hr/pull/1022

@Tecnativa TT31606

### DIFF
--- a/hr_employee_calendar_planning/migrations/13.0.1.2.0/post-migration.py
+++ b/hr_employee_calendar_planning/migrations/13.0.1.2.0/post-migration.py
@@ -1,0 +1,15 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if version >= "13.0.1.0.0":
+        openupgrade.logged_query(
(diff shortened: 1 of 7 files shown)


hr_employee_calendar_planning/migrations/13.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,14 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE resource_calendar
```
pedrobaeza: You have to take into account that you can come from `12.0.1.1.0` and this was already done. One possible technique is to check `version` that stores version from which you come, so if it's `>= 13.0.1.0.0`, then you can perform this.

pedrobaeza (commented):
Do you want to be the maintainer of the module (I can share with you such load if you want)?

pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
/ocabot merge nobump

To v14!

victoralmau:
I think the bot is blocked, can you try again?

PR #1025, opened 2021-09-08
[14.0][FIX] hr_employee_calendar_planning: Force to auto generate calendar even when calendar is not active

victoralmau:
FWP from 13.0: https://github.com/OCA/hr/pull/1023

@Tecnativa TT31606


pedrobaeza (approved):
I'm seeing now that the correct spelling would be `auto_generated`, not `auto_generate`, but let's keep it for not requiring more work. Please take care of the correct English next time.

pedrobaeza:
Fast-tracking this as present in 2 previous versions:

/ocabot merge nobump

PR #1033, opened 2021-10-04
[FIX] hr_employee_service_contract: Set fields as not prefetched

etobella:
Without this change, an error is raised if a non HR user creates a leave, as it replaces hr.employee by hr.employee.public, but it cannot read this fields.


alexey-pelykh (approved):

LoisRForgeFlow (approved):

etobella:
@OCA/human-resources-maintainers cant this be merged?

PR #1035, opened 2021-10-15
[13.0][FIX] hr_employee_calendar_planning: Regenerate automatic calendar 2 weeks.

victoralmau:
Regenerate automatic calendar correctly when calendars are 2 weeks old (or any of them are 2 weeks old).

Steps to reproduce:
- Define several calendar planning in an employee.
- The calendars must be of type 2 weeks (or at least one of them).

This is necessary to apply also to v14.

@Tecnativa TT32425

### DIFF
--- a/hr_employee_calendar_planning/migrations/13.0.1.3.0/post-migration.py
+++ b/hr_employee_calendar_planning/migrations/13.0.1.3.0/post-migration.py
@@ -1,0 +1,12 @@
+# Copyright 2021 Tecnativa - Víctor Martínez
+# Copyright 2021 Tecnativa - Pedro M. Baeza
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env["hr.employee"].with_context(active_test=False).search([]).filtered(
+        lambda x: any(c.calendar_id.two_weeks for c in x.calendar_ids)
+        and any(not c.calendar_id.two_weeks for c in x.calendar_ids)
+    ).regenerate_calendar()
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -72,41 +72,33 @@ class HrEmployee(models.Model):
             self.resource_calendar_id.attendance_ids.unlink()
             self.resource_calendar_id.two_weeks_calendar = two_weeks
         seq = 0
-        for line in self.calendar_ids:
-            for attendance_line in line.calendar_id.attendance_ids:
-                if attendance_line.display_type == "line_section":
-                    continue
-                duplicate = two_weeks and not line.calendar_id.two_weeks_calendar
-                week_odd = attendance_line.week_type == "1"
-                data = attendance_line.copy_data(
-                    {
-                        "calendar_id": self.resource_calendar_id.id,
-                        "date_from": line.date_start,
-                        "date_to": line.date_end,
-                        "week_type": "0" if duplicate else attendance_line.week_type,
-                        "sequence": seq + 26 if week_odd else 24 - seq,
-                        # To make sure sequence of odd weeks is greater than 25
-                        # and sequence of even weeks is less than 25, as in
-                        # /resource/models/resource.py#L266
-                    }
-                )[0]
+        for week in ["0", "1"] if two_weeks else ["0"]:
+            if two_weeks:
+                section_vals = SECTION_LINES[int(week)]
+                section_vals[2]["sequence"] = seq
+                vals_list.append(section_vals)
                 seq += 1
-                vals_list.append((0, 0, data))
-                if duplicate:
+            for line in self.calendar_ids:
+                if line.calendar_id.two_weeks_calendar:
+                    attendances = line.calendar_id.attendance_ids.filtered(
+                        lambda x: x.week_type == week
+                    )
+                else:
+                    attendances = line.calendar_id.attendance_ids
+                for attendance_line in attendances:
+                    if attendance_line.display_type == "line_section":
+                        continue
                     data = attendance_line.copy_data(
                         {
                             "calendar_id": self.resource_calendar_id.id,
                             "date_from": line.date_start,
                             "date_to": line.date_end,
-                            "week_type": "1",
-                            "sequence": seq + 26,
-                        },
+                            "week_type": week if two_weeks else False,
+                            "sequence": seq,
+                        }
                     )[0]
                     seq += 1
-                    vals_list += [(0, 0, data)]
-        if two_weeks:
-            SECTION_LINES[0][2]["sequence"] = -seq
-            vals_list = SECTION_LINES + vals_list
+                    vals_list.append((0, 0, data))
         self.resource_calendar_id.attendance_ids = vals_list
         # Set the hours per day to the last (top date end) calendar line to apply
         if self.calendar_ids:
@@ -115,7 +107,8 @@ class HrEmployee(models.Model):
             ].calendar_id.hours_per_day
 
     def regenerate_calendar(self):
-        self._regenerate_calendar()
+        for item in self:
+            item._regenerate_calendar()
 
 
 class HrEmployeeCalendar(models.Model):
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -50,25 +50,7 @@ class HrEmployee(models.Model):
         two_weeks = bool(
             self.calendar_ids.mapped("calendar_id").filtered("two_weeks_calendar")
         )
-        if (
-            not self.resource_id.calendar_id
-            or not self.resource_id.calendar_id.auto_generate
-        ):
-            self.resource_id.calendar_id = (
-                self.env["resource.calendar"]
-                .create(
-                    {
-                        "active": False,
-                        "auto_generate": True,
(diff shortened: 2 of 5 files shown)


hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -3,36 +3,28 @@
...
-            "sequence": "25",
-            "hour_from": 0,
-            "day_period": "morning",
-            "week_type": "1",
-            "hour_to": 0,
-            "display_type": "line_section",
-        },
-    ),
-]
+SECTION_LINES = {
```
pedrobaeza: I don't see the need to change this data structure and increasing the diff

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -50,10 +42,49 @@ def _regenerate_calendar(self):
         two_weeks = bool(
             self.calendar_ids.mapped("calendar_id").filtered("two_weeks_calendar")
         )
-        if (
-            not self.resource_id.calendar_id
-            or not self.resource_id.calendar_id.auto_generate
-        ):
+        if self.resource_id.calendar_id.auto_generate:
+            self.resource_calendar_id.attendance_ids.unlink()
```
pedrobaeza, line 54: Here you are removing all the previous entries, but I think this has problems when you have put previous leaves linked to these entries. Can you please check?

pedrobaeza (commented):
Can you reduce the diff in general for being more reviewable?

hr_employee_calendar_planning/models/hr_employee.py
victoralmau: Previously all data was deleted. I think this is necessary because if we want to re-generate the calendar it is important that there is no previous data in the auto-generated calendar exclusively for the related employee.

I have moved it because there was an error in this regard if calendars were defined in the creation of the employee.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -117,6 +119,13 @@ def _regenerate_calendar(self):
     def regenerate_calendar(self):
         self._regenerate_calendar()
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        if res.calendar_ids:
+            res.regenerate_calendar()
```
chienandalu: Maybe this line would deserve a test cover

victoralmau:
I have optimized the code so that there is a smaller diff and I have added a migration script to re-generate the calendars that will be incorrect.
Can you check it again to merge it?

hr_employee_calendar_planning/models/hr_employee.py
victoralmau: Reviewing this again, leaves is not the same as attendances, in this case we only eliminate attendances (the working days of this calendar) therefore, the leaves will continue to be linked to this same calendar.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -50,10 +50,40 @@ def _regenerate_calendar(self):
...
         )
-        if (
-            not self.resource_id.calendar_id
-            or not self.resource_id.calendar_id.auto_generate
-        ):
+        if self.resource_id.calendar_id.auto_generate:
+            self.resource_calendar_id.attendance_ids.unlink()
+            self.resource_calendar_id.two_weeks_calendar = two_weeks
+        seq = 0
+        for week in {"0", "1"} if two_weeks else {"0"}:
```
pedrobaeza: Sorry, I didn't remember that sets don't respect order:

```suggestion
        for week in ["0", "1"] if two_weeks else ["0"]:
```

hr_employee_calendar_planning/migrations/13.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if version >= "13.0.1.0.0":
+        calendars = (
+            env["resource.calendar"]
+            .with_context(active_test=False)
+            .search([("auto_generate", "=", True), ("two_weeks_calendar", "=", True)])
+        )
+        employees = env["hr.employee"].search([("calendar_ids", "in", calendars.ids)])
```
pedrobaeza: Shouldn't only those with mix of both types of calendar be regenerated?

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -50,10 +50,40 @@ def _regenerate_calendar(self):
...
+            if two_weeks:
+                key_section = 0 if week == "0" else 1
+                section_item = SECTION_LINES[key_section][2]
+                section_item["sequence"] = seq
+                vals_list.append((0, 0, section_item))
```
pedrobaeza:
```suggestion
                section_vals = SECTION_LINES[int(key_section)
                section_vals[2]["sequence"] = seq
                vals_list.append(section_vals)
```

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -62,52 +92,14 @@ def _regenerate_calendar(self):
...
-                            "sequence": seq + 26,
-                        },
-                    )[0]
-                    seq += 1
-                    vals_list += [(0, 0, data)]
-        if two_weeks:
-            SECTION_LINES[0][2]["sequence"] = -seq
-            vals_list = SECTION_LINES + vals_list
-        self.resource_calendar_id.attendance_ids = vals_list
+            self.resource_calendar_id.attendance_ids = vals_list
```
pedrobaeza, line 101: Why do we need this reassignation on the else?

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -117,6 +109,13 @@ def _regenerate_calendar(self):
     def regenerate_calendar(self):
         self._regenerate_calendar()
 
+    @api.model_create_multi
+    def create(self, vals_list):
```
pedrobaeza, line 113: Why this extra `create` overriding?

hr_employee_calendar_planning/models/hr_employee.py
victoralmau: Remember that now the employee's automatic calendar does not create a new one each time, for this reason, when there is already one defined and we modify the employee's calendars, the calendar lines must be modified completely.

Actually this is not something new, it is something that is already done.

hr_employee_calendar_planning/models/hr_employee.py
victoralmau: This is necessary because if calendars are defined in the creation of the employee, the default calendar of the company is defined as the employee calendar (`resource_calendar_id` field), and this is incorrect, the automatic employee calendar should be auto-generated. The `test_employee_with_calendar_ids()` test validates just this.

hr_employee_calendar_planning/migrations/13.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+    if version >= "13.0.1.0.0":
+        calendars_two_weeks = env["resource.calendar"].search(
+            [("auto_generate", "=", False), ("two_weeks_calendar", "=", True)]
+        )
+        calendars = env["resource.calendar"].search(
+            [("auto_generate", "=", False), ("two_weeks_calendar", "=", False)]
+        )
+        employees = env["hr.employee"].search(
+            [
+                ("calendar_ids.calendar_id", "in", calendars_two_weeks.ids),
```
pedrobaeza: As these are subqueries, the AND conditions doesn't work properly. You have to loop on all employees.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -62,60 +91,30 @@ def _regenerate_calendar(self):
...
 
     def regenerate_calendar(self):
-        self._regenerate_calendar()
+        for item in self:
+            item._regenerate_calendar()
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        res = super().create(vals_list)
+        if res.calendar_ids:
```
pedrobaeza: This can be multi-record

hr_employee_calendar_planning/migrations/13.0.1.3.0/post-migration.py
```diff
@@ -0,0 +1,21 @@
...
+    if version >= "13.0.1.0.0":
+        calendars_two_weeks = env["resource.calendar"].search(
+            [("auto_generate", "=", False), ("two_weeks_calendar", "=", True)]
+        )
+        calendars = env["resource.calendar"].search(
+            [("auto_generate", "=", False), ("two_weeks_calendar", "=", False)]
+        )
+        for employee in env["hr.employee"].search([]):
+            if bool(
+                employee.mapped("calendar_ids.calendar_id").filteed(
```
pedrobaeza: You have to check if there's any employee with mix of two and one calendar to regenerate. The rest are OK

victoralmau: In my opinion, all the employees that in their auto-generated calendar have a calendar of 2 weeks and another without 2 weeks should be regenerated because so far the auto-generated calendar of that employee is partially incorrect.
Do you think it should be filtered in another way?

pedrobaeza: But this is not what you are checking right now. Anyways, it shouldn't be on the autogenerated calendar, but checking `calendar_ids` if there's a mix of both types.

pedrobaeza (approved):

PR #1029, opened 2021-09-29
[13.0][ADD] hr_personal_equipment_stock

albariera:
This module expands the functionality of the hr_personal_equipment request module proposed at #1026, allowing integration with stock.

PR #1028, opened 2021-09-29
[ADD] hr_personal_equipment_variant_configurator

albariera:
This module expands the functionality of the hr_personal_equipment request module proposed at #1026, allowing to select between product attributes.

PR #1027, opened 2021-09-28
[IMP] hr_employee_ppe: adapt it to new module hr_personal_equipment_request

albariera:
This PR changes the module of hr_employee_ppe adapting it to the new module hr_personal_equipment_request that is proposed in this PR #1026. This allows managing employee equipment as a product, creating equipment requests, and controlling allocations. 
This module includes additional functionalities to hr_personal_equipment_request as managing equipment duration and receipt report for ppe.
To pass the tests, the branch from #1026 has been added at test-requirements, but the PR should be merged in order to work properly.

PR #1026, opened 2021-09-27
[13.0][MIG] hr_personal_equipment_request: Migration to 13.0

albariera:
This module is still not merged into version 12.0. The PR can be found at #1004

PR #1042, opened 2021-10-25
[MIG] hr_employee_ppe: Migration to 14.0

TelmoSuarezConinpe:


albariera:
In version 12.0, after talking with @marcelsavegnago, the developer of hr_employee_ppe, a refactor was decided to be made basing this module on a new module called hr_personal_equipment, which enables to manage personal equipment, not only ppe. You can see all the information at #1004. The PR of all the migration of that modules are also published at #1026, #1027, #1028 and #1029. 
I hope this new refactor fits to you as well. If so, please close this PR in order to avoid confusion. Thank you

PR #1044, opened 2021-10-26
[IMP] hr_employee_calendar_planning: Define calendar in N employees

luistorresm:
Now was added a wizard that allow define the same calendar planning for
N employees.

This is very common when the company has employees in the same group
and assign the same planning for all the group.

### DIFF
--- a/hr_employee_calendar_planning/__init__.py
+++ b/hr_employee_calendar_planning/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from . import models
+from . import wizards
 from .hooks import post_init_hook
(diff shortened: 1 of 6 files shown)


luistorresm:
Hello @pedrobaeza @victoralmau 

What do you think about this improvement?

Note: Runbot has a warning, but, also the stable version for v14.0.

Regards

victoralmau:
Thanks for the input.
In my opinion it is a good idea but I think it would be necessary to change the focus and assign something like `wizard.generate.calendar.planning.calendar` (`date_from`, `date_to` and `calendar_id`) instead of `hr.employee.calendar`. 

Clicking on **Generate** button should add those records (creating new records from hr.employee.calendar) to all the selected employees.

Adding tests would be useful to validate if it works correctly.

pedrobaeza:
Is this going to be finished to reopen it? Or should it stay closed?

PR #1045, opened 2021-10-26
[FIX] hr_employee_calendar_planning: Avoid error with parent

luistorresm:
Parent is not necessary on context, this returns the next error when you
try to select a calendar planning on the employee:
``Uncaught Error: NameError: name 'parent' is not defined``

Screen Shot 2021-10-26 at 14 00 15

### DIFF
--- a/hr_employee_calendar_planning/views/hr_employee_views.xml
+++ b/hr_employee_calendar_planning/views/hr_employee_views.xml
@@ -10,10 +10,7 @@
                 <attribute name="invisible">1</attribute>
             </field>
             <field name="resource_calendar_id" position="after">
-                <field
-                    name="calendar_ids"
-                    context="{'default_company_id': parent.company_id}"
-                >
+                <field name="calendar_ids" context="{'default_company_id': company_id}">
                     <tree editable="top">
                         <field name="company_id" invisible="1" />
                         <field name="date_start" />


luistorresm:
Hello @pedrobaeza @victoralmau 

Could you review this fix and approve if you agree, please?

Regards

pedrobaeza:
If you try in OCA runbot before the patch, there's no such error, so you should have something bad in your end.

etobella:
@pedrobaeza I replicated it on the runbot. In order to replicate it, you need to do the following:

1- Access in debug mode
2- Access Employee
3- Enter on one employee
4- Click on one calendar plan without being on edition

Used this runbot from 13.0 (https://runbot.odoo-community.org/runbot/build/3510056)

luistorresm:
> If you try in OCA runbot before the patch, there's no such error, so you should have something bad in your end.

Hello @pedrobaeza 

Here is the video:

https://user-images.githubusercontent.com/7606656/139085805-59477133-416b-4f55-a6e5-9eb8d112aa3a.mp4

pedrobaeza:
The problem is that this is removing the intention of the fix in #1012, as you are losing the default company coming from the parent when creating a new entry, and thus failing on saving due to the check_company.

luistorresm:
> The problem is that this is removing the intention of the fix in #1012, as you are losing the default company coming from the parent when creating a new entry, and thus failing on saving due to the check_company.

Sorry @pedrobaeza 

I do not understand,

What do you propose to fix this error?

Regards

pedrobaeza:
I don't have any proposal right now. I'm just saying that doing this, you are fixing one thing, and breaking another.

etobella:
But there is no parent on that point, the default company is kept, so I don't get the issue

pedrobaeza:
`parent` is the employee form, as the context is passed to the one2many field.

etobella:
But parent is used on the onemany field definition not in the form or tree inside

etobella (approved):

etobella:
@pedrobaeza Can you recheck the issue, we have the same problem on our installation and it is solved with the change

albariera (approved):

victoralmau (approved):

victoralmau:
Can you retry again merge process?

PR #1058, opened 2021-12-06
[12.0][FIX] hr_holidays_public: Remove existing link with meeting

ypapouin:
Otherwise new holidays dates will overwrite existing meetings dates


pedrobaeza:
Does this create new meetings then?

ypapouin:
Yes, it was already handled by the `create` method from `hr.holidays.public`
The issue comes from the constraint `_update_calendar_event` immediately
called after the `create`, just before a new meeting is created and
attached.

>

pedrobaeza (approved):

PR #1065, opened 2021-12-24
[MIG] resource_hook: Migration to 14.0

olgamarcocb:
Migrated into odoo addons (resource) (!)

PR #1068, opened 2022-01-20
[MIG] hr_calendar_rest_time: Migration to 14.0

olgamarcocb:
Standard Migration.
Depends on  #1065


pedrobaeza:
Is `resource_hook` still needed?

olgamarcocb:
> Is `resource_hook` still needed?

It is, since resource_hook uses get_work_hours_interval. In fact, do you think we should make a proposal to Odoo in order to use hooks on the master branch?

hr_calendar_rest_time/models/resource_calendar_attendance.py
```diff
@@ -0,0 +1,31 @@
...
+
+    day_period = fields.Selection(
+        selection_add=[("all_day", "All Day")], ondelete={"all_day": "cascade"}
+    )
+    rest_time = fields.Float(string="Rest Time")
+
+    @api.onchange("rest_time")
+    def _onchange_rest_time(self):
+        # avoid negative or after midnight
+        self.rest_time = min(self.rest_time, 23.99)
```
albariera, line 20: Does this line make sense? It is not being overwritten by the next?

PR #1071, opened 2022-01-31
[14.0][FIX] hr_employee_service incorrect return format

remi-filament:
Function _get_date_start_work() should return a datetime, but it is overloaded here to return a date.
This causes an error when trying to process leave allocation accrual : 
```ValueError: <class 'TypeError'>: "can't compare datetime.datetime to datetime.date" while evaluating 'model._update_accrual()'```
when running cron job.

This PR converts service_start_date to datetime before being returned (if set) so that the error is fixed

### DIFF
--- a/hr_employee_service/models/hr_employee.py
+++ b/hr_employee_service/models/hr_employee.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2018-2019 Brainbean Apps (https://brainbeanapps.com)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from datetime import timedelta
+from datetime import datetime, time, timedelta
 from math import fabs
 
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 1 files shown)


remi-filament:
Most probably the issue is also present on v13.0, since we have the same function 
- on hr_employee_service returning a date : https://github.com/OCA/hr/blob/7f8fbdea46acc59ab91cbac22375b53251639bb0/hr_employee_service/models/hr_employee.py#L103
- on Odoo in hr_holidays module returning a datetime : https://github.com/odoo/odoo/blob/416a8502dba65c7109a0125824cf6079367dfb3c/addons/hr_holidays/models/hr_employee.py#L41

legalsylvain (approved):

hparfr (approved):

remi-filament:
not sure why runbot reports red, I see success in the logs and can access the instance...

jordancoest (approved):

nimarosa (approved):

Saran440:
@remi-filament Please, you update code again. it red because OCA change runbot to runboat already.

remi-filament:
@Saran440 done !

PR #1080, opened 2022-02-23
[13.0][FIX] resource_hook: Avoid crash when validating leaves

mrgoetz:
Fix the validation of leaves
    
    With module resource_hook installed the method
    _get_work_hours_interval(start, stop, meta)
    is called from a hr.employee (self) Object.
    This is not implemented for hr.employee unless you do it by yourself
    and Odoo crashes on second Approve.
    Since a calendar object exists you can use calendar instead of self.

### DIFF
--- a/resource_hook/hooks.py
+++ b/resource_hook/hooks.py
@@ -173,7 +173,7 @@ def post_load_hook():  # noqa: C901
         )[resource.id]
         result = defaultdict(float)
         for start, stop, meta in intervals:
-            result[start.date()] += self._get_work_hours_interval(start, stop, meta)
+            result[start.date()] += calendar._get_work_hours_interval(start, stop, meta)
         return sorted(result.items())
 
     def __new_list_leaves(self, from_datetime, to_datetime, calendar=None, domain=None):


mrgoetz:
Test: Install hr_holiday,project_timesheet_holidays,resource_hook in a Odoo DB 13.0
As admin (At least approver for hr_leaves) create a "time off" and approve.
The secoand approval "validate" lets Odoo crash with:

File "/opt/odoo/lib/odoo/addons/hr_holidays/models/hr_leave.py", line 1021, in action_validate
    employee_requests._validate_leave_request()
  File "/opt/odoo/lib/odoo/addons/project_timesheet_holidays/models/hr_holidays.py", line 68, in _validate_leave_request
    holiday._timesheet_create_lines()
  File "/opt/odoo/lib/odoo/addons/project_timesheet_holidays/models/hr_holidays.py", line 76, in _timesheet_create_lines
    self.date_to,
  File "/opt/odoo/lib/addons/oca-hr/resource_hook/hooks.py", line 176, in __new_list_work_time_per_day
    result[start.date()] += self._get_work_hours_interval(start, stop, meta)
AttributeError: 'hr.employee' object has no attribute '_get_work_hours_interval'

Instead of self._get_work_hours_interval we use calendar._get_work_hours_interval. 

Hint: self._get_work_hours_interval  appears several times, but we encountered no problem.

mrgoetz:
Hi all, as you can see the added line in my commit fails the test coverage check, since there is no test at all.
Since resource_hook overloads some upstream methods it is necessary to test with the concerned modules.
It might a solution to add an extra test module inheriting some modules with unit tests.

PR #1083, opened 2022-03-01
[13.0][FIX] hr_personal_equipment_stock: fix partner_id and origin wh…

albariera:
…en creating stock.picking
When a personal equipment request was done with more than one allocation, the picking did not have partner_id, nor origin. The reason was that stock moves did not have the same origin, as it was set the name of the allocation, instead of the name of the personal equipment request. This commit solves this issue.


etobella (approved):
LGTM :+1:

albariera:
@AntoniRomera @FernandoRomera

AntoniRomera (approved):

PR #1095, opened 2022-04-26
[14.0][FIX] hr_employee_calendar_planning: Avoid error with parent

victoralmau:
FWP from 13.0: https://github.com/OCA/hr/pull/1045

Parent is not necessary on context, this returns the next error when you
try to select a calendar planning on the employee:
``Uncaught Error: NameError: name 'parent' is not defined``

@Tecnativa


luistorresm (approved):
👍

PR #1118, opened 2022-06-09
15.0 [MIG] hr job category

ghost:
I migrated hr_job_category module from 14.0 version to 15.0 version.


ivantodorovich (approved):

ivantodorovich:
Tihs commit: https://github.com/OCA/hr/pull/1118/commits/cbcee1a6716e4771779cc36bc806a0238d1161d1 is removing other module's files.

I agree those files should be removed, because they're migration scripts for 14.0. But please do that (ideally) on a separate PR, or at least put a meaningful commit message. `[MIG] hr_job_category: Migration to 15.0` is not accurate.

sebalix:
@ivantodorovich removing the migration files is a normal step when migrating an OCA module, documented in the [migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0), and the way it is written it can be part of the migration commit. No issue to me :)

ivantodorovich:
> @ivantodorovich removing the migration files is a normal step when migrating an OCA module, documented in the [migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0), and the way it is written it can be part of the migration commit. No issue to me :)

Yes but that's not what I mean 😅 
It's removing migration files from `hr_course` module, and that module is not being migrated in this PR.
But even if we include it in this PR (which would be fine IMO), the commit message refers to `hr_job_category`, not `hr_course`

My suggestion is to rename https://github.com/OCA/hr/commit/cbcee1a6716e4771779cc36bc806a0238d1161d1 's message to `[FIX] hr_course: remove old migration scripts` or something like it 😅

sebalix:
ohhhh, didn't see it was related to `hr_course`, thanks

Saran440:
@juliette-blanc please, fix conflict file.

leemannd:
@ivantodorovich @sebalix The migration scripts are related to another module, but still they should be present as they are related to the version `14.0`

leemannd:
Can be closed. It has been superseed by https://github.com/OCA/hr/pull/1121

PR #1124, opened 2022-06-20
[15.0][MIG] hr_period

JuanyDForgeflow:
Adds the concept of period in the human resources management.

The objective of the module is to create periods of time to
be used in the human resources management flows such as
specific payroll period of time or timesheet periods.

Regarding to payrolls, it adds the date of payment on the
payslip and payslip batch. This date is automatically filled
when selecting a period. It also adds a sequence on the payslip
batch name and the company on the payslip batch.


Saran440:
if this module use with payroll. maybe you move this module to https://github.com/OCA/payroll

nimarosa:
I agree. I'm the PSC of oca/payroll you are free to move this module there. We still making improvements to version 14.0 before we start 15.0 migration so we can start with moving the 14.0 version.

nimarosa:
Closed. This module will be moved to OCA/payroll currently in PR for 14.0 branch. https://github.com/OCA/payroll/pull/89

When merged, we can migrate it there.

PR #1128, opened 2022-07-06
[14.0][IMP] hr_employee_calendar_planning: add global leaves and time zone

jonasbuchholz:
Add global leaves and employee time zone to auto-generated calendar:
When creating auto-generated calendars, currently global leaves from parent calendars are not transferred. With this commit, global leaves from parent calendars are copied and linked to the auto-generated calendar while leaving the parent's global leaves untouched. If a calendar has a start or end date defined for calendar planning, only global leaves that fall between start and end date are copied to the auto-generated calendar.
Use employee time zone as default time zone when creating an auto-generated calendar, otherwise the post-install hook will revert to UTC.
Add tests for global leaves in auto-generated calendars.


victoralmau (approved):
The code would be correct and seems to make quite a bit of sense.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -97,18 +97,54 @@ def _regenerate_calendar(self):
...
                         "attendance_ids": vals_list,
                         "two_weeks_calendar": two_weeks,
+                        "tz": self.tz,  # take employee timezone as default
                     }
                 )
                 .id
             )
         else:
             self.resource_calendar_id.attendance_ids = vals_list
+
```
pedrobaeza: Please don't add empty lines in the middle of the method.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -97,18 +97,54 @@ def _regenerate_calendar(self):
...
+            ]
+
+    def copy_global_leaves(self, leave_ids):
+        vals = [
+            leave.copy_data({})[0]
+            for leave in self.env["resource.calendar.leaves"].browse(leave_ids)
+        ]
+        return self.env["resource.calendar.leaves"].create(vals).ids
+
+    @staticmethod
```
pedrobaeza: Don't add this kind of Python stuff that is not inheritabe in the Odoo ORM framework.

hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
```diff
@@ -359,3 +359,63 @@ def test_employee_with_calendar_ids(self):
...
+                    "date_from": "2020-03-09",
+                    "date_to": "2020-03-10",
+                },
+            )
+        ]
+        global_leave_ids_cal1 = self.calendar1.global_leave_ids.ids
+        self.employee.resource_calendar_id = self.calendar1.id
+
+        # test that global leaves are copied to the autogenerated calendar on post install hook
+        post_init_hook(self.env.cr, self.env.registry, self.employee)
```
pedrobaeza: Don't do this in the same method of the test

pedrobaeza (changes requested):

jonasbuchholz:
Thanks for your comments @pedrobaeza . I refactored the code as you suggested.
Since I think it doesn't make sense to create another separate test for the post_init_hook just for the global leaves I added the assertions to the existing test. However, this required some changes to the setup and existing tests, because global leaves also count towards the leave_ids. I hope you agree that this setup makes sense.
I also added the "global_leave_ids" parameter in the write method of the resource_calendar model. This way, changes in the global_leaves of calendars will be passed down to the auto-generated calendars.

pedrobaeza:
Great, thanks. Please squash commits into one.

jonasbuchholz:
Hi @pedrobaeza , is this ready to merge or do you request additional changes?

pedrobaeza (approved):

PR #1131, opened 2022-07-26
[14.0][MIG] hr_employee_document: Migration to 14.0

victoralmau:
Migration to 14.0

Please @cesar-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT37024


cesar-tecnativa (approved):

hr_employee_document/models/hr_employee.py
```diff
@@ -13,7 +13,6 @@ class HrEmployeeBase(models.AbstractModel):
     )
 
     def _compute_document_count(self):
-        self.documents_count = 0
```
pedrobaeza: Why removing this? I think it's still needed for avoiding missing caches.

pedrobaeza (approved):

PR #1133, opened 2022-07-29
[15.0][MIG] hr_expense_sale_invoice

Murtaza-OSI:


nimarosa:
@Murtaza-OSI Please move expense modules to the OCA/hr-expense repository. Thanks.

PR #1134, opened 2022-08-03
[IMP] hr_attendance_report_theoretical_time: department record rule

chienandalu:
Department responsibles should be able to check their subordinates theoretical hours reports.

cc @Tecnativa TT38378

please review @victoralmau @pedrobaeza


carlosdauden (approved):

victoralmau (approved):

pedrobaeza:
I don't think we should put this breaking change in a so old stable version, and more being about to migrate in our case. I think it's better to put in the instance.

chienandalu:
Should we add it in v15 then?

pedrobaeza:
We may evaluate it, but there are several possibilities, as you have the manager, but also the time off manager. Which one should be enabled? I think this is very specific to each company.

chienandalu:
I see. Ok, let's leave it as a custom choice then

PR #1145, opened 2022-10-25
[12.0][FIX] hr_expense_invoice: Change _validate_expense_invoice() function to use amount_total from invoices.

victoralmau:
Change `_validate_expense_invoice()` function to use amount_total from invoices.

If residual amount from invoice is not the same, it should not show an error.

@Tecnativa TT40175


pedrobaeza:
@kittiu what is the reason to use `residual` instead when you did the migration?

pedrobaeza (approved):

chienandalu (approved):

PR #1151, opened 2022-11-03
[14.0] Fix Github Actions

etobella:
updating template
make hr_period depend from the one from OCA/payroll and remove all logic here. Still some work to do, as commit history was not kept in OCA/payroll

@nimarosa


etobella:
Another option would be: delete the module from payroll and pass the logic here. On 15 we can migrate ith there and keep the original name

nimarosa:
Hello, didn't notice it was breaking here. 

I think the best will be to make the dummy module here like you are doing. 

In payroll I will try to copy the commit history to the other repo. I've also made some modifications to the module views in payroll since payroll module changes a lot and hr_period was breaking some views. 

The idea is to continue maintaining module in payroll repo. 

Is removing the module here a solution? Maybe we can remove the module from 14.0 here. I didn't do it because people might have it installed. But making the dummy module seems a good option.

sanchonuria (approved):

LoisRForgeFlow (approved):
LG from my POV, in later version can be finally removed and handled with a rename in migration scripts

nimarosa:
Merging. 

Today in a few hours i will work in copying the commit history to the new repo. But let's merge this to fix CI. 

/ocabot merge nobump

PR #1160, opened 2022-11-15
[15.0][FIX] hr_employee_birth_name: Fixed bug when accessing hr.employee

vincent-hatakeyama:


nimarosa:
@vincent-hatakeyama can you explain what bug is fixed?

moctarjallo:
@nimarosa we are getting the following error using the module:

`ValueError: Invalid field 'birth_name' on model 'hr.employee.public'`

This is because Odoo has already prevented this situation [here](https://github.com/odoo/odoo/blob/15.0/addons/hr/models/hr_employee.py) from line 21 to 27.

Basically you need to add access rights to the field like this `groups="hr.group_hr_user"`

etobella:
Wouldn't be better to add the data on the res.users profile in order to make it visible to the user:

https://github.com/odoo/odoo/blob/15.0/addons/hr/models/res_users.py#L157-L163

vincent-hatakeyama:
> Wouldn't be better to add the data on the res.users profile in order to make it visible to the user:
> 
> https://github.com/odoo/odoo/blob/15.0/addons/hr/models/res_users.py#L157-L163

I agree but I think this would be an improvement to add to a new issue/pr.

nimarosa (approved):

Saran440 (approved):

etobella (changes requested):
Add it following my comment please

pedrobaeza:
Actually, this is the official way to fix this kind of problem: https://github.com/odoo/odoo/blob/c55badb8d70bdccfeb82bffb330d3f0b1b162ed1/addons/hr/models/hr_employee.py#L23, so I see it correct.

nimarosa:
For me this fix it's okay. I think we should merge since it's important to fix al permissions errors. 

Is this also affecting 14.0 branch? Or only 15.0? 

/ocabot merge patch

vincent-hatakeyama:
> Is this also affecting 14.0 branch? Or only 15.0?

Looking at the code in 14.0, it seems affected too.

nimarosa:
> > Is this also affecting 14.0 branch? Or only 15.0?
> 
> 
> 
> Looking at the code in 14.0, it seems affected too.

@vincent-hatakeyama It would be good if you can port it to 14.0 and I think 16.0 will have the same problem. That way we maintain the same code in all versions.

PR #1168, opened 2022-11-30
[14.0][FIX] hr_employee_calendar_planning: prevent loop

jonasbuchholz:
When creating a new employee, the regenerate_calendar method creates a new auto-generated calendar and assigns it to the resource_id.calendar_id field. However, the global_leave_ids and hours_per_day are written to the resource_calendar_id field. Even though resource_calendar_id and resource_id.calendar_id are related fields due to the resource.mixin model, the method initially edits the wrong (original) calendar instead of the auto-generated calendar. This triggers the write method in the resource.calendar model, which again calls regenerate_calendar, thus creating a loop.
By writing global_leaves and hours_per_day into resource_id.calendar_id, the correct (auto-generated) calendar is edited, preventing the loop.

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -106,11 +106,11 @@ class HrEmployee(models.Model):
             self.resource_calendar_id.attendance_ids = vals_list
         # Set the hours per day to the last (top date end) calendar line to apply
         if self.calendar_ids:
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

victoralmau (approved):

pedrobaeza:
Please fw-port it to upper versions:

/ocabot merge patch

PR #1170, opened 2022-12-05
15.0 mig hr holidays leave overlap

slominchar:


hr_holidays_leave_overlap/models/hr_holidays.py
```diff
@@ -0,0 +1,117 @@
...
+
+class HrHolidays(models.Model):
+    _inherit = "hr.leave"
+
+    @api.model
+    def create(self, values):
+        values["employee_id"] = values.get("employee_id", False)
+        values["category_id"] = values.get("category_id", False)
+
+        including = self.search(
```
hieulucky111, line 18: use `search_count` to improve performance.

hr_holidays_leave_overlap/models/hr_holidays.py
```diff
@@ -0,0 +1,117 @@
...
+                ("date_to", ">", values["date_from"]),
+                ("date_to", "<=", values["date_to"]),
+                ("employee_id", "=", values["employee_id"]),
+                ("category_id", "=", values["category_id"]),
+                ("state", "!=", "refuse"),
+            ],
+            limit=1,
+        )
+
+        for leave in left_overlapping:
```
hieulucky111, line 52: use `limit=1` above, so do you need to use `for` here?

hieulucky111 (changes requested):

PR #895, opened 2020-09-29
[13.0][MIG] hr_job_category

jarroyomorales:
Migration of `hr_job_category` and simplified code.

@etobella @Nikul-Chaudhary @i-vyshnevska @alexey-pelykh

PR #1178, opened 2022-12-22
FIX: hr_job_category removes employee tags that are not into the contract

leemannd:
The module has been simplified during the migration 12.0 -> 13.0 https://github.com/OCA/hr/pull/895
Since this simplification contained a regression:
-  The employee tags are now stricly binded from the contracts & job positions
- Updating a contract would flush away all the existing tags

Tags on employee could come from distinct sources (manually added) or linked to other models.
We want to avoid this hard link.

For the moment, I only took back the code from 12.0 (with a small addition) which would require some optimizations.

### DIFF
--- a/hr_job_category/models/hr.py
+++ b/hr_job_category/models/hr.py
@@ -59,12 +59,13 @@ class HRContract(models.Model):
 
     def write(self, vals):
         prev_data = self.read(["job_id"])
-
+        if "employee_id" in vals and self.employee_id != vals.get("employee_id"):
+            self._remove_tags(self.employee_id.id, self.job_id.id)
         res = super().write(vals)
-
         # Go through each record and delete tags associated with the previous
         # job, then add the tags of the new job.
         #
+
         for contract in self:
             for data in prev_data:
                 if (
--- a/hr_job_category/models/hr.py
+++ b/hr_job_category/models/hr.py
@@ -23,16 +23,32 @@ class HRJob(models.Model):
 class HRContract(models.Model):
     _inherit = "hr.contract"
 
-    def _tag_employees(self, job_id):
-        if job_id:
-            job = self.env["hr.job"].browse(job_id)
-            self.mapped("employee_id").write(
-                {"category_ids": [(6, 0, job.category_ids.ids)]}
-            )
-        else:
-            for contract in self:
-                categories = contract.job_id and contract.job_id.category_ids.ids or []
-                contract.employee_id.write({"category_ids": [(6, 0, categories)]})
+    def _remove_tags(self, employee_id=None, job_id=None):
+        # TODO write tags only once
+        if not employee_id or not job_id:
+            return
+        employee = self.env["hr.employee"].browse(employee_id)
+        empl_tags = employee.category_ids
+        job = self.env["hr.job"].browse(job_id)
+        _logger.debug(
+            "Removing employee tags if tag exists on contract " "job: %s", empl_tags
+        )
+        for tag in job.category_ids:
+            if tag in empl_tags:
+                employee.write({"category_ids": [(3, tag.id)]})
+
+    def _tag_employees(self, employee_id=None, job_id=None):
+        if not employee_id or not job_id:
+            return
+        employee = self.env["hr.employee"].browse(employee_id)
+        empl_tags = employee.category_ids
+        job = self.env["hr.job"].browse(job_id)
+        for tag in job.category_ids:
+            if tag not in empl_tags:
+                _logger.debug(
+                    "Adding employee tag if job tag doesn't " "exists: %s", tag.name
+                )
+                employee.write({"category_ids": [(4, tag.id)]})
 
     @api.model
     def create(self, vals):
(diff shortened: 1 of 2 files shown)


hr_job_category/models/hr.py
```diff
@@ -42,13 +58,31 @@ def create(self, vals):
         return res
 
     def write(self, vals):
-        if "employee_id" in vals:
-            self.mapped("employee_id").write({"category_ids": [(5,)]})
+        prev_data = self.read(["job_id"])
+        if "employee_id" in vals and self.employee_id != vals.get("employee_id"):
```
leemannd, line 65: Leaving soon for holidays. Hugly and needs to be changed.
But at least we have the base behavior

hr_job_category/models/hr.py
```diff
@@ -23,16 +23,32 @@ class HRJob(models.Model):
...
-                {"category_ids": [(6, 0, job.category_ids.ids)]}
-            )
-        else:
-            for contract in self:
-                categories = contract.job_id and contract.job_id.category_ids.ids or []
-                contract.employee_id.write({"category_ids": [(6, 0, categories)]})
+    def _remove_tags(self, employee_id=None, job_id=None):
+        # TODO write tags only once
+        if not employee_id or not job_id:
+            return
```
santostelmo: Easy to test that condition for better test coverage. Or maybe this method does not need to return a value

santostelmo (approved):

leemannd:
Hello @jarroyomorales You have been doing a migration of this module with a simplification of the code. But it has been introducing a regression (listed in the PR description).
cc @etobella if you want to have a look at it as you reviewed the previous PR
cc @Saran440 as you merged the migration PR

hr_job_category/models/hr.py
```diff
@@ -42,13 +58,31 @@ def create(self, vals):
         return res
 
     def write(self, vals):
-        if "employee_id" in vals:
-            self.mapped("employee_id").write({"category_ids": [(5,)]})
+        prev_data = self.read(["job_id"])
```
florentx: prev_data could be a dict:
```suggestion
        prev_data = {
            res["id"]: res["job_id"][0]
            for res in self.read(["job_id"]) if res["job_id"]
        }
```

hr_job_category/models/hr.py
```diff
@@ -42,13 +58,31 @@ def create(self, vals):
...
+        #
+
+        for contract in self:
+            for data in prev_data:
+                if (
+                    data.get("id") == contract.id
+                    and data["job_id"]
+                    and data["job_id"][0] != contract.job_id.id
+                ):
+                    self._remove_tags(contract.employee_id.id, data["job_id"][0])
```
florentx: Then if prev_data is a dict, it simplifies:
```suggestion
            job_id = prev_data.get(contract.id)
            if job_id:
                contract._remove_tags(job_id)
```

hr_job_category/models/hr.py
```diff
@@ -42,13 +58,31 @@ def create(self, vals):
...
+                self._tag_employees(contract.employee_id.id, contract.job_id.id)
```
florentx: When signature is changed back to original one:
```suggestion
                self._tag_employees(contract.job_id.id)
```

hr_job_category/models/hr.py
```diff
@@ -42,13 +58,31 @@ def create(self, vals):
...
-        self.mapped("employee_id").write({"category_ids": [(5,)]})
-        return super().unlink()
+        prev_data = self.read(["job_id"])
+        # Go through each record and delete tags associated with the previous
+        # job, then add the tags of the new job.
+        #
+        for contract in self:
+            for data in prev_data:
+                self._remove_tags(contract.employee_id.id, data["job_id"][0])
+        return super(HRContract, self).unlink()
```
florentx: This can be simplified:
```suggestion
        prev_data = {
            res["id"]: res["job_id"][0]
            for res in self.read(["job_id"]) if res["job_id"]
        }
        # Go through each record and delete tags associated with the previous job
        for contract in self:
            job_id = prev_data.get(contract.id)
            if job_id:
                contract._remove_tags(job_id)
        return super().unlink()
```

hr_job_category/models/hr.py
```diff
@@ -23,16 +23,32 @@ class HRJob(models.Model):
...
+            return
+        employee = self.env["hr.employee"].browse(employee_id)
+        empl_tags = employee.category_ids
+        job = self.env["hr.job"].browse(job_id)
+        _logger.debug(
+            "Removing employee tags if tag exists on contract " "job: %s", empl_tags
+        )
+        for tag in job.category_ids:
+            if tag in empl_tags:
+                employee.write({"category_ids": [(3, tag.id)]})
```
florentx: Simplify method signature?
```suggestion
    def _remove_tags(job_id):
        job = self.env["hr.job"].browse(job_id)
        for employee in self.mapped("employee_id"):
            _logger.debug(
                "Removing employee tags if tags exist on contract job: %s",
                employee.category_ids,
            )
            tags_to_remove = [
                (3, tag.id)
                for tag in job.category_ids & employee.category_ids
            ]
            if tags_to_remove:
                employee.write({"category_ids": tags_to_remove})
```

hr_job_category/models/hr.py
```diff
@@ -23,16 +23,32 @@ class HRJob(models.Model):
...
+            return
+        employee = self.env["hr.employee"].browse(employee_id)
+        empl_tags = employee.category_ids
+        job = self.env["hr.job"].browse(job_id)
+        for tag in job.category_ids:
+            if tag not in empl_tags:
+                _logger.debug(
+                    "Adding employee tag if job tag doesn't " "exists: %s", tag.name
+                )
+                employee.write({"category_ids": [(4, tag.id)]})
```
florentx: I would recommend to keep same method signature as before
```suggestion
    def _tag_employees(self, job_id):
        job = self.env["hr.job"].browse(job_id)
        _logger.debug(
            "Adding employee tags if job tags doesn't exist: %s", job.category_ids
        )
        for employee in self.mapped("employee_id"):
            tags_to_add = [
                (4, tag.id)
                for tag in job.category_ids - employee.category_ids
            ]
            if tags_to_add:
                employee.write(tags_to_add)
```

florentx (changes requested):
Need to remove the `stale` label on this PR.

And refactor proposed to simplify and have less writes.

PR #1180, opened 2022-12-27
[16.0][MIG] hr_employee_relative: Migration to 16.0.

BT-anieto:


BT-anieto:
@pedrobaeza  I created this PR for hr_employee_relative.

requirements.txt
```diff
@@ -0,0 +1,2 @@
+# generated from manifests external_dependencies
```
BT-rmartin, line 1: @pedrobaeza Is this auto-generated? How?

hr_employee_relative/views/hr_employee_relative.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="name">hr.employee.relative.view.tree</field>
+        <field name="model">hr.employee.relative</field>
+        <field name="arch" type="xml">
+            <tree editable="bottom">
+                <field name="relation_id" />
+                <field name="name" />
+                <field name="partner_id" />
+                <field name="gender" />
+                <field name="date_of_birth" />
+                <field name="age" readonly="1" />
```
BT-rmartin: No need for readonly, a non-stored computed field is now by default readonly

hr_employee_relative/views/hr_employee_relative.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
BT-rmartin, line 1: XML header is not needed

hr_employee_relative/tests/test_hr_employee_relatives.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
+from odoo.tests import Form, common
+
+_ns = "hr_employee_relative"
```
BT-rmartin: Is there really a need for this? It's just reused twice and the legibility is much worse in the self.env.ref
In fact you could load at the beginning of the test self.env.ref("hr_employee_relative.relation_sibling") and reuse it in lines 28 and 48

hr_employee_relative/tests/test_hr_employee_relatives.py
```diff
@@ -0,0 +1,50 @@
...
+            }
+        )
+        relative = self.EmployeeRelative.browse(employee.relative_ids[0].id)
+        self.assertEqual(int(relative.age), 42)
+        # onchange partner
+        ctx = {
+            "active_ids": [relative.id],
+            "active_id": relative.id,
+            "active_model": "hr.employee.relative",
+        }
```
BT-rmartin: Is this really needed?
When using Form in line 46 it will try to open any editable view of hr.employee.relative. And we have defined the tree view as editable, and it has partner_id and relation_id
No need at all then to do this. By default it will load that view and will be able to apply the onchanges there. 

with Form(relative) as f: should be enough

hr_employee_relative/tests/test_hr_employee_relatives.py
```diff
@@ -0,0 +1,50 @@
...
+        self.assertEqual(relative.name, "Relative")
+        with Form(self.EmployeeRelative.with_context(**ctx)) as f:
+            f.partner_id = self.env.ref("base.res_partner_2")
+            f.relation_id = self.env.ref(_ns + ".relation_sibling")
+        relative = f.save()
```
BT-rmartin: This is also not needed as when you exit from with the changes are already saved

hr_employee_relative/models/hr_employee_relative.py
```diff
@@ -0,0 +1,45 @@
...
+    name = fields.Char(required=True)
+    partner_id = fields.Many2one(
+        "res.partner",
+        string="Partner",
+        domain=["&", ("is_company", "=", False), ("type", "=", "contact")],
+    )
+    gender = fields.Selection(
+        selection=[("male", "Male"), ("female", "Female"), ("other", "Other")],
+    )
+    date_of_birth = fields.Date(string="Date of Birth")
```
BT-rmartin: No need for string

hr_employee_relative/models/hr_employee_relative.py
```diff
@@ -0,0 +1,45 @@
...
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class HrEmployeeRelative(models.Model):
+    _name = "hr.employee.relative"
+    _description = "HR Employee Relative"
+
+    employee_id = fields.Many2one(string="Employee", comodel_name="hr.employee")
```
BT-rmartin: No need for string and I would also remove the comodel_name, just giving as first parameter the model value

hr_employee_relative/models/hr_employee_relative.py
```diff
@@ -0,0 +1,45 @@
...
+    relation_id = fields.Many2one(
```
BT-rmartin: No need for string

hr_employee_relative/models/hr_employee_relative.py
```diff
@@ -0,0 +1,45 @@
...
+        "hr.employee.relative.relation", string="Relation", required=True
+    )
+    name = fields.Char(required=True)
+    partner_id = fields.Many2one(
```
BT-rmartin, line 18: No need for string

hr_employee_relative/models/hr_employee.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    relative_ids = fields.One2many(
```
BT-rmartin, line 10: No need for string

hr_employee_relative/i18n/de.po
```diff
@@ -0,0 +1,190 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* hr_employee_relative
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 14.0\n"
```
BT-rmartin, line 7: @pedrobaeza Not sure how the translations work now with Weblate, if these po files need to be here or they should be removed and once the translation is in Weblate they are automatically merged here. 
Should the person doing the migration do the entries in Weblate?

hr_employee_relative/data/data_relative_relation.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
BT-rmartin, line 1: XML header not needed

hr_employee_relative/README.rst
```diff
@@ -0,0 +1,79 @@
+=====================
```
BT-rmartin, line 1: @pedrobaeza the README file is auto-generated when merging. Should it be included in the migration and adapted or just removed so that later it can be added automatically?

BT-rmartin (changes requested):
@BT-anieto Good job. I still think the code can be improved in terms of legibility with some small changes. Would you mind to do them?

requirements.txt
pedrobaeza: With pre-commit

hr_employee_relative/i18n/de.po
pedrobaeza: Yes, they should stay.

hr_employee_relative/data/data_relative_relation.xml
pedrobaeza: That's not true. Please don't remove them.

hr_employee_relative/README.rst
pedrobaeza: Don't touch it to reduce the diff.

hr_employee_relative/data/data_relative_relation.xml
BT-rmartin: Why is it still needed?

pedrobaeza: Because the full XML compliant is with this header. Having `lxml` as current processing library that doesn't force to use it doesn't mean that is the correct thing. If tomorrow the library is changed (or a new version enforces it), we will need to reintroduce it again. Even more, your principle in migration should be to reduce the diff at maximum, unless the change means a good improvement. And the header also specifies the encoding, which is valuable information in case of weird characters (if some texts are put directly in other non-English language, which is not the case here, but imaging in localization repositories).

luistorresm:
Hello @BT-anieto 

Please apply this task to your MR: (You can see this [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0))

`Squash administrative commits (if any) with the previous commit for reducing commit noise. They are named as "[UPD] README.rst", "[UPD] Update $MODULE.pot", "Update translation files" and similar names, and comes from OCA-git-bot, oca-travis or oca-transbot. IMPORTANT: Don't squash legit translation commits, authored by their translators, with the message "Translated using Weblate (...)".`

hr_employee_relative/data/data_relative_relation.xml
BT-rmartin: thanks for clarifying it. Regarding reducing the diff at maximum, then my suggestions of improvement shouldn't have been done in this PR but in a different one once the migration is merged?

BT-anieto:
@luistorresm Thanks for your review! Is my rebase okay? If it is, I will do it for the other two open PRs

luistorresm (approved):

hr_employee_relative/data/data_relative_relation.xml
pedrobaeza: Cosmetic changes on others' work should be usually prevented, unless you do a big refactorization that touches most part of such code.

BT-rmartin (approved):
LGTM. Good job!

hieulucky111 (approved):
LGTM
Code review

nimarosa (approved):
Code review looks good.

PR #1190, opened 2023-02-13
[15.0][mig]-hr_worked_days_from_timesheet

JuanyDForgeflow:


hr_worked_days_from_timesheet/views/hr_payslip_view.xml
```diff
@@ -0,0 +1,27 @@
...
+                        <button
+                            string="Import from Timesheets"
+                            name="import_worked_days"
+                            type="object"
+                            states="draft"
+                        />
+                    </div>
+                </group>
+            </field>
+            <xpath expr="//sheet/notebook/page[1]/separator[1]" position="replace" />
```
AaronHForgeFlow: Why is this needed? I mean, why replace instead of invisible? I know this comes form previous versions, but I am not sure why it is like this.

AaronHForgeFlow (approved):
Functional review

hr_worked_days_from_timesheet/views/hr_payslip_view.xml
JuanyDForgeflow: I'ts invisible now, the replace was just to hide the original separator

AaronHForgeFlow: ok!

MiquelRForgeFlow (approved):

PR #1167, opened 2022-11-29
[FIX] hr_employee_calendar_planning: no constraint on install mode

yajo:
When another module was installed with demo data but without testing mode, if the demo data includes some `hr.employee` record, it was impossible to install it after `hr_employee_calendar_planning` was installed.

@moduon MT-1622

PR #1191, opened 2023-02-15
[13.0][IMP] hr_employee_calendar_planning: Allow duplicate employee

victoralmau:
Allow duplicate employee.

@Tecnativa TT41700

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -1,5 +1,5 @@
 # Copyright 2019 Tecnativa - Pedro M. Baeza
-# Copyright 2022 Tecnativa - Víctor Martínez
+# Copyright 2022-2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
@@ -45,6 +45,7 @@ class HrEmployee(models.Model):
         comodel_name="hr.employee.calendar",
         inverse_name="employee_id",
         string="Calendar planning",
+        copy=True,
     )
 
     def _regenerate_calendar(self):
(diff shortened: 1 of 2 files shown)


hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -113,6 +114,15 @@ def regenerate_calendar(self):
         for item in self:
             item._regenerate_calendar()
 
+    def copy(self, default=None):
+        self.ensure_one()
+        new = super().copy(default)
+        # We define the first as the main calendar
+        new.resource_id.calendar_id = fields.first(new.calendar_ids).calendar_id
```
pedrobaeza, line 121: Can we avoid this? We are regenerating after the main calendar, and it's too late for filling the required field. I'm also doubting if the regenerate call is needed.

victoralmau: I have tried other possibilities and this is the "_least hard_".

Maybe another option would be that `calendar_id` field of `resource.resource` is not copy, do you want me to investigate a bit more here?

pedrobaeza: OK, if that's the reason, then please clarify it with a good comment. The current one is not clarifying anything. If I'm understanding OK the problem, the comment would be something like `# Define a good main calendar for being able to regenerate it later`.

pedrobaeza (approved):

pedrobaeza:
@yajo maybe this solves your problem in #1167

chienandalu (approved):

yajo:
I don't think so. That problem was happening when installing an addon's demo data, but code here only affects when copying other employees AFAICS.

pedrobaeza:
OK, let's merge this one and fw-port it, and we'll continue checking edge cases:

/ocabot merge patch

PR #1200, opened 2023-02-28
[14.0][FIX] hr_employee_ppe: compute field contains PPE

miguel-S73:
Odoo show an error when you save personal equipment request without lines because `contains_ppe` field becomes empty.

error_compute_field

### DIFF
--- a/hr_employee_ppe/models/hr_personal_equipment_request.py
+++ b/hr_employee_ppe/models/hr_personal_equipment_request.py
@@ -12,12 +12,12 @@ class HrPersonalEquipmentRequest(models.Model):
 
     def _compute_contains_ppe(self):
         for rec in self:
+            contains_ppe = False
             for line in rec.line_ids:
                 if line.is_ppe:
-                    rec.contains_ppe = True
-                    return
-                else:
-                    rec.contains_ppe = False
+                    contains_ppe = True
+                    break
+            rec.contains_ppe = contains_ppe
 
     def action_view_ppe_report(self):
         report = self.env["ir.actions.report"]._get_report_from_name(


marcelsavegnago (approved):
LGTM.

Note: If you could take a look at the test coverage that would be great.

hr_employee_ppe/models/hr_personal_equipment_request.py
```diff
@@ -12,6 +12,9 @@ class HrPersonalEquipmentRequest(models.Model):
 
     def _compute_contains_ppe(self):
         for rec in self:
+            if not rec.line_ids:
```
Saran440:
```suggestion
            rec.contains_ppe = False
            for line in rec.line_ids:
                ...
```
Is it possible to add `rec.contains_ppe = False` first line?
I think it will simple and easy code.

miguel-S73: @Saran440 I have updated my PR with more simple and easy code.
What do you think?

Saran440: @miguel-S73 Can you try it?
```
def _compute_contains_ppe(self):
    for rec in self:
        rec.contains_ppe = bool(rec.line_ids.filtered("is_ppe"))
```

miguel-S73: @Saran440  I am reluctant to use `filtered` specially in a `compute` function. `filtered` method has performance issues.

hr_employee_ppe/models/hr_personal_equipment_request.py
```diff
@@ -12,12 +12,12 @@ class HrPersonalEquipmentRequest(models.Model):
...
     def _compute_contains_ppe(self):
         for rec in self:
+            contains_ppe = False
             for line in rec.line_ids:
                 if line.is_ppe:
-                    rec.contains_ppe = True
-                    return
-                else:
-                    rec.contains_ppe = False
+                    contains_ppe = True
```
Saran440, line 18: I'm not sure, This line should `rec.contains_ppe = True`  or not?

hr_employee_ppe/models/hr_personal_equipment_request.py
Saran440: Ok, I see.

nimarosa (approved):

nimarosa:
/ocabot merge patch 

Merging. Tested on runboat and code looks good.

PR #1207, opened 2023-03-07
[FIX] hr_course_survey: use render template properly

etobella:
With the change to odoo 14, we need to pass the id key, as it returns a dictionary


sanchonuria (approved):

sanchonuria (approved):

kluna1998 (approved):
👍🏽

pedrobaeza (approved):
/ocabot merge patch

PR #1215, opened 2023-03-20
[14.0] [MIG] hr_contract_rate

nimarosa:
Standard Migration


nimarosa:
@pedrobaeza Can you and someone on your team review this migration so i can merge it? I need it in v14 for an implementation for payroll. It's an standard migration so i think it's ready since it has no changes.

pedrobaeza (approved):
OK, we don't use this module, but it seems pretty-straightforward

/ocabot merge nobump

PR #1217, opened 2023-03-30
[FIX] hr_course_survey: URL was not working properly

etobella:
Fields were renamed


nimarosa:
Yes this fix is correct. Merging. 

/ocabot merge patch

PR #1221, opened 2023-04-18
[14.0] [IMP] hr_employee_service: Add helper function

nimarosa:
This PR adds a helper function to return the employee service duration but for a given date. 
This is useful for modules like payroll, when you usually need the service years or months of a employee but calculated to the date of the payslip. 

Currently, this module fields, will return ACTUAL service years, months and days, but if for example you are calculating a March payslip in April, you will get April values, not March. 

This new simple helper function, will help users in payroll to return the values needed based on payslip dates (using `employee.get_service_duration_from_date(payslip.start_date)`) and also could be useful for another scenarios when you need to compute this values for a fixed date. 

Regards.

### DIFF
--- a/hr_employee_service/models/hr_employee.py
+++ b/hr_employee_service/models/hr_employee.py
@@ -104,3 +104,21 @@ class HrEmployee(models.Model):
             return datetime.combine(service_start_date, time(0, 0, 0))
         else:
             return super()._get_date_start_work()
+
+    def get_service_duration_from_date(self, date=None):
+        """
+        Returns the employee service duration for the given date.
+        This function is used in OCA/payroll modules as a helper function
+        to calculate employee service duration calculated for the given date.
+        """
+        self.ensure_one()
+        if not date or not self.service_start_date:
+            return {"years": 0, "months": 0, "days": 0}
+        if date > self.service_start_date:
+            service_duration = relativedelta(date, self.service_start_date)
+            return {
+                "years": service_duration.years,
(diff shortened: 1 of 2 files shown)


nimarosa:
Added tests. 

@pedrobaeza @apps2grow Can you review this for me so i can merge it?

The utility of this new helper function is to calculate the service_duration to a fixed date passed as a param to the function. This is useful for OCA/payroll to be used inside salary rules to use payslip date for the service duration calculation. Without this, if you want to know the employee service duration, it will always return the todays service duration, but in payroll you need it calculated up to the payslip date, which might not be the today's date.

hr_employee_service/models/hr_employee.py
```diff
@@ -104,3 +104,22 @@ def _get_date_start_work(self):
...
+        if not date or not self.service_start_date:
+            return {"years": 0, "months": 0, "days": 0}
+        if date > self.service_start_date:
+            service_duration = relativedelta(date, self.service_start_date)
+            return {
+                "years": service_duration.years,
+                "months": service_duration.months,
+                "days": service_duration.days,
+            }
+        else:
```
pedrobaeza: No need of the `else` clause

pedrobaeza (approved):

hr_employee_service/models/hr_employee.py
nimarosa: You are right. Fixed.

PR #1228, opened 2023-05-05
[15.0][IMP] hr_employee_calendar_planning: Do not set start_date in calendar in test mode to avoid side effects (hr_holidays_natural_period for example).

victoralmau:
Do not set `start_date` in calendar in test mode to avoid side effects (`hr_holidays_natural_period` for example).

@Tecnativa

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -152,22 +152,21 @@ class HrEmployee(models.Model):
         return new
 
     def _sync_user(self, user, employee_has_image=False):
+        """Set calendar_ids from 'Create employee' button from user and set date_start
+        as today is totally correct in a real use case.
+        On the other hand, in test use cases it is problematic to define today in
+        date_start because fixed dates are frequently used.
+        Frequently used case (hr_holidays_natural_period):
+        - Creation of employee linked to a new user.
+        - Creation of a hr.leave record linked to the employee with fixed dates
+        (from 2021-01-02 to 2021-01-05).
+        - The number of days of hr.leave will be 0 because."""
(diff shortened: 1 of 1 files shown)


hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -154,20 +154,11 @@ def copy(self, default=None):
...
-                            "date_start": fields.Date.today(),
-                            "calendar_id": user.company_id.resource_calendar_id.id,
-                        },
-                    ),
-                ]
-            }
-        )
+        calendar_data = {
+            "date_start": fields.Date.today() if not config["test_enable"] else False,
+            "calendar_id": user.company_id.resource_calendar_id.id,
```
pedrobaeza, line 167: Why are you not creating now a calendar line? Isn't inconsistent?

victoralmau: I have added an explanation in the function code for better understanding.

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -152,22 +152,21 @@ def copy(self, default=None):
         return new
 
     def _sync_user(self, user, employee_has_image=False):
+        """Set calendar_ids from 'Create employee' button from user and set date_start
+        as today is totally correct in a real use case.
+        On the other hand, in test use cases it is problematic to define today in
+        date_start because fixed dates are frequently used.
```
pedrobaeza, line 158: If this is the problem, you have to use in the affected tests `freeze_time` from freezegun library instead of doing other weird things.

victoralmau:
Closed in favor of https://github.com/OCA/hr-holidays/pull/73

PR #1224, opened 2023-04-28
[14.0][IMP] hr_employee_calendar_planning: set calendar_ids from Create employee button from user

victoralmau:
Set `calendar_ids` from "_Create employee_" button from user.

Related to: https://github.com/OCA/hr/issues/1163

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa

Issue #1163, opened 2022-11-22
[14.0][hr_employee_calendar_planning] We cannot create employee from a user


cvinh:
## Module
hr_employee_calendar_planning

## Describe the bug
We cannot create an employee from the user form

## To Reproduce
14.0

1. From a user, click on the button "Create employee"
2. User Error : You can not create employees without any calendar.

**Expected behavior**
The employee can be created and we can add planning afterwards

albig:
We noticed the same error in Odoo 15.0.

albig:
I did some investigation on this issue.

First a screen capture of the problem:

![Odoo - Test User (1)](https://user-images.githubusercontent.com/3354500/234517398-4bb4749f-444b-4d24-adee-003c1900ed9e.png)

The check was introduced with the following commits in August 2022:

* Odoo [14.0] https://github.com/OCA/hr/commit/7d3c2fe9a03c35c084f9a2d9806a6adcb4b84758
* Odoo [15.0] https://github.com/OCA/hr/commit/b217ca530712726453d1dd5fd66fdfaedddf03ef

Of course, I understand the intention of this check. Without `hr_employee_calendar_planning`, the employee record is created and the default calendar from `Settings -> Employees -> Work Organization` are assigend.

![Odoo - New](https://user-images.githubusercontent.com/3354500/234519196-4b51aecf-df53-4461-8413-9608db6f0456.png)

IMHO in case of creating employees from users the tasks to do should be:

1. assign default calendar from global settings
2. auto generate calendar by `hr_employee_calendar_planning`

What do you think, @victoralmau, @pedrobaeza?

victoralmau:
Solved with https://github.com/OCA/hr/pull/1224#issuecomment-1532327677

PR #1230, opened 2023-05-16
[15.0][IMP] hr_employee_calendar_planning: remove date_start from _sync_user

jonasbuchholz:
Setting a date_start in the _sync_user method causes some of Odoo's tests to fail as soon as hr_employee_calendar_planning is among the installed modules. This is due to the fact that the users created in tests will not have any planned attendances prior to their create date, which some tests are expecting.
Examples:
```
[odoo.addons.hr_holidays.tests.test_leave_requests:770]

ERROR: TestLeaveRequests.test_archived_allocation
Traceback (most recent call last):
  File "/data/build/odoo/addons/hr_holidays/tests/test_leave_requests.py", line 770, in test_archived_allocation
    leave_2021.with_user(self.user_hrmanager_id).action_approve()
  File "/data/build/odoo/addons/hr_holidays/models/hr_leave.py", line 1130, in action_approve
    self.filtered(lambda hol: not hol.validation_type == 'both').action_validate()
  File "/data/build/odoo/addons/hr_holidays/models/hr_leave.py", line 1142, in action_validate
    raise ValidationError(_('The following employees are not supposed to work during that period:\n %s') % ','.join(leaves.mapped('employee_id.name')))
odoo.exceptions.ValidationError: The following employees are not supposed to work during that period:
 David Employee
```
```
[odoo.addons.project_timesheet_holidays.tests.test_timesheet_holidays:106]

FAIL: TestTimesheetHolidays.test_validate_with_timesheet
Traceback (most recent call last):
  File "/data/build/odoo/addons/project_timesheet_holidays/tests/test_timesheet_holidays.py", line 106, in test_validate_with_timesheet
    self.assertEqual(len(holiday.timesheet_ids), number_of_days, 'Number of generated timesheets should be the same as the leave duration (1 per day between %s and %s)' % (fields.Datetime.to_string(self.leave_start_datetime), fields.Datetime.to_string(self.leave_end_datetime)))
AssertionError: 0 != 3 : Number of generated timesheets should be the same as the leave duration (1 per day between 2018-02-05 07:00:00 and 2018-02-08 07:00:00)
```

Since date_start is not a required field for the calendars, I suggest leaving it empty by default, as to not interfere with the tests.

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -161,7 +161,6 @@ class HrEmployee(models.Model):
                         0,
                         0,
                         {
-                            "date_start": fields.Date.today(),
                             "calendar_id": user.company_id.resource_calendar_id.id,
                         },
                     ),


pedrobaeza:
@victoralmau which was the reason to put that field? I understand that the method is called on every sync or only once? If it's the first one, I'm not sure the rest of the code is correct.

victoralmau:
This was added in v14 at https://github.com/OCA/hr/pull/1224
(related to https://github.com/OCA/hr/issues/1163)

The `_sync_user()` method is called on several sides:
- Create employee from user: https://github.com/odoo/odoo/blob/15.0/addons/hr/models/res_users.py#L255
- Defining the user in the employee: https://github.com/odoo/odoo/blob/15.0/addons/hr/models/hr_employee.py#L265
- When creating (https://github.com/odoo/odoo/blob/15.0/addons/hr/models/hr_employee.py#L289) or updating an employee (https://github.com/odoo/odoo/blob/15.0/addons/hr/models/hr_employee.py#L317)

In a normal use case this process is correct, but in tests, if you don't use mutte_logger you can get unexpected results. I made a similar proposal at https://github.com/OCA/hr/pull/1228

pedrobaeza:
@victoralmau if it's called on "updating an employee", it's incorrect, as you are resetting the start date each time. Imagine you have an employee since 2020. You now update a detail in the employee. The start date of its calendar becomes to today in 2023!!

victoralmau (approved):

pedrobaeza:
Propagate this to all affected branches:

/ocabot merge patch

PR #1238, opened 2023-06-02
[14.0][FIX] hr_employee_calendar_planning: fill calendar if empty

etobella:
We should only fill the calendar with sync_user if and only if you don't have one

Sync user is used also when we modify the user of an employee

@victoralmau as you introduced the functionality, I would appreciate your comments :smile:


victoralmau (approved):

LoisRForgeFlow (approved):
LGTM

PR #1239, opened 2023-06-02
[14.0][IMP] hr_employee_calendar_planning: Allow to archive when calendars have finished

etobella:
Otherwise, we may loose the original calendars of employees if we want to archive old calendars.

Show them on calendar view too in order to check which employees uses the calendar.


nimarosa (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #1242, opened 2023-06-05
[16.0][MIG] hr_employee_calendar_planning

jterolc:
Standard migration to 16.0


hr_employee_calendar_planning/views/hr_employee_views.xml
```diff
@@ -0,0 +1,37 @@
...
+<odoo>
+    <record id="view_employee_form" model="ir.ui.view">
+        <field name="model">hr.employee</field>
+        <field name="inherit_id" ref="hr.view_employee_form" />
+        <field name="arch" type="xml">
+            <field name="resource_calendar_id" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </field>
+            <field name="resource_calendar_id" position="after">
+                <field name="calendar_ids">
```
navarrorico, line 13: @jterolc Hi! Does not seem to be displayed correctly. I tried with a label for the field but doesn't work fine.
image

navarrorico (changes requested):

hr_employee_calendar_planning/views/hr_employee_views.xml
jterolc: The problem is the css rule that is applying a right and left marge to table who is not needed. I can't change this rule because is sourced from core/web...

Can you or someone help to found a clean solution?
Captura desde 2023-06-07 13-17-48

navarrorico: I think that you can put a new page with full width and put there the field. What do you think about that?

victoralmau:
Please, cherry-pick https://github.com/OCA/hr/pull/1253

victoralmau:
Please, cherry-pick https://github.com/OCA/hr/pull/1266

rafaelbn:
@jterolc , please could you review comments and cherry-pick them? Thank you ! 😄

jonasbuchholz:
I have checked out and tested this migration. There seems to be an issue with the global leaves when creating a new employee. 
Steps to reproduce:
- create a new employee with working times that contain any global leave

This is due to the fact the upon creation of the new auto-generated calendar, the global leaves are copied from the original calendar. However, they are initially created with the same field values as the original global leaves, including the calendar_id, before they get linked to the auto-generated calendar. Having overlapping global leaves for the same calendar is prevented by this constrain: https://github.com/odoo/odoo/blob/55cd29c5f2cb7787752d77103dbcc4c4007ff504/addons/hr_holidays/models/resource.py#L15-L33

PR #1225, opened 2023-04-28
[16.0][MIG] hr_employee_service: Migration to 16.0

SoporteOperu:

PR #1243, opened 2023-06-05
[16.0][MIG] hr_employee_service

mmarchpunt:


nimarosa:
Please cherry pick this PR to the migration. 

https://github.com/OCA/hr/pull/1221

So the functionality is included en v16

mmarchpunt:
@nimarosa done

nimarosa (approved):

remi-filament (changes requested):
Hi @mmarchpunt thank your for this migration.
It seems there is already a PR opened for this migration : #1225  (though it does not include #1221).
Also, you did not preserve commit history from previous module which is not the expected way to migrate modules to new versions, since we would lose history, commiter names and we cannot easily compare what has really been modified in v16 (more info here : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0)

mmarchpunt:
@remi-filament Thanks for your input, I will fix those issues

mmarchpunt:
Closing PR

PR #1257, opened 2023-06-14
[14.0][RM] hr_holidays_setting: migrated in wrong repository

### LINKED ISSUE #1256: [14.0] hr_holidays_settings: duplicated in OCA / hr-holidays repo
The repo [OCA / hr-holidays](https://github.com/OCA/hr-holidays) also has an addon called [hr_holidays_settings](https://github.com/OCA/hr-holidays/tree/14.0/hr_holidays_settings), which appears to be functionally identical

## Module
hr_holidays_settings

**Expected behavior**
If both addons are the same, as it appears to be the case, then one of them should be removed, if not, one of them should be renamed

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

JordiMForgeFlow:
Fixes https://github.com/OCA/hr/issues/1256

CC @pedrobaeza @MiquelRForgeFlow @javimurcia @nimarosa @gjotten


pedrobaeza:
For other times, the tag should be `[REM]` (https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message), but it's not so important, and more removing files, so let's fast-track it:

/ocabot merge nobump

PR #1262, opened 2023-06-29
[15.0][ADD] hr_announcement: New module

pilarvargas-tecnativa:
The hr_announcement module extends the announcement module by adding 3 new options to the announcement type. This module allows the creation and management of announcements, which can be sent to specific employees, departments, or job positions.

cc @Tecnativa TT42581

@victoralmau @chienandalu please review :)

ping @CarlosRoca13


hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,106 @@
...
+        ondelete={
+            "employee": "set default",
+            "department": "set default",
+            "job_position": "set default",
+        },
+    )
+
+    employee_ids = fields.Many2many(
+        comodel_name="hr.employee",
+        relation="hr_employee_announcements",
```
chienandalu: Usually we can go with the default relation name for m2m fields unless there are conflicting m2m with same relation in the same model

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,106 @@
...
+        "specific_user_ids",
+        "user_group_ids",
+        "employee_ids",
+        "department_ids",
+        "position_ids",
+    )
+    def _compute_allowed_user_ids(self):
+        res = super()._compute_allowed_user_ids()
+        for announcement in self.filtered(
+            lambda a: a.announcement_type in ["employee", "department", "job_position"]
```
chienandalu, line 69: do a filter per announcement type so you don't have to do the ifs later on

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,106 @@
...
+        ):
+            if announcement.announcement_type == "employee":
+                announcement.allowed_user_ids = announcement.employee_ids.mapped(
+                    "user_id"
+                )
+                announcement.allowed_users_count = len(
+                    announcement.employee_ids.mapped("user_id")
```
chienandalu: You don't need a mapped for chained relations from v13
```suggestion
                    announcement.employee_ids.user_id
```

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,106 @@
...
+                )
+            elif announcement.announcement_type == "department":
+                announcement.allowed_user_ids = (
+                    announcement.department_ids.member_ids.mapped("user_id")
```
chienandalu: Same here
```suggestion
                    announcement.department_ids.member_ids.user_id
```

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @api.depends(
+        "specific_user_ids",
+        "user_group_ids",
+        "employee_ids",
+        "department_ids",
+        "position_ids",
+    )
+    def _compute_allowed_user_ids(self):
+        res = super()._compute_allowed_user_ids()
```
chienandalu: Are you sure super is excluding your new types?: https://github.com/OCA/server-ux/blob/8311de64e311120a758db38cabf8993bb86907cf/announcement/models/announcement.py#L83

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,106 @@
...
+                    "employee_ids.user_id"
+                )
+                announcement.allowed_users_count = len(
+                    announcement.position_ids.mapped("employee_ids.user_id")
+                )
+        return res
+
+    @api.onchange("announcement_type")
+    def _onchange_announcement_type(self):
+        """We want to reset the values on screen"""
```
chienandalu: Your squashing the original method, don't you? (call `super`)

hr_announcement/models/res_users.py
```diff
@@ -0,0 +1,82 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, models
+
+
+class ResUsers(models.Model):
+    _inherit = "res.users"
+
+    @api.model
+    def announcement_user_count(self):
+        # res = super().announcement_user_count()
```
chienandalu: ??

hr_announcement/models/res_users.py
```diff
@@ -0,0 +1,82 @@
...
+                announcement["id"]
+                for announcement in position_announcements
+                if any(
+                    position_id in jobs_with_employees.ids
+                    for position_id in announcement["position_ids"]
+                )
+            ]
+        )
+
+        return super().announcement_user_count() + [
```
chienandalu: This kinda ugly. It's preferable to call to res as you intended and then return res += your_new_results

hr_announcement/models/res_users.py
```diff
@@ -0,0 +1,82 @@
...
+    _inherit = "res.users"
+
+    @api.model
+    def announcement_user_count(self):
+        # res = super().announcement_user_count()
+        employee_announcements = self.env["announcement"].search_read(
+            [
+                ("announcement_type", "=", "employee"),
+                ("in_date", "=", True),
+                ("id", "not in", self.env.user.read_announcement_ids.ids),
```
chienandalu: But you can write to the related employee user the announcement as it's done with the specific_user type so you don't have to go down to this logic: https://github.com/OCA/server-ux/blob/8311de64e311120a758db38cabf8993bb86907cf/announcement/models/announcement.py#L63-L71

hr_announcement/models/res_users.py
```diff
@@ -0,0 +1,82 @@
...
+        department_announcements = self.env["announcement"].search_read(
+            [
+                ("announcement_type", "=", "department"),
+                ("in_date", "=", True),
+                ("id", "not in", self.env.user.read_announcement_ids.ids),
+            ],
+            ["department_ids"],
+        )
+        departments_with_members = self.env["hr.department"].search(
+            [("member_ids.user_id", "!=", False)]
```
chienandalu: Why do you search all the departments? You just need to cross the departments in the department_announcements with the departments the user's employee belongs to.

hr_announcement/models/res_users.py
```diff
@@ -0,0 +1,82 @@
...
+        position_announcements = self.env["announcement"].search_read(
+            [
+                ("announcement_type", "=", "job_position"),
+                ("in_date", "=", True),
+                ("id", "not in", self.env.user.read_announcement_ids.ids),
+            ],
+            ["position_ids"],
+        )
+        jobs_with_employees = self.env["hr.job"].search(
+            [("employee_ids.user_id", "!=", False)]
```
chienandalu: And the same here. Indeed you already have it in the user's employee_id

hr_announcement/readme/USAGE.rst
```diff
@@ -0,0 +1 @@
+To learn about the usage of this module, please refer to the announcement module.
```
chienandalu, line 14: Please develop further :)

chienandalu (changes requested):
Code review

pilarvargas-tecnativa:
> Code review

Thank you for your review above, I have already made the relevant changes. Can you check again if everything is correct?

augustodinizl (approved):
Functional tests on runboat - OK - LGTM
Captura de Tela 2023-07-18 às 18 36 06
Captura de Tela 2023-07-18 às 18 35 52
 
Captura de Tela 2023-07-18 às 18 37 22
Captura de Tela 2023-07-18 às 18 36 46
Captura de Tela 2023-07-18 às 18 36 29

hr_announcement/models/__init__.py
```diff
@@ -0,0 +1 @@
+from . import announcement, res_users
```
zamberjo: As a convention, one import per line

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,108 @@
...
+            ("department", "Department"),
+            ("job_position", "Job Position"),
+        ],
+        ondelete={
+            "employee": "set default",
+            "department": "set default",
+            "job_position": "set default",
+        },
+    )
+
```
zamberjo:
```suggestion
```

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,108 @@
...
+    employee_ids = fields.Many2many(
+        comodel_name="hr.employee",
+        string="Employees",
+        inverse="_inverse_employee_ids",
```
zamberjo: From my lack of knowledge in version 15.0, is it possible to apply an `inverse=`
 to a field that is not `compute=`? 

Wouldn't it be cleaner to launch a trigger in the write/create.

zamberjo (changes requested):
Functionally it seems to work well. Only a few details

hr_announcement/models/announcement.py
pilarvargas-tecnativa: You are right, thank you for your corrections

hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,120 @@
...
+                )
+            ):
+                employee.user_id.unread_announcement_ids |= announcement
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        announcements = super().create(vals_list)
+        for vals in vals_list:
+            if vals.get("employee_ids"):
+                announcements._update_read_unread_announcements()
```
zamberjo: By using `model_create_multi` you would be running it for all announcements with only 1 value having `employee_ids`.

```suggestion
        announcements = super().create(vals_list)
        for index, vals in enumerate(vals_list):
            if vals.get("employee_ids"):
                announcements[index]._update_read_unread_announcements()
```

Or you can directly use `@api.model` in a unitary way.

pilarvargas-tecnativa: I still have a lot to learn but looking at the documentation I have learned the difference between both decorators, thanks.

zamberjo: Don't worry, in this world we are all learning all the time.

zamberjo (approved):
The `super` that @chienandalu points out is missing, otherwise it looks good.

pilarvargas-tecnativa:
> The `super` that @chienandalu points out is missing, otherwise it looks good.

thanks!!!

hr_announcement/models/announcement.py
chienandalu: You're still redefining the method without calling super()

rvalyi:
Hello, yesterday I tried the announcement module on the v15 Runboat and got some issues: basically as the admin I sent an announcement for the demo user and the I had a demo user session in an anonymous browser window.

Problem: the demo user didn't get the notification in the tray unless he reloaded the page. Also I got polling XHR timeouts. Also for the admin user he got the notification in the tray when the announcement was for him, but he had to refresh the page to get the popup. I don't know if this is may be related to the Runboat environment (possibly creating the XHR polling timeouts) or if I did something wrong, but I think this is worth a check on your side.

pedrobaeza:
@rvalyi bus is not working in runboats unless you put https on your URLs.

rvalyi:
> @rvalyi bus is not working in runboats unless you put https on your URLs.

ah ok thank you for clarifying.

pedrobaeza:
AFAIK, you can type `https://` in your browser when using a runboat instance, and a self-signed certificate will arise. Accepting the warning, you can then try it.

hr_announcement/models/announcement.py
pilarvargas-tecnativa: Done, thanks

hr_announcement/models/res_users.py
pilarvargas-tecnativa: Sorry I hadn't pushed all the changes.

gurneyalex (approved):
Looks good. An automated test would make it great.

chienandalu (approved):
Working in prod for months. You can mention the lack of tests in the roadmap though

PR #1263, opened 2023-06-29
[13.0][IMP] hr_employee_calendar_planning: Set calendar_ids default value to cover all use cases

victoralmau:
Set calendar_ids default value to cover all use cases.

Steps to reproduce the error (before):
- Go to Recruitment > Apllications and enter one of them.
- Click on the "_Create employee_" button and confirm.

An error is displayed when trying to create the employee because there is no calendar defined.

@Tecnativa TT44093

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -48,6 +48,16 @@ class HrEmployee(models.Model):
         copy=True,
     )
 
+    @api.model
+    def default_get(self, fields):
+        """Set calendar_ids default value to cover all use cases."""
+        vals = super().default_get(fields)
+        if "calendar_ids" in fields and not vals.get("calendar_ids"):
+            vals["calendar_ids"] = [
+                (0, 0, {"calendar_id": self.env.company.resource_calendar_id.id}),
+            ]
+        return vals
+
     def _regenerate_calendar(self):
         self.ensure_one()
         vals_list = []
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

pedrobaeza:
I don't think injecting specific sources is the right patch. Instead, a default should exist for covering all cases.

pedrobaeza:
With this you can remove also the `_synchronize_user` code, isn't it?

victoralmau:
Changes done and test fixed

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -48,6 +48,21 @@ class HrEmployee(models.Model):
         copy=True,
     )
 
+    @api.model
+    def default_get(self, fields):
+        """Set calendar_ids default value to cover all use cases."""
+        vals = super().default_get(fields)
+        if (
+            not config["test_enable"]
```
pedrobaeza: Why do you need this condition?

victoralmau: The change is necessary to avoid errors in the tests of other modules (https://github.com/OCA/hr/actions/runs/5441518693/jobs/9895568077#step:8:379).

pedrobaeza: But this is because there's a missing `ondelete="cascade"` here:

https://github.com/OCA/hr/blob/6a0b62b59bc349c85a06faadc70edfc4976d718e/hr_employee_calendar_planning/models/hr_employee.py#L163

and `ondelete="restrict"` here:

https://github.com/OCA/hr/blob/6a0b62b59bc349c85a06faadc70edfc4976d718e/hr_employee_calendar_planning/models/hr_employee.py#L166

victoralmau:
Changes done.

pedrobaeza (approved):
/ocabot merge patch

PR #1265, opened 2023-07-05
[14.0][IMP] hr_employee_calendar_planning: Set calendar_ids default value to cover all use cases

victoralmau:
FWP from 13.0: https://github.com/OCA/hr/pull/1263

Set `calendar_ids` default value to cover all use cases.

Steps to reproduce the error (before):
- Go to Recruitment > Apllications and enter one of them.
- Click on the "_Create employee_" button and confirm.

An error is displayed when trying to create the employee because there is no calendar defined.

@Tecnativa TT44093

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -187,9 +187,7 @@ class HrEmployeeCalendar(models.Model):
         string="End Date",
     )
     employee_id = fields.Many2one(
-        comodel_name="hr.employee",
-        string="Employee",
-        required=True,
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why you put that "This change is not necessary in v14"? Isn't this v14 branch?

victoralmau:
Sorry, that comment really isn't correct with the current approach (I've removed it and also in the v13 PR to avoid confusion). Although now the employee is NOT created directly and all the fields are defined to be able to create it (with the form view) these changes are necessary so that the `calendar_ids` field is defined correctly by default.

pedrobaeza (approved):
Fast-tracking for equaling branches:

/ocabot merge patch

PR #1268, opened 2023-07-13
[15.0][IMP] hr_employee_age: Add groups="hr.group_hr_user" to age field according to birthday field definition

victoralmau:
Add `groups="hr.group_hr_user"` to age field according to `birthday` field definition (https://github.com/odoo/odoo/blob/15.0/addons/hr/models/hr_employee.py#L73C45-L73C70)

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT43046

### DIFF
--- a/hr_employee_age/models/hr_employee.py
+++ b/hr_employee_age/models/hr_employee.py
@@ -8,7 +8,11 @@ from odoo import api, fields, models
 class HrEmployee(models.Model):
     _inherit = "hr.employee"
 
-    age = fields.Integer(compute="_compute_age")
+    # The age field uses a depends (birthday) that has defined
+    # groups="hr.group_hr_user", if a user without permissions in HR tries to get
+    # the value of this field will have an error.
+    # The correct way to avoid this inconsistency is to define groups to field age
+    age = fields.Integer(compute="_compute_age", groups="hr.group_hr_user")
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Indicate as a comment why this is needed, copying the one from upstream. If not, it's not very understandable why do you need this.

hr_employee_age/models/hr_employee.py
```diff
@@ -8,7 +8,8 @@
 class HrEmployee(models.Model):
     _inherit = "hr.employee"
 
-    age = fields.Integer(compute="_compute_age")
+    # Add groups="hr.group_hr_user" to age field according to birthday field definition
```
pedrobaeza: This is not adding value as explanation. It doesn't say the why, just the what, which is pretty obvious reading the code.

victoralmau: Comment modified.

pedrobaeza: OK, thank you. Maybe is now too much verbose, as it can be put simply as `# Group is needed as the dependency field `birthday` already has it`, but no special problem in being longer.

CarlosRoca13 (approved):

PR #1271, opened 2023-07-21
[15.0][FIX] hr_employee_document: Remove record rules from hr.employee to avoid side effects..

victoralmau:
Remove record rules from hr.employee to avoid side effects.

**Example of use case**:
- An employee X is created and a Manager Y (`parent_id`) is defined.
- User X creates an expense report (`hr.expense.sheet`).
- The "_Submit to manager_" button is clicked.
- A permissions error is displayed when accessing employee Y (`hr.employee`) because the record rules that this module had added do not allow access to that record.

**Solution**:
This change overrides some things to allow the user's employee attachments to be displayed.

@Tecnativa TT44536


pedrobaeza:
You have to put the steps to reproduce the problem in the commit message.

Apart from that, why that ACLs was there? (isn't they record rules, not ACL?) Seeing a bit the code, it seems you are solving it acting at code level. Please explain all of this in the commit message, docstring, etc.

victoralmau:
Sorry, changed the commit message and explained a use case.

hr_employee_document/models/ir_attachment.py
```diff
@@ -0,0 +1,67 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class IrAttachment(models.Model):
+    _inherit = "ir.attachment"
+
+    @api.model
+    def _search(
```
pedrobaeza: Security shouldn't be handled in `_search` method. This is a security hole, as you can browse directly some random records.

pedrobaeza (changes requested):

victoralmau:
I have changed the approach to try to solve the problem in another way, although there is still some detail to be solved, what do you think about this way?

hr_employee_document/models/hr_employee.py
```diff
@@ -22,15 +23,33 @@ def _compute_document_count(self):
...
+        """Return access to the hr.employee model if we pass a specific context,
+        is a trick to list the attachments related to an employee."""
+        if (
+            not self.env.is_superuser()
+            and not self.env.user.has_group("hr.group_hr_user")
+            and operation == "read"
+            and self._name == "hr.employee"
+        ):
+            if (
+                self.env.context.get("search_attachments_from_hr_employee")
```
pedrobaeza, line 37: Shouldn't this be limited to the specific employee?

hr_employee_document/models/ir_rule.py
```diff
@@ -0,0 +1,38 @@
...
+            "mode",
+            "tuple(self._compute_domain_context_values())",
+        ),
+    )
+    def _compute_domain(self, model_name, mode="read"):
+        """We need to add for security purposes an extra domain in the hr.employee
+        model to restrict only the user's employees when search employee attachments."""
+        res = super()._compute_domain(model_name, mode=mode)
+        user = self.env.user
+        if model_name == "hr.employee" and not self.env.su:
```
pedrobaeza: Shouldn't this also exclude HR managers?

victoralmau: Now the extra domain is not added to the rule if the user has the Employees > Officer permission (this is the same behavior as the module currently has).

pedrobaeza: But our use case wants to exclude Officers. We want this only for HR managers, not officers.

pedrobaeza (approved):

chienandalu (approved):
LGTM :+1:

PR #1290, opened 2023-11-07
[16.0][MIG] hr_employee_calendar_planning

jonasbuchholz:
Module migration, since there hasn't been any activity on #1242 for some time. Includes the commits mentioned by @victoralmau and a fix for the bug when creating new employees. For better formatting, the calendar_ids field is placed in a separate notebook page in the employee view


hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
```diff
@@ -0,0 +1,454 @@
...
+            context=dict(
+                cls.env.context,
+                mail_create_nolog=True,
+                mail_create_nosubscribe=True,
+                mail_notrack=True,
+                no_reset_password=True,
+                tracking_disable=True,
+                test_hr_employee_calendar_planning=True,
+            )
+        )
```
victoralmau:
```suggestion
from ..hooks import post_init_hook
from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT


class TestHrEmployeeCalendarPlanning(common.TransactionCase):
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
```

victoralmau (changes requested):
Some comments:

- It is not clear to me that adding the calendars on a new page could not have side effects.
- Check the tests and make the necessary fixes.

hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
jonasbuchholz: thanks, that is indeed more convenient

hr_employee_calendar_planning/views/resource_calendar_views.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2023 Creu Blanca
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="resource_calendar_form" model="ir.ui.view">
+        <field name="model">resource.calendar</field>
+        <field name="inherit_id" ref="resource.resource_calendar_form" />
+        <field name="arch" type="xml">
+            <notebook position="inside">
+                <page name="employee_calendar" string="Employee Calendars">
```
victoralmau, line 10: Adding a new tab doesn't quite convince me, maybe you could move `employee_calendar_ids` field to another place (inside the tab it was previously in) to be the full size of the tab. What do you think @pedrobaeza ?

pedrobaeza: Yes, a whole new tab just for this seems over-killing.

jonasbuchholz: What do you think about leaving the label and setting a colspan? Looks like this:
employee_calendar_planning_form

pedrobaeza: It seems it overflows a bit the limits.

jonasbuchholz: Yes, I know what you mean. But to me, this seems like the way Odoo is treating fields like this in views, see eg [here](https://github.com/odoo/odoo/blob/d4b366f2d741f2087a947d61d7e0618492cc2bf1/addons/account/views/partner_view.xml#L207). Defining custom css just for this view seems a bit unnecessary. Alternatively, we could place it in a sheet, like this:
image

pedrobaeza: OK, that can be another option.

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge nobump
/ocabot migration hr_employee_calendar_planning

PR #1304, opened 2023-11-14
[17.0][MIG] hr_employee_ssn : Migration to 17.0

manishkumarbohra:
Module Name : HR Employee SSN & SIN (hr_employee_ssn)

Migration to 17.0

For Issue #1291


pedrobaeza:
Now you are missing the other commit. Both should be here, the pre-commit one and the migration one. Please check guide and follow technical steps.

PR #1308, opened 2023-11-14
[17.0][MIG] hr_employee_ssn : Migration to 17.0

manishkumarbohra:
Module Name : HR Employee SSN & SIN (hr_employee_ssn)

Migration to 17.0

For Issue #1291


pedrobaeza:
Wrong final commit message, and you are changing the manifest version in the pre-commit one. I think you are not following strictly the guide, where it states that the "pre-commit fixes" commit is invoked with `--no-verify` for avoiding the linter errors.

manishkumarbohra:
@pedrobaeza  now, I've read the doc and now i understand the process and i will follow all the rules strictly.
also, i've update the commit.

Thanks for feedback.

PR #1314, opened 2023-11-16
[17.0][MIG] hr_employee_firstname: Migration to 17.0

andreagidaltig:
- Update definition of required modifier as part of [1].
- Update incoming parameters of post_init_hook method since `cr, registry`
  were replaced by `env` in [2].
- Deprecate the use of _onchange_spec method and use the method new method
  _get_fields_spec to return the fields specification from a view
  description since it was introduced in [3].

[1] https://github.com/odoo/odoo/pull/104741
[2] https://github.com/odoo/odoo/commit/b4a7996
[3] https://github.com/odoo/odoo/commit/f5e6494


andreagidaltig:
Hi @luisg123v could you review this please?

cc. @luistorresm

hr_employee_firstname/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2010-2014 Savoir-faire Linux (<http://www.savoirfairelinux.com>)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "HR Employee First Name, Last Name",
+    "version": "17.0.1.0.1",
```
luisg123v:
```suggestion
    "version": "17.0.1.0.0",
```

luisg123v (changes requested):

hr_employee_firstname/__manifest__.py
andreagidaltig: I just updated this  👍

luisg123v (approved):
LGTM :+1:

PR #1315, opened 2023-11-16
[17.0][MIG] hr_employee_lastnames: Migration to 17.0

andreagidaltig:
- Rename module from hr_employee_lastnames → hr_employee_second_lastname
  in order to be consistent with partner_second_lastname module.
- Fix some new lints.
- Update incoming parameters of post_init_hook method since `cr, registry`
  were replaced by `env` in [1].
- Introduce the use of new onchange method [2] (renamed in [3]) to
  return the fields specification from a view description.

[1] https://github.com/odoo/odoo/commit/b4a7996
[2] https://github.com/odoo/odoo/commit/f5e6494


andreagidaltig:
Hi @luisg123v could you review this please?

cc. @luistorresm

FernandoRomera:
@andreagidaltig 
Could we change the name of the module to synchronize it with partner_second_lastname?.

Then "hr_employee_lastnames" -> "hr_employee_second_lastname"

hr_employee_lastnames/models/hr_employee.py
```diff
@@ -0,0 +1,160 @@
...
+            if "firstname" in vals:
+                firstname = vals["firstname"]
+            else:
+                firstname = self.firstname
+            if "lastname2" in vals:
+                lastname2 = vals["lastname2"]
+            else:
+                lastname2 = self.lastname2
+            vals["name"] = self._get_name_lastnames(lastname, firstname, lastname2)
+        elif vals.get("name"):
```
FrancoMaxime: Hi @andreagidaltig, i created a PR fixing some issues in the v14 module ( #1283) I could reproduce the bug on the runboat v17. 
To reproduce you can create an employee with a first name equals to " Name 1 Name 2 name 3" and a  lastname equals to "name4"  if you export the data (including the external_id) and importing again the data, the name will be changed. 

Could you fix it or cherry pick my commit? Thanks

FrancoMaxime (changes requested):

hr_employee_lastnames/models/hr_employee.py
luisg123v: @FrancoMaxime since your PR is not merged yet, it shouldn't be included in the migration (because it's still subject to changes).

luisg123v:
@FernandoRomera 

> Could we change the name of the module to synchronize it with partner_second_lastname?.

Not sure if that's a good idea. That will "break" systems being migrated from previous versions, because the old module would result uninstallable and the new one not installed.

hr_employee_second_lastname/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import models
+from .hooks import post_init_hook
```
luisg123v, line 2: Could you rename hooks file in the previous version, please?

luisg123v (changes requested):

FernandoRomera:
> @FernandoRomera
> 
> > Could we change the name of the module to synchronize it with partner_second_lastname?.
> 
> Not sure if that's a good idea. That will "break" systems being migrated from previous versions, because the old module would result uninstallable and the new one not installed.

@luisg123v 
Don't worry, there is already a procedure for this in OpenUpgrade.

If you want we can propose the PR in OpenUpgrade.



apriory.py in OpenUpgrade

Renamed modules is a mapping from old module name to new module name
renamed_modules = {
    ...
    # OCA/hr
    "hr_employee_lastnames": "hr_employee_second_lastname",
    ...

This PR is a example: https://github.com/OCA/OpenUpgrade/pull/4246/files

luisg123v:
Got it. I think module renaming should be on a separate commit, though, before running `pre-commit`

@moylop260 what do you think about renaming the module?

moylop260:
I'm not a fan to rename modules

But I don't have the enough context to do it

Consider that you will need to change all the xmlids and you will need to change all the depends from the manifest of modules using this module

FernandoRomera:
> Don't worry, there is already a procedure for this in OpenUpgrade.

@moylop260
Don't worry, there is already a procedure for all changes in OpenUpgrade.



> I'm not a fan to rename modules
> 
> But I don't have the enough context to do it
> 
> Consider that you will need to change all the xmlids and you will need to change all the depends from the manifest of modules using this module

@moylop260
Don't worry, there is already a procedure for all changes in OpenUpgrade.

luisg123v:
> Don't worry, there is already a procedure for all changes in OpenUpgrade.

Not for this:

> and you will need to change all the depends from the manifest of modules using this module

:smile:

FernandoRomera:
> > Don't worry, there is already a procedure for all changes in OpenUpgrade.
> 
> Not for this:
> 
> > and you will need to change all the depends from the manifest of modules using this module
> 
> 😄

You are right, but now is the time to do it and make the nomenclature of the modules consistent.

in the first line of the commit you indicate that you synchronize with the module partner_second_lastname

Therefore partner_contact is like this:
partner_firstname
partner_second_lastname

and in hr
hr_employee_firstname
hr_employee_second_lastname

This is my last comment for it, thanks for all the work done.

andreagidaltig:
Hi @luisg123v could you review this, please? I just added the commit to rename the module.

hr_employee_second_lastname/__init__.py
luisg123v, line 1: Commit message:
- AFAIK tag for module renaming is `[MOV]`
- Commit message doesn't mention where the module is being renamed from. It should include both from and to, e.g.:
  `Module is renamed from hr_employee_lastnames → hr_employee_second_lastname inorder to be consistent with...`

hr_employee_second_lastname/__manifest__.py
```diff
@@ -1,7 +1,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
luisg123v, line 1: Commit message:
- sicronize -> synchronize. Besides, this would only go in PR description because it's not done on this commit anymore
- "Rename hooks file to follow guidelines.": not done on this commit anymore. This was done on v16, so not part of the migration anymore (hence should be removed also from PR's description).

hr_employee_second_lastname/tests/test_hr_employee_lastnames.py
```diff
@@ -1,10 +1,11 @@
 from odoo import exceptions
 from odoo.tests.common import TransactionCase
+from odoo.tools import submap
 
 
 class TestEmployeeLastnames(TransactionCase):
     def setUp(self):
-        super(TestEmployeeLastnames, self).setUp()
+        super().setUp()
```
luisg123v: Document in commit description

hr_employee_second_lastname/hooks.py
```diff
@@ -0,0 +1,24 @@
...
+    # set name fields correctly.
+    #
+    # After the installation, previously the dependency hr_employee_firstname
+    # splitting the name into two parts: firstname and lastname, so for this
+    # module to be able to process the new field lastmane2 it is necessary to
+    # reset the values to empty to be able to correctly set the three fields
+    # (firstname, lastname and lastname2).
+    #
+    # For example:
+    #  After install hr_employee_fisrtname and before install hr_employee_lastnames:
```
luisg123v: Missing renaming module here

hr_employee_second_lastname/hooks.py
```diff
@@ -0,0 +1,24 @@
...
+    #     name = 'John Peterson Clinton'
+    #     firstname = 'John'
+    #     lastname = 'Peterson Clinton'
+    #
+    #  After install hr_employee_lastnames:
+    #     name = 'John Peterson Clinton'
+    #     firstname = 'John'
+    #     lastname = 'Peterson'
+    #     lastname2 = 'Clinton'
+    env.cr.execute("UPDATE hr_employee SET firstname = NULL, lastname = NULL")
```
luisg123v, line 22: Missing renaming module here

hr_employee_second_lastname/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (C) 2014 Savoir-faire Linux. All Rights Reserved.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from . import test_hr_employee_lastnames
```
luisg123v: And here.

Please fix all similar cases.

luisg123v (changes requested):

andreagidaltig:
Hi @luisg123v changes applied, could you review again, please?

hr_employee_second_lastname/tests/test_hr_employee_second_lastname.py
```diff
@@ -0,0 +1,173 @@
+from odoo import exceptions
+from odoo.tests.common import TransactionCase
+from odoo.tools import submap
+
+
+class TestEmployeeLastnames(TransactionCase):
```
luisg123v: You forgot to update lastnames -> second lastname here.

andreagidaltig: I just updated this, could you review again, please?

luisg123v (approved):
LGTM :+1: 

@moylop260 could you review/merge, please?

hr_employee_lastnames/models/hr_employee.py
FrancoMaxime: @luisg123v Thanks, I hadn’t thought of that!

PR #1317, opened 2023-11-20
[10.0][FIX] hr_holidays_compute_days: get_work_days_count() throws error if no employee is selected

AungKoKoLin1997:
Recreate PR that was closed https://github.com/OCA/hr/pull/1229

@qrtl

### DIFF
--- a/hr_holidays_compute_days/models/hr_holidays.py
+++ b/hr_holidays_compute_days/models/hr_holidays.py
@@ -162,7 +162,8 @@ class HrHolidays(models.Model):
                 include_rest_days=not self.holiday_status_id.exclude_rest_days,
                 compute_full_days=self.holiday_status_id.compute_full_days,
             )
-            days = employee.get_work_days_count(
-                from_datetime=date_from, to_datetime=date_to,
-            )
-            self.number_of_days_temp = days or 0
+            if employee:
+                days = employee.get_work_days_count(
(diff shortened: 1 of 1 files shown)


kanda999 (approved):
Functional review: It worked as expected.

AaronHForgeFlow (approved):

hr_holidays_compute_days/models/hr_holidays.py
```diff
@@ -162,7 +162,8 @@ def _onchange_data_hr_holidays_compute_days(self):
                 include_rest_days=not self.holiday_status_id.exclude_rest_days,
                 compute_full_days=self.holiday_status_id.compute_full_days,
             )
-            days = employee.get_work_days_count(
-                from_datetime=date_from, to_datetime=date_to,
-            )
-            self.number_of_days_temp = days or 0
+            if employee:
```
FrancoMaxime, line 165: In the case of no employee is found, can we put a default value like 0 ?
Like set days equal to 0, then if there is an employee make some computation then in all case set number_of_days_temp = days 

What do you think about this approach?

AungKoKoLin1997: > In the case of no employee is found, can we put a default value like 0 ?

I don't think we need it because `number_of_days_temp` is float field and default is already `0.00`.

PR #1342, opened 2024-04-09
[16.0][FIX]hr_employee_calendar_planning:create multi calendars

ChristophAbenthungCibex:
Changed the create method of the model hr.employee.calendar to work the same as the write and unlink methods.
Reason: when creating multiple records of this model at the same time the method  _regenerate_calendar() will be called with multiple records. So the ensure_one() will throw an error.
The same happens when trying to create multiple employees.


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1355, opened 2024-05-23
[17.0][MIG] hr_employee_calendar_planning

pedro-gonzalez-aulora:
Work done:

- [x] Adapting __manifest__.py version
- [x] Adapt employee public view as resource_calendar_id is not present on it  anymore
- [x] On tests assert on number of attendances changed slightly because of the lunch break
- [x] On create redefinition adapted the employee calendar regeneration to avoid a singleton Value Error
- [x] Fix ruff B023, UP008, E501


juanjosesegui-tecnativa:
@pedrobaeza

juanjosesegui-tecnativa (approved):
TT50117

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -185,7 +185,8 @@ def create(self, vals_list):
             and res.filtered(lambda x: not x.calendar_ids)
         ):
             raise UserError(_("You can not create employees without any calendar."))
-        res.filtered("calendar_ids").regenerate_calendar()
+        for employee in res.filtered("calendar_ids"):
+            employee.sudo().regenerate_calendar()
```
pedrobaeza, line 189: Why changing this to one record calls?

hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
```diff
@@ -170,10 +170,11 @@ def test_calendar_planning(self):
...
         self.assertTrue(self.employee.resource_calendar_id.two_weeks_calendar)
 
-        # Calendar 1 has 20 lines + Calendar 2 has 6 lines that are duplicated
-        # in the odd and even week + even week label + odd week label
+        # Calendar 1 has 30 lines including lunch times
+        # Calendar 2 has 6 lines that are duplicated in the odd and even week
+        # + even week label + odd week label
         self.assertEqual(
-            len(self.employee.resource_calendar_id.attendance_ids), 20 + 6 * 2 + 2
+            len(self.employee.resource_calendar_id.attendance_ids), 30 + 6 * 2 + 2
```
pedrobaeza: Why all these number changings? Can you please avoid change them? The test should be immutable as a check of the functionality is preserved (unless something involved with data initialization itself).

pedro-gonzalez-aulora: Absolutely. The idea of a unitary test is to protect the technical evolution of the module. In this case the attendance_ids was changed because of the inclusion of the breaks... I do come back to you on Wednesday with concrete details of it as I did this work some time ago.

hr_employee_calendar_planning/models/hr_employee.py
pedro-gonzalez-aulora: @pedrobaeza as you see in my PR description this change is to avoid the singleton error. If you find another better way just let me know

About the test changes I also made a comment on the description of the work done: "On tests assert on number of attendances changed slightly because of the lunch break"

Let me know if you need further explanations

pedrobaeza: OK, seen. I just saw the diff and feel weird to change this without any other modification, but it seems the potential problem has been there all of this time.

pandoc-3.2-1-amd64.deb
pedrobaeza, line 1: Please remove this file. This can't be merged with this file here.

pedrobaeza: And please remove it in the commit that is adding it (the "pre-commit auto fixes" one).

hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
pedrobaeza: I see that we are creating the calendar deterministically, so the number should be fixed. Maybe this version is adding calendar attendances entries automatically that changes this number, that you should remove or prevent to be created, or maybe you need to filter according the criteria to get the same number.

pedro-gonzalez-aulora: @pedrobaeza did a pre-filtering of those attendances

pandoc-3.2-1-amd64.deb
pedro-gonzalez-aulora: @pedrobaeza removed

pedrobaeza (approved):
Great, I think the tests and the diff is clearer now for knowing why the changes.

Thanks for all.

/ocabot merge nobump

PR #1359, opened 2024-06-07
[16.0][ADD] hr_employee_phone_pin module

arielbarreiros96:
This module adds the following to employees:
* Mobile phone PIN
* Mobile phone PUK


adasatorres (approved):

antoniodavid (approved):
LGTM

etobella:
@arielbarreiros96 The code itself is right, however, I wanted to raise some functionality concerns.

Usually, Phones are not managed by HR, are managed by IT (at least on my customers). I think that IT should not access HR data, because it is senstive data, and we want to avoid issues there. At least this is what happens in my experience. In that companies, they use maintenance.equipments for phone numbers. It is a good approach because the "owner" of the phone number might change. Also, it is recommended to have an special group just for this two fields, becuase we want to avoid access to this data for other teams or less experienced people.

WDYT?

arielbarreiros96:
> @arielbarreiros96 The code itself is right, however, I wanted to raise some functionality concerns.
> 
> Usually, Phones are not managed by HR, are managed by IT (at least on my customers). I think that IT should not access HR data, because it is senstive data, and we want to avoid issues there. At least this is what happens in my experience. In that companies, they use maintenance.equipments for phone numbers. It is a good approach because the "owner" of the phone number might change. Also, it is recommended to have an special group just for this two fields, becuase we want to avoid access to this data for other teams or less experienced people.
> 
> WDYT?

Hi @etobella I have just added a security group to restrict access to this data by users outside this group, which will indeed be useful as some HR users are not technically savvy. On the other hand, I understand your concern about access to HR data by technical staff, in this situation the feature makes sense for a client who do not use maintenance itself, but only want to keep the PIN/PUK of the company phone assigned to the employee. 

According to what you mention, it would be nice to make a similar functionality for the maintenance repo covering the other use case.

hr_employee_phone_pin/views/hr_employee_views.xml
```diff
@@ -6,8 +6,14 @@
...
-                <field name="mobile_phone_pin" groups="hr.group_hr_user" />
-                <field name="mobile_phone_puk" groups="hr.group_hr_user" />
+                <field
+                    name="mobile_phone_pin"
+                    groups="hr_employee_phone_pin.group_mobile_phone_pin_manager"
+                />
+                <field
+                    name="mobile_phone_puk"
+                    groups="hr_employee_phone_pin.group_mobile_phone_pin_manager"
+                />
```
mtelahun: I'm wondering if maybe there should be an additional security check to allow the employee him/herself at least read access to this field.

etobella: Well, Actually, employees don't have access to hr.employee view, so we should need to do something different :thinking: However, I think that this info is more related to IT

arielbarreiros96: > Well, Actually, employees don't have access to hr.employee view, so we should need to do something different 🤔 However, I think that this info is more related to IT

@mtelahun @etobella I´ve reviewed the approach in order to make the PIN/PUK visible from the hr.employee.public form, in a way that the employee can only see its own PIN/PUK, and only if it belongs to a specific group (group_mobile_phone_pin_user). I´ve done it this way because hr.employee form view is only available for hr.group_hr_user group users, or any other group which inherits hr.group_hr_user.

On the other hand I understand that this info is related to IT in several ways, but unless a new group inherits hr.group_hr_user the hr.employee form wont be visible (at least I don't know hot to make it available 😅 with limited permissions), which in my opinion shows way too much private information for any user outside HR. Also, in this particular scenario we need the feature without having maintenance involved.

Any ideas are appreciated 👌

hr_employee_phone_pin/models/hr_employee_public.py
```diff
@@ -0,0 +1,23 @@
...
+    mobile_phone_puk = fields.Char(
+        related="employee_id.mobile_phone_puk", readonly=True
+    )
+
+    is_current_user = fields.Boolean(compute="_compute_is_current_user")
+
+    @api.depends("user_id")
+    def _compute_is_current_user(self):
+        for rec in self:
+            rec.is_current_user = rec.user_id == self.env.user
```
mtelahun, line 23: Is it necessary for these fields to be in the employee's public profile?

arielbarreiros96: I could remove the related property from pin and puk, but they are related to the ones on hr.employee anyway 😅. Also tried to remove them just to be sure, and it fails to install the app if they are not on the public model.
On the other hand is_current_user  is required to be on the view since pin and puk are only visible when is_current_user is true. This logic only happens on the public form

mtelahun:
LGTM

nimarosa (approved):

PR #1372, opened 2024-06-25
[13.0][FIX] hr_employee_document: Remove record rules from hr.employee to avoid side effects

victoralmau:
Backport from 14.0: https://github.com/OCA/hr/pull/1320

Remove record rules from hr.employee to avoid side effects.

**Example of use case**:
- An employee X is created and a Manager Y (`parent_id`) is defined.
- User X creates an expense report (`hr.expense.sheet`).
- The "_Submit to manager_" button is clicked.
- A permissions error is displayed when accessing employee Y (`hr.employee`) because the record rules that this module had added do not allow access to that record.

**Solution**:
This change overrides some things to allow the user's employee attachments to be displayed.

@Tecnativa TT49736


pedrobaeza (approved):

kaynnan (approved):
LGTM

PR #1374, opened 2024-07-01
[13.0][FIX] hr_employee_document: Correct condition if there are several employees for one user

victoralmau:
Correct condition if there are several employees for one user

@Tecnativa TT49830


pedrobaeza (approved):
Fast-tracking this straight patch:

/ocabot merge patch

PR #1376, opened 2024-07-01
[15.0][FIX] hr_employee_document: Correct condition if there are several employees for one user

victoralmau:
FWP from 14.0: https://github.com/OCA/hr/pull/1375

Correct condition if there are several employees for one user

@Tecnativa TT49830


pedrobaeza (approved):
/ocabot merge patch

PR #1391, opened 2024-09-10
[16.0][REM] hr_holidays_settings: Wrong repo

pedrobaeza:
It should be in OCA/hr-holidays.


gjotten (approved):
Thanks for the quick response.

JordiBForgeFlow:
Thanks @pedrobaeza

PR #1401, opened 2024-11-07
[16.0][MIG] hr_contract_rate: Migration to 16.0

ntsirintanis:
Standard migration from 14.0 to 16.0


hr_contract_rate/models/hr_contract.py
```diff
@@ -12,7 +12,6 @@ class HrContract(models.Model):
         inverse="_inverse_wage",
         store=True,
         required=False,
-        track_visibility=None,
```
dreispt: This is removing the tracking fature from the wage field. Can we keep it?

ntsirintanis: Thanks for the notification, missed that!

PR #1415, opened 2024-12-13
[18.0][MIG] hr_job_category: Migration to 18.0

BhaveshHeliconia:


hr_job_category/models/hr.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+        for employee in self.mapped("employee_id"):
+            tags_to_add = [
+                (4, tag.id) for tag in job.category_ids - employee.category_ids
+            ]
+            if tags_to_add:
+                employee.write({"category_ids": tags_to_add})
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
luisDIXMIT, line 58: Since the create method in version 18 receives a list of vals as a parameter, it must be iterated to validate each record. Therefore, the method must be written as follows, otherwise it won't work.
```suggestion
    def create(self, vals_list):
     res = super().create(vals_list)
        for vals in vals_list:
            if "job_id" in vals:
                res._tag_employees(vals.get("job_id"))
        return res
```

BhaveshHeliconia:
@luisDIXMIT, Please Review.it's done.

luisDIXMIT (approved):
That's correct, LGTM!!

jdoutreloux (approved):
lgtm

rvalyi (approved):

PR #1417, opened 2024-12-13
[18.0][MIG] hr_study: Migration to 18.0

BhaveshHeliconia:


dreispt:
I hope I'm understanding properly this feature, because I believe it is already available in standard Odoo:

image

BhaveshHeliconia:
@dreispt, Yes you are right this feature is available in standard odoo 18 that's why i am closing this

PR #1420, opened 2024-12-17
[FIX] hr_employee_firstname: apply same order (lastname, then firstname)

legalsylvain:
Trivial PR.

rational : 
for the time being, in partner form view and user form view, the fields are displayed with the following order. lastname, then firstname.

However, in the employee form view, the order is reversed.

this PR fixes this inconsistency.

Ref : 

- user view : https://github.com/OCA/partner-contact/blob/16.0/partner_firstname/views/res_user.xml#L13-L20
- partner view : https://github.com/OCA/partner-contact/blob/16.0/partner_firstname/views/res_partner.xml#L18-L29

### DIFF
--- a/hr_employee_firstname/views/hr_view.xml
+++ b/hr_employee_firstname/views/hr_view.xml
@@ -12,12 +12,12 @@
             <xpath expr="//h1//field[@name='name']/.." position="after">
                 <group>
                     <field
-                        name="firstname"
-                        attrs="{'required': [('lastname', '=', False)]}"
-                    />
-                    <field
                         name="lastname"
                         attrs="{'required': [('firstname', '=', False)]}"
+                    />
+                    <field
+                        name="firstname"
+                        attrs="{'required': [('lastname', '=', False)]}"
                     />
                 </group>
             </xpath>


luisg123v (changes requested):
Please fix commit message. It should be a short title, followed by a blank line, followed by a description. It's currently a very long line.

luisg123v:
Shouldn't we reverse the other one instead?

legalsylvain:
> Please fix commit message. It should be a short title, followed by a blank line, followed by a description. It's currently a very long line.

done.

> Shouldn't we reverse the other one instead?

;-)

Well, we have to choose, because it's totally a matter of taste. (in fact, matter of culture. see : https://www.quora.com/What-is-the-correct-order-of-arranging-a-name-last-name-first-or-first-name-first)

My PR is not about to change the order, it's just about to fix an inconsistency. we can not provide a form with field A / field B and another form with field B / field A. It's error prone.

BhaveshHeliconia (approved):
Funtional Review LGTM

legalsylvain:
@luisg123v : could you update your review ? 

thanks !

luisg123v (approved):
LGTM :+1:

legalsylvain:
thanks for your quick review !

PR #1450, opened 2025-03-03
[17.0][MIG] hr_holidays_team_manager: Migration to 17.0

BhaveshHeliconia:


hr_holidays_team_manager/data/hr_holidays_security.xml
```diff
@@ -0,0 +1,66 @@
...
+        <field name="name">Time off/Team manager</field>
+        <field
+            name="rule_groups"
+            eval="[Command.set([ref('hr_holidays.resource_leaves_holidays_user'), ref('hr_holidays.hr_leave_report_rule_group_holiday_user')])]"
+        />
+    </record>
+
+    <record id="group_hr_holidays_officer" model="res.groups">
+        <field name="name">Time off/Holiday Officer</field>
+        <field
```
alexey-pelykh, line 13: In 16.0 this was `(4, ref('hr_holidays.group_hr_holidays_user'))` which uses link/add semantics. `Command.set()` replaces all implied groups, which could remove implied groups added by other installed modules. Should this be `Command.link(ref('hr_holidays.group_hr_holidays_user'))` instead to preserve the original behavior?

alexey-pelykh (commented):
Thanks for the migration. The code and tests look good overall -- `search()` signature updated correctly, Command API applied, and `action_validate` properly adapted for 17.0. One thing to double-check: in `hr_holidays_security.xml`, the `implied_ids` for `group_hr_holidays_officer` was changed from `(4, ref(...))` (link) to `Command.set(...)` (replace). The 16.0 version used link semantics to add the user group without disturbing other implied groups that might be set by third-party modules. Consider using `Command.link()` here instead to preserve the same behavior.

hr_holidays_team_manager/data/hr_holidays_security.xml
BhaveshHeliconia: Yes, you're right — it should use `Command.link()` to preserve the original link semantics.
I've replaced `Command.set(ref('hr_holidays.group_hr_holidays_user'))` with `Command.link(ref('hr_holidays.group_hr_holidays_user'))`.
Thanks for pointing this out and for the review.

alexey-pelykh (approved):
Thanks for the update, @BhaveshHeliconia. Re-reviewed the current diff: `implied_ids` on `group_hr_holidays_officer` in `data/hr_holidays_security.xml` now reads `Command.link(ref('hr_holidays.group_hr_holidays_user'))`, which restores the additive link semantics from 16.0 (`(4, ref(...))`) instead of replacing the group's implied groups. That was my only concern, and it's resolved.

The rest of the 17.0 step holds up — the `hr.employee.base.search()` override uses the 17.0 `(domain, offset, limit, order)` signature, the `action_validate` guard on `hr.leave.allocation` is covered by `test_hr_leave_allocation`, and CI is green. LGTM.

PR #1455, opened 2025-03-26
[17.0][ADD] add hr_governance module

Luca-Policastro:


hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+        required=True,
+        readonly=False,
+        compute="_compute_fields_from_type",
+        store=True,
+    )
+    parent_id = fields.Many2one(
+        "governance.circle",
+        ondelete="restrict",
+        index=True,
+        domain="[('is_circle', '=', True)]",
```
ivantodorovich:
```suggestion
        domain=[('is_circle', '=', True)],
```

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+        is_greyscale_mode_on = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("hr_governance.governance_check_grayscale")
+        )
+        return is_greyscale_mode_on
+
+    @api.model
+    def get_stripe_param(self):
+        is_stripe_all_roles = (
```
ivantodorovich:
```suggestion
        is_stripe_all_roles = str2bool(
```

import from `odoo.tools.misc`

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+            # include member of steering roles of its subcircles
+            # only the next level, not the whole hiearchy
+            steering_roles = rec.child_ids.child_ids.filtered(
+                lambda x: not x.is_circle and x.is_steering_role
+            )
+            rec.member_ids |= steering_roles.member_rel_ids.mapped("member_id")
+
+    @api.model
+    def get_greyscale_mode_param(self):
+        is_greyscale_mode_on = (
```
ivantodorovich:
```suggestion
        is_greyscale_mode_on = str2bool(
```

import from `odoo.tools.misc`

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+                if not rec.user_input_expectation:
+                    rec.expectation = rec.type_id.expectation
+
+                if not rec.user_input_authority:
+                    rec.authority = rec.type_id.authority
+
+    @api.depends("type_id.type")
+    def _compute_is_editable(self):
+        for rec in self:
+            rec.is_editable = bool(not rec.type_id or rec.type_id.type == "template")
```
ivantodorovich:
```suggestion
            rec.is_editable = not rec.type_id or rec.type_id.type == "template"
```
redundant

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+    @api.depends("is_circle")
+    def _compute_is_addable(self):
+        for rec in self:
+            is_single_mode_on = (
+                self.env["ir.config_parameter"]
+                .sudo()
+                .get_param("hr_governance.governance_single_assignee_mode")
+            )
+            if is_single_mode_on:
+                rec.is_addable = not rec.is_circle and not len(rec.member_ids)
```
ivantodorovich:
```suggestion
                rec.is_addable = not rec.is_circle and not rec.member_ids
```
redundant

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+
+    @api.depends("type_id.type")
+    def _compute_is_editable(self):
+        for rec in self:
+            rec.is_editable = bool(not rec.type_id or rec.type_id.type == "template")
+
+    @api.depends("is_circle")
+    def _compute_is_addable(self):
+        for rec in self:
+            is_single_mode_on = (
```
ivantodorovich:
```suggestion
            is_single_mode_on = str2bool(
```

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+        create_table=False,
+        string="Members",
+        compute="_compute_member_ids",
+    )
+    # assigned members for role
+    member_rel_ids = fields.One2many("governance.circle.member.rel", "circle_id")
+    assigned_user_ids = fields.Many2many(
+        "res.users", compute="_compute_assigned_user_ids", string="Assigned Users"
+    )
+    member_cnt = fields.Integer(compute="_compute_member_cnt")
```
ivantodorovich:
```suggestion
    member_count = fields.Integer(compute="_compute_member_cnt")
```

there's point in skipping these 2 characters here. It only hurts readability and it's against odoo codebase conventions

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+            {"type_id": temp.id, **temp._required_role_fields()} for temp in templates
+        ]
+
+    @api.model
+    @ormcache()
+    def _get_enable_edit_circle_role(self):
+        roles = self.env["governance.role.type"].search(
+            [("enable_edit_circle", "=", True)]
+        )
+        return [{"id": role.id, "name": role.name} for role in roles]
```
ivantodorovich: I'd move these 2 methods to the `governance.role.type` model

Reason:
- They do actually just compile a list of `governance.role.type` records data
- The ormcache decorator would be closer to the cache invalidation methods, making it easier to maintain

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+
+    def write(self, vals):
+        context = self.env.context
+        forbidden_fields = self._get_forbidden_change_fields()
+        if forbidden_fields and not context.get("skip_update_user_input"):
+            self._sanity_check(forbidden_fields, vals)
+
+        # Permission constraint
+        allowed_to_edit = (
+            self.id in self.env.user.allowed_edit_governance_ids.ids
```
ivantodorovich:
```suggestion
            self.env.su
            or self.id in self.env.user.allowed_edit_governance_ids.ids
```

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+                    ("parent_id", "=", self.id),
+                    ("type_id", "=", steering_role.id),
+                    ("member_rel_ids", "!=", False),
+                ],
+            )
+            > 0
+        )
+
+    @api.model
+    def _get_steering_role_user_ids(self, circle_id):
```
ivantodorovich: I don't think it makes sense to have this method as `@api.model`.

```suggestion
    def _get_steering_role_user_ids(self):
```

And, instead of calling the method with `model._get_steering_role_user_ids(circle)`, you'd call with `circle._get_steering_role_user_ids()`

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,638 @@
...
+                    ("member_rel_ids", "!=", False),
+                ],
+            )
+            > 0
+        )
+
+    def _is_assigned_w_steering(self):
+        self.ensure_one()
+        steering_role = self.env.ref(
+            "hr_governance.steering_gct", raise_if_not_found=False
```
ivantodorovich, line 502: IMO we shouldn't have business code depend on configurable records like the role types.

Instead, we should perhaps have a "type"-like selection field on the role types, with predefined values like "steering", and use it instead.

That way the code wouldn't rely on given records being created (through xmlid), an rely instead on record configurations

not blocking, though. can be a future improvement

hr_governance/static/src/components/circlepack_colorlist.esm.js
```diff
@@ -0,0 +1,21 @@
...
+    _t("Light Gray"),
+    _t("Aero Blue"),
+    _t("French Pass"),
+    _t("Tasman"),
+    _t("Oyster Bay"),
+    _t("Fog"),
+    _t("Fall Green"),
+    _t("Vanilla Ice"),
+    _t("Shalimar"),
+    _t("Cream Brulee"),
```
ivantodorovich: Do these translations actually work?
I would've assumed we need to use `_lt` here instead, since we can't guarantee translations are loaded at this stage

ivantodorovich (commented):
partial review

hr_governance/models/governance_circle.py
Luca-Policastro: I'm not sure to understand why this is necessary. Does a Fields.Boolean return a string when we use `.get_param` ?

ivantodorovich: There's no `fields.Boolean` at play here, just `ir.config_parameter`'s `get_param`, which always returns a string. Parameters have no type

hr_governance/__manifest__.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "HR Governance",
+    "version": "17.0.1.4.0",
```
SilvioC2C:
```suggestion
    "version": "17.0.1.0.0",
```

hr_governance/migrations/17.0.1.1.0/post-migrate.py
SilvioC2C, line 1: Remove the `migrations` folder: a new module doesn't need any migration

hr_governance/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+from . import res_config_settings
+from . import ir_ui_view
+from . import governance_authority
+from . import governance_expectation
+from . import governance_role_type
+from . import governance_circle
+from . import res_users
+from . import ir_http
```
SilvioC2C: Non blocking: sort them alphabetically to ease readability (unless there's a technical reason to sort them this way)

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+
+    @api.depends("member_ids", "member_rel_ids")
+    def _compute_assigned_user_ids(self):
+        for rec in self:
+            if rec.member_ids:
+                rec.assigned_user_ids = [
+                    Command.link(mem.user_id.id)
+                    for mem in rec.member_ids
+                    if mem.user_id
+                ]
```
SilvioC2C:
```suggestion
                rec.assigned_user_ids = rec.member_ids.user_id
```

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+            elif rec.member_rel_ids:
+                rec.assigned_user_ids = [
+                    Command.link(mem.member_id.mapped("user_id").id)
+                    for mem in rec.member_rel_ids
+                    if mem.member_id.mapped("user_id")
+                ]
```
SilvioC2C:
```suggestion
                rec.assigned_user_ids = rec.member_rel_ids.member_id.user_id
```

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+                rec.suitable_type_ids = self.env["governance.role.type"].search(
+                    [
+                        ("source_circle_id", "=", rec.parent_id.id),
+                    ]
+                )
+                rec.suitable_type_ids |= rec.parent_id.suitable_type_ids
+            else:
+                rec.suitable_type_ids = [Command.clear()]
+
+    def _compute_member_ids(self):
```
SilvioC2C, line 205: No `@api.depends`?

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+    @api.model
+    def get_stripe_param(self):
+        is_stripe_all_roles = str2bool(
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("hr_governance.stripe_all_unassigned_roles")
+        )
+        return is_stripe_all_roles
+
+    def _compute_is_color_field_invisible(self):
```
SilvioC2C, line 235: No `@api.depends`?

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+    @api.model
+    def _get_default_color(self):
+        """List of colors can be found in static/src/components/circlepack_colorlist.js
+        By default,
+            - a circle gets a random color (except white)
+            - a role gets White unless its type is assigned with color
+        """
+        return random.randint(2, 11)
+
+    @api.depends("child_ids")
```
SilvioC2C: I would add `is_root` to the dependencies

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+                rec.is_addable = not rec.is_circle and not rec.member_ids
+            else:
+                rec.is_addable = True
+
+    @api.depends("member_ids")
+    def _compute_member_count(self):
+        for rec in self:
+            rec.member_count = len(rec.member_ids)
+
+    @api.constrains("parent_id")
```
SilvioC2C: Please add `is_root`

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+                new_child_ids = existing_child_ids + [
+                    (0, 0, val) for val in new_roles_vals
+                ]
+                val["child_ids"] = new_child_ids
+
+        records = super().create(vals)
+
+        return records
+
+    def _get_fields(self):
```
SilvioC2C: I would rename this method to make it easier to understand what it does. For example: what are these fields used for?

hr_governance/models/governance_circle.py
```diff
@@ -0,0 +1,620 @@
...
+    )
+    percentage = fields.Float(string="Time dedicated (%)")
+    focus = fields.Text()
+    _sql_constraints = [
+        (
+            "unique_member_per_role",
+            "unique(member_id, circle_id)",
+            "This member is already assigned to the role/circle",
+        ),
+    ]
```
SilvioC2C: I would move these 2 models in 2 different files

hr_governance/models/governance_role_type.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "purpose" in vals:
+            self._update_governance_circle("purpose", vals.get("purpose", ""))
+        if "expectation" in vals:
+            self._update_governance_circle("expectation", vals.get("expectation", ""))
+        if "authority" in vals:
+            self._update_governance_circle("authority", vals.get("authority", ""))
+        self.env.registry.clear_cache()
```
SilvioC2C: Clearing the cache every time a `write()` is called might be not optimal perf-wise. I would add an `if` block and check whether we actually need to do it (for example, cache is cleared only if we change a field used by an `@ormcache`-decorated method).

hr_governance/models/governance_circle.py
Luca-Policastro: I'm not sure to understand how this  2 codes do the same thing ?

SilvioC2C: The old code creates a list of `x2many` fields' commands. Each command adds a `res.users` ID; the users are retrieved by cycling through the `rec.member_ids` recordset and checking whether the member record is linked to a user.
The new code directly retrieves all the users related to the members, and sets them to the field. It is equivalent as doing this:
```
users = self.env["res.users"]
for member in rec.member_ids:
    if member.user_id:
        users += member.user_id
rec.assigned_user_ids = users
```

vvrossem:
superseded by https://github.com/OCA/hr/pull/1521

PR #1456, opened 2025-04-03
[16.0][IMP] hr_course: Add chatter to courses and schedules

etobella:
Add chatter and activities on courses and activities on shcedules

PR #1457, opened 2025-04-03
[16.0][IMP] hr_course: inherit activity mixing model

kluna1998:
Filtering or searching by activities (activity_ids) on the model hr.course.schedule raises an exception because the model does not inherit from mail.activity.mixin.

@etobella


etobella:
Check #1456 It fixes the same

PR #1458, opened 2025-04-10
[18.0][FIX] hr_employee_relative: add compute value for fields impacted by the compute method

jeromeguerriat-msf:
the method "_compute_age" assign values to the fields "age_month" and "age_day", but those fields are not marked as computed.
Therefore, when you have no write rights on the record, displaying "age_year" will end up in an access error


AlvaroRM11 (approved):
LGTM

dreispt (approved):

rvalyi (approved):

PR #1416, opened 2024-12-13
[18.0][MIG] hr_personal_equipment_request: Migration to 18.0

BhaveshHeliconia:

PR #1462, opened 2025-04-22
[MIG] hr_personal_equipment_stock: Migration to 18.0

Alexgars73:
Depends #1416


hr_personal_equipment_stock/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "depends": ["hr_personal_equipment_request", "stock"],
+    "data": [
+        "views/stock_location.xml",
+        "views/stock_move.xml",
+        "views/procurement_group.xml",
+        "views/hr_personal_equipment.xml",
+        "views/hr_personal_equipment_request.xml",
+        "views/stock_picking.xml",
+    ],
+    "auto_install": True,
```
Reyes4711-S73, line 21:
```suggestion
    "auto_install": True,
    "installable": True,
    "application": False
```

Reyes4711-S73 (approved):
LGTM

eugenios73 (approved):
LGFM

rvalyi (approved):

quirino95:
Hi @rvalyi, please could you retry merge? Thank you!

HekkiMelody:
@rvalyi @OCA/human-resources-maintainers could you please retry the merge? Thanks!

Reyes4711-S73:
@OCA/human-resources-maintainers Please, can you review/merge this PR?

hr_personal_equipment_stock/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
...
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    equipment_request_id = fields.Many2one(
+        related="group_id.equipment_request_id", store=True
+    )
+
+    def _action_done(self):
+        res = super()._action_done()
+        if self.equipment_request_id:
```
alexey-pelykh: `_action_done` can be called on a multi-record recordset. Accessing `self.equipment_request_id` directly without iterating will raise if `len(self) > 1`, or silently process only the first record. Wrap the body in `for picking in self:` and replace `self` references with `picking`.

hr_personal_equipment_stock/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
...
+            for move in self.move_ids_without_package:
+                if move.state == "done":
+                    request_lines = self.equipment_request_id.sudo().line_ids.filtered(
+                        lambda x, move=move: x.product_id == move.product_id
+                    )
+                    for line in request_lines:
+                        qty_initial = line.product_uom_id._compute_quantity(
```
alexey-pelykh: Both `qty_initial` and `qty_done` convert from `line.product_uom_id` to `line.product_uom_id` -- this is converting a UoM to itself, so it's always a no-op. Was this meant to convert from the move's UoM?

hr_personal_equipment_stock/models/hr_personal_equipment.py
```diff
@@ -0,0 +1,121 @@
...
+                lambda r, line=line: r.state == "done"
+                and line.product_id == r.product_id
+            ):
+                qty += move.product_uom._compute_quantity(
+                    move.product_uom_qty, line.product_uom_id
+                )
+            line.qty_delivered = qty
+
+    def _skip_procurement(self):
+        return self.product_id.type not in ("consu", "product")
```
alexey-pelykh: Since Odoo 17+, the `"product"` type no longer exists (storable products are `"consu"` with `is_storable=True`). This check still works but the `"product"` branch is dead code. For an 18.0 module, consider simplifying to `self.product_id.type == "service"`.

alexey-pelykh (commented):
Thanks for the migration, overall a solid port. A few things I noticed:

**`stock_picking.py` -- `_action_done` treats `self` as singleton**

`_action_done` is called on a recordset but the code accesses `self.equipment_request_id` directly without iterating. This will break (or silently misbehave) when multiple pickings are validated together. Should be wrapped in `for picking in self:`.

**`stock_picking.py` -- UoM conversion is a no-op**

Lines computing `qty_initial` and `qty_done` both convert from `line.product_uom_id` to `line.product_uom_id` (same UoM). This was probably intended to normalize from the move's UoM to the line's UoM, but as written it does nothing. Consider whether this conversion is actually needed or if a direct comparison of `line.quantity <= line.qty_delivered` would suffice.

**`hr_personal_equipment.py` -- dead product type check**

In `_skip_procurement`, the check for `"product"` in product types is a leftover from pre-17.0. Since Odoo 17+ the storable type was merged into `"consu"` (with `is_storable` flag). Not a bug, but cleaning it up would be consistent with an 18.0 migration. Could be simplified to `self.product_id.type != "consu"` or, for 18.0 correctness, check `self.product_id.type == "service"`.

Alexgars73:
I've tested the entire module but I can't achieve project coverage.

alexey-pelykh (approved):
Revisiting this migration — apologies @Alexgars73, my earlier inline notes were mistaken: I flagged them against an older version of the source rather than this diff, and I've removed them. Re-reading the current code:

- `_action_done` correctly iterates `for picking in self:` — no singleton issue.
- The quantity computation does a real `move.product_uom._compute_quantity(move.product_uom_qty, line.product_uom_id)` conversion — not a no-op.
- `_skip_procurement` already uses `type == "service"`.

No concerns from me — clean port. Approving. Thanks @Reyes4711-S73, @eugenios73 and @rvalyi for the reviews.

BhaveshHeliconia (approved):
LGTM!

PR #1466, opened 2025-05-02
[18.0] [MIG] hr_employee_id: Migration to 18.0

luisDIXMIT:
Standard migration


luisDIXMIT:
Tests are failing due to a warning log, but that is the expected behavior, so it shouldn't be considered an error.

hr_employee_id/readme/CONFIGURE.md
```diff
@@ -0,0 +1,8 @@
+If you want to modify the format of the sequence, go to Settings -\>
+Technical -\> Sequences & Identifiers -\> Sequences and search for the
+"Employee ID" sequence, where you modify its prefix and numbering
+formats.
+
+To configure the 'ID Generation Method', the '# of Digits' and the
+'Sequence', activate the developer mode and go to Employees -\>
+Configuration -\> Employee ID.
```
hailangvn: I could not locate `Employee ID` in `Configuration` menu. It has been removed by the commit [hr_employee_id: Migration to 12.0](https://github.com/OCA/hr/pull/1466/commits/ef544674cbdc27d62e9e5d84ca4049a25211b7c5#diff-493586f2dd74806ceeca74782430dc18e02eb7e3504ba6578868e0582427b22cL48-L49).

hailangvn (approved):
Functional test is ok. Code LGTM. Thanks for the work.

ivantodorovich:
> Tests are failing due to a warning log, but that is the expected behavior, so it shouldn't be considered an error.

We still need to get a green build from the CI, thuogh 😓 

There are two options, AFAICS:

1: Preferred) In the test `test_no_sequences_id_generation `, capture the logs with `self.assertLogs`. This way we test the log is outputted, and we also silence it at the same time

2) IMO not ideal here, but sharing just in case it's useful in the future: it's possible to add a `checklog-odoo.cfg` file in the root of the repository, to configure https://github.com/acsone/checklog-odoo to ignore some log lines

ivantodorovich (approved):
Pre-approving: test need fixing according to previous comment

etobella (approved):
LGTM

dreispt (approved):

dreispt:
Something is missing in the testes:
`WARNING odoo odoo.addons.hr_employee_id.models.hr_employee: No sequence configured for employee ID generation`

etobella:
@luisDIXMIT can you check?

luisDIXMIT:
> > Tests are failing due to a warning log, but that is the expected behavior, so it shouldn't be considered an error.
> 
> We still need to get a green build from the CI, thuogh 😓
> 
> There are two options, AFAICS:
> 
> 1: Preferred) In the test `test_no_sequences_id_generation `, capture the logs with `self.assertLogs`. This way we test the log is outputted, and we also silence it at the same time
> 
> 2. IMO not ideal here, but sharing just in case it's useful in the future: it's possible to add a `checklog-odoo.cfg` file in the root of the repository, to configure https://github.com/acsone/checklog-odoo to ignore some log lines

Hi @ivantodorovich, I used the first one. Thanks for the suggestion, your revision, and your approval.

hr_employee_id/readme/CONFIGURE.md
luisDIXMIT: I updated it. Thanks for your review.

luisDIXMIT:
> Something is missing in the testes: `WARNING odoo odoo.addons.hr_employee_id.models.hr_employee: No sequence configured for employee ID generation`

Hi @dreispt, I captured it and checked if it's the correct warning that should be generated. I think everything is fine. This way, it shouldn't now appear in the warnings. Thanks!

PR #1470, opened 2025-05-12
[18.0] [ADD] hr_contract_renew: New module

luisDIXMIT:
This module allows you to renew a contract by reusing the original contract's data, while choosing the start date for the new contract and marking automatically the original contract as expired, setting its end date to the day before the new contract's start date.

@etobella


hr_contract_renew/models/hr_contract.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2025 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class HrContract(models.Model):
+    _inherit = "hr.contract"
```
etobella: If you don't need, it, do not declare it :wink:

hr_contract_renew/views/hr_contract_view.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="inherit_id" ref="hr_contract.hr_contract_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//header">
+                <button
+                    name="%(hr_contract_renew.action_open_renew_contract_wizard)d"
+                    type="action"
+                    string="Renew Contract"
+                    class="oe_highlight"
+                    context="{'default_contract_id': id}"
+                    invisible="contracts_count == 0 or date_end"
```
etobella: Why contracts_count == 0?

etobella:
The compute of employee is not done properly. To test it, make a new employee, set a different employee calendar on the contract and the renewed contract will not be fine. Can you check it?

hr_contract_renew/views/hr_contract_view.xml
luisDIXMIT: Because I think if we are creating the first contract has no sense show the renew button. Is not correct?

etobella: Actually, it has no sense if id is not set, not if contracts_count is 0 :smile:

luisDIXMIT: I changed it to "not id or date_end"

hr_contract_renew/models/hr_contract.py
luisDIXMIT: Deleted. Thanks!!

luisDIXMIT:
> The compute of employee is not done properly. To test it, make a new employee, set a different employee calendar on the contract and the renewed contract will not be fine. Can you check it?

resource_calendar_id has copy=false and set company calendar as default. So, i passed it in the copy dict. I think is already fine.

etobella (approved):

hr_contract_renew/views/hr_contract_view.xml
```diff
@@ -0,0 +1,20 @@
...
+        <field name="inherit_id" ref="hr_contract.hr_contract_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//header">
+                <button
+                    name="%(hr_contract_renew.action_open_renew_contract_wizard)d"
+                    type="action"
+                    string="Renew Contract"
+                    class="oe_highlight"
+                    context="{'default_contract_id': id}"
+                    invisible="not id or date_end"
```
kluna1998, line 15:
```suggestion
                    states="draft,open"
                    invisible="not id"
```
Even if date_end is set (as in temporary contracts), renewal might still be needed.
Would it make sense to relax that condition? What do you think, @luisDIXMIT ?

luisDIXMIT, line 15: states doesn't work in 18

etobella:
@kluna1998 Can you review it again?

kluna1998 (approved):

PR #1473, opened 2025-06-02
[18.0][ADD] hr_shift_planning

dreispt:


dreispt:
I might consider moving this to https://github.com/OCA/resource, as it is not really tied to HR and Employees.

pedrobaeza:
@dreispt we are just presenting in these days our shift planning module: https://github.com/OCA/shift-planning/pull/11

https://www.aeodoo.org/event/xviii-jornadas-odoo-2025-130/track/gestion-de-turnos-de-personal-con-odoo-885

victoralmau:
Hello,

Reviewing the PR, I see there's no description of how it works or how to use it, no views to test it functionally, and no tests. I think the approach is "similar" to what already exists in `hr_shift` https://github.com/OCA/shift-planning/pull/11. What do you think?

dreispt:
@victoralmau Indeed, it is in an early stage of development, and it is missing all of those.

I will do a have a closer look at the shift planning repo.
I may have missed it, but from an early overview I couldn't find support for rotating shift there.

Here is an example:

```
Basic use case:
- A position must be filled by a person 24/7.
- It's divided in shifts, for example two 12h shifts: Day / Night.  
- This requires 4 workers. Each worker has schedule with each days's shift.

For example, the DuPont shift schedule, used in the United States in several
manufacturing industries and police departments, uses 12-hour rotating shifts:
    
	Week  Mon 	Tue 	Wed 	Thu 	Fri 	Sat 	Sun
	1 	Night Night Night Night Off 	Off 	Off
	2 	Day 	Day 	Day 	Off 	Night Night Night
	3 	Off 	Off 	Off 	Day 	Day 	Day 	Day
	4 	Off 	Off 	Off 	Off 	Off 	Off 	Off

    
Schedule template, with 2 shifts [D,N] and 4 roster positions [A,B,C,D]:

     Pos  Week1	Week2	    Week3	Week4  
	A   NNNN---  DDD-NNN  ---DDDD  -------
	B   -------  NNNN---  DDD-NNN  ---DDDD
	C   ---DDDD  -------  NNNN---  DDD-NNN
	D   DDD-NNN  ---DDDD  -------  NNNN---

Schedules can also be mapped by shift instead of roster position:

    Shift Week1	 Week2    Week3	 Week4  
	D   DDDCCCC  AAADDDD  BBBAAAA  CCCBBBB
	N   AAAADDD  BBBBAAA  CCCCBBB  DDDDCCC   
```

victoralmau:
Okay, I think I understand the use case, and it's not currently covered, but I think it could be covered with "not too much effort," and it would be interesting to have that functionality. 

I'll talk about it in person tomorrow, and I'll tell you my approach.

victoralmau:
I'm writing down my ideas (to address the needs you mentioned by improving `hr_shift`):

- Ability to define the days of the week to which shifts apply (from Monday to Sunday, a Boolean field for each day of the week to be able to decide exactly which days you want for that shift).
- Add a sequence field to shifts to sort them.
- Add a field for employees to indicate the rotating shifts they will use.

With these changes, the idea is that the first schedule must be done "manually," thus indicating the corresponding shift for each employee. However, in subsequent schedules, the next shift for each employee would be automatically defined for each assignment (until starting over).

What do you think of this approach?

PR #1479, opened 2025-06-19
[17.0][IMP] hr_employee_medical_examination: changes & imrpovements

Abranes:
Hello,

We have made some changes to the `hr_employee_medical_examination` module. The updates/improvements are as follows:

- We have changed the examination date field from `date` to `datetime`. This allows specifying an exact time for the medical examination.
- We have added the ability to set an examination location (`res.partner`) to indicate where the examination will take place. By default, this field will be populated with the partner associated with the company.

We are also working on an improvement that will allow us to send a reminder to the employee, indicating when and where their medical examination will take place.

Any suggestions are welcome.

@Loregs2 @jaenbox

### DIFF
--- a/hr_employee_medical_examination/models/hr_employee_medical_examination.py
+++ b/hr_employee_medical_examination/models/hr_employee_medical_examination.py
@@ -46,6 +46,11 @@ class HrEmployeeMedicalExamination(models.Model):
     year = fields.Char(default=lambda r: str(datetime.date.today().year))
 
     note = fields.Text(tracking=True)
+    address_id = fields.Many2one(
+        "res.partner",
+        tracking=True,
+        default=lambda self: self.env.company.partner_id.id,
+    )
 
     @api.onchange("date")
     def _onchange_date(self):
--- a/hr_employee_medical_examination/models/hr_employee_medical_examination.py
+++ b/hr_employee_medical_examination/models/hr_employee_medical_examination.py
@@ -27,7 +27,7 @@ class HrEmployeeMedicalExamination(models.Model):
         tracking=True,
     )
 
-    date = fields.Date(
+    date = fields.Datetime(
         string="Examination Date",
         tracking=True,
     )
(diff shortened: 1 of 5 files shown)


mymage (commented):
Changing to datetime is good for me.
On location I would ad a second field, string not coded, to set where in the first location: ie. Infirmary

mymage:
@Abranes give a push to this PR (mybe with my suggestion ;-)) to keep it alive

jaenbox:
LGTM

jaenbox (approved):

alexey-pelykh (changes requested):
Thanks for the migration! A few things to address:

- The version is set to `17.0.1.0.0` but the 16.0 branch already had `16.0.1.1.0`, so this should be `17.0.1.1.0` to avoid a downgrade.

- The `expiry_date` field changed from `Date` to `Datetime` -- that needs a pre-migrate or post-migrate script to convert existing data, otherwise the column type mismatch will cause issues on upgrade.

- In the test, `medical_examination_date` is set as a `Datetime` string but the model field is `fields.Date`, so the assertion works by accident. Worth aligning the test data types.

Happy to re-review once updated.

hildickethan (approved):

alexey-pelykh (approved):
Re-reviewed, and I owe you a correction here, @Abranes. My earlier CHANGES_REQUESTED referenced an `expiry_date` field, a `medical_examination_date` test value, and a `17.0.1.0.0` version — none of which are in this PR. Apologies for the noise; those points don't apply.

What the PR actually does, and it looks good:

- `models/hr_employee_medical_examination.py`: the `date` (Examination Date) field changes `Date` → `Datetime`; `_onchange_date` still derives `year` correctly from `record.date.year`. date→timestamp is a compatible widening, so no data migration is needed.
- Adds `address_id` (res.partner) on both the examination and the generate wizard, defaulting to `env.company.partner_id`, and propagates it through `_create_examination_vals`.
- Test coverage is solid — it asserts the wizard's default `address_id`, then exercises the `address_id=False` path through to the created examination.

`__manifest__.py` isn't touched (already `17.0.1.1.0` on 17.0), which is fine — the bump is ocabot's job on merge. Clearing my earlier request. LGTM.

PR #1480, opened 2025-06-19
[16.0][FIX] hr_employee_medical_examination: Use _for_xml_id() to avoid access errors

victoralmau:
Changes done:
- [x] Use `_for_xml_id()` to avoid access errors.
- [x] Change `TransactionCase` to `BaseCommon`

@Tecnativa TT56829


pedrobaeza (approved):
Clear patch, so fast-tracking it:

/ocabot merge patch

PR #1483, opened 2025-06-25
[18.0][FIX] hr_employee_firstname: Use work_contact_id instead of address_home_id

ByteMeAsap:
Changes to adapt the module as per changes in base Odoo module in https://github.com/odoo/odoo/commit/e8c48f824e078d643a3385fb910707d5525e927d


jdoutreloux (approved):

micheledic (approved):

micheledic:
@pedrobaeza on every instance where this module is installed appear an error when you update the name and last name of employee . Imho this PR should be approved asap

PR #1485, opened 2025-06-25
[18.0][MIG] hr_ personal_equipment_request_tier_validation

BhaveshHeliconia:
Depends on

-  https://github.com/OCA/hr/pull/1416


hr_personal_equipment_request_tier_validation/data/tier_definition.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+    Copyright 2021 Creu Blanca
+    License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
```
alexey-pelykh: Nit: this header says `LGPL-3.0` but the manifest and all `.py` files declare `AGPL-3.0`. Should be `AGPL-3.0` for consistency.

alexey-pelykh (approved):
Thanks for the migration. The module follows the standard tier validation bridge pattern and CI is all green. One minor note: the license header in `data/tier_definition.xml` says LGPL-3 while the module is AGPL-3 -- not blocking but worth aligning if you get a chance.

hr_personal_equipment_request_tier_validation/data/tier_definition.xml
BhaveshHeliconia: Thanks for pointing it out. I've updated the license header to align with the module license.

PR #1488, opened 2025-07-23
[FIX] hr_job kanban view to show to_recruit instead of no_of_recruitment

dcamacho20:


luisDIXMIT (commented):
You are making the job editable for interviewers, but it shouldn't be. You are also removing the ability for an interviewer to see all recruitments for a job — they can now only see the ones that are still pending.

PR #1489, opened 2025-07-24
[FIX] Permissions for time-off

dcamacho20:


luisDIXMIT (commented):
I'm not sure if time off team manager should have same permissions as time off administrator. Anyway the code works fine. Also two test are failing.

PR #1490, opened 2025-07-29
[16.0][IMP] hr_employee_medical_examination: Add Activities filter + kanban view

victoralmau:
Add Activities filter + kanban view

**Before**
antes

**After**
despues

@Tecnativa TT57402


hr_employee_medical_examination/views/hr_employee_medical_examination_views.xml
```diff
@@ -113,13 +132,58 @@
...
+                                        </strong>
+                                        <span class="o_kanban_record_subtitle"><field
+                                                name="employee_id"
+                                            /></span>
+                                        <span class="o_kanban_record_subtitle"><field
+                                                name="date"
+                                            /></span>
+                                        <span class="o_kanban_record_subtitle"><field
+                                                name="state"
+                                            /></span>
```
carlos-lopez-tecnativa: Could you use `widget="label_selection"` and add some classes to differentiate the states based on colors?
image

Like as 
image

carlos-lopez-tecnativa (changes requested):
Could you improve the model description to remove “HR” and keep only “Employee Medical Examination”? I understand that activities read the [model description](https://github.com/odoo/odoo/blob/1adee2b7452c072bcce51cd75641a9e4c2b7f219/addons/mail/models/res_users.py#L213-L220).

https://github.com/Tecnativa/hr/blob/768c50a810ebbcfc3a66743d696cbdd565531473/hr_employee_medical_examination/models/hr_employee_medical_examination.py#L12
image

hr_employee_medical_examination/views/hr_employee_medical_examination_views.xml
victoralmau: Thanks.

ejemplo

carlos-lopez-tecnativa (approved):
LGTM

hr_employee_medical_examination/views/hr_employee_medical_examination_views.xml
```diff
@@ -113,13 +132,62 @@
...
+        </field>
+    </record>
     <record
         model="ir.actions.act_window"
         id="hr_employee_medical_examination_act_window"
     >
         <field name="name">Employee Medical Examinations</field>
         <field name="res_model">hr.employee.medical.examination</field>
-        <field name="view_mode">tree,form</field>
+        <field name="view_mode">kanban,tree,form</field>
```
pedrobaeza: Why `kanban` is before `tree` one?

victoralmau: Initially, I thought that the kanban view was more useful, which is why I defined it first, but your question made me think that it might cause more confusion than improvement, so I changed it so that it is not defined first.

pedrobaeza (approved):
/ocabot merge minor

PR #1491, opened 2025-07-30
[16.0][FIX]hr_employee_calendar_planning: Fix incorrect behaviour _check_company_id

ChristophAbenthungCibex:
The check does not consider the calendar itself. It only checks if any calendar has inconsistency with companies.

Steps to reproduce:
1. Create second company (If only one exists)
2. Change the company of any employee with a Working time (for example Abigail Peterson)
3. Go to working times
4. Create a new working time
5. Click save

The Validation Error of _check_company_id of resource.calendar is shown. The new resource.calendar is currently not used for any employee. The Error should not be thrown.

PR #1492, opened 2025-07-30
[17.0][FIX]hr_employee_calendar_planning: Fix _check_company_id

ChristophAbenthungCibex:
V17 version of #1491 

The check does not consider the calendar itself. It only checks if any calendar has inconsistency with companies.

Steps to reproduce:
1. Create second company (If only one exists)
2. Change the company of any employee with a Working time (for example Abigail Peterson)
3. Go to working times
4. Create a new working time
5. Click save

The Validation Error of _check_company_id of resource.calendar is shown. The new resource.calendar is currently not used for any employee. The Error should not be thrown.


pedrobaeza (approved):
Fast-tracking as the same as 16:

/ocabot merge patch

Only 18 rests.

PR #1500, opened 2025-09-25
[18.0][MIG]hr_personal_equipment_request: migration to 18.0

iammiracle:


hr_personal_equipment_request/models/hr_employee.py
```diff
@@ -0,0 +1,55 @@
...
+        comodel_name="hr.personal.equipment",
+        inverse_name="employee_id",
+        domain=[("state", "not in", ["draft", "cancelled"])],
+    )
+
+    equipment_request_count = fields.Integer(
+        compute="_compute_equipment_request_count",
+    )
+
+    personal_equipment_count = fields.Integer(
```
alexey-pelykh, line 25: This compute doesn't iterate over `self`, so it will fail on multi-record recordsets. Should be:
```python
def _compute_equipment_request_count(self):
    for rec in self:
        rec.equipment_request_count = len(rec.equipment_request_ids)
```
Same issue with `_compute_personal_equipment_count` below.

hr_personal_equipment_request/models/hr_personal_equipment_request.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _accept_request_vals(self):
+        return {"state": "accepted"}
+
+    def cancel_request(self):
+        for rec in self:
+            rec.state = "cancelled"
+            rec.line_ids.update({"state": "cancelled"})
+
+    def _compute_equipment_request_count(self):
```
alexey-pelykh, line 56: This method references `self.equipment_request_count` and `self.equipment_request_ids`, but neither field exists on `hr.personal.equipment.request`. Looks like dead code copied from `hr_employee.py` -- should be removed.

hr_personal_equipment_request/models/hr_personal_equipment.py
```diff
@@ -0,0 +1,86 @@
...
+    equipment_request_id = fields.Many2one(
+        comodel_name="hr.personal.equipment.request", required=True, ondelete="cascade"
+    )
+    quantity = fields.Integer(default=1)
+    product_uom_id = fields.Many2one("uom.uom", "Unit of Measure")
+
+    @api.onchange("product_id")
+    def _onchange_uom_id(self):
+        if self.product_id:
+            self.product_uom_id = self.product_id.uom_id
```
alexey-pelykh, line 46: Returning a domain dict from `onchange` is deprecated in 17.0+. Consider moving the UoM domain to the field definition in the view XML using the `domain` attribute, or defining it on the field directly.

hr_personal_equipment_request/views/hr_personal_equipment_views.xml
```diff
@@ -0,0 +1,167 @@
...
+                        <field
+                            name="equipment_request_id"
+                            readonly="state != 'draft'"
+                        />
+                    </group>
+                    <group name="date">
+                        <field
+                            name="start_date"
+                            readonly="state not in  ['draft','accepted']"
+                        />
```
alexey-pelykh, line 58: Minor: double space in `state not in  ['draft','accepted']`.

hr_personal_equipment_request/views/hr_personal_equipment_request_views.xml
```diff
@@ -0,0 +1,174 @@
...
+                <sheet>
+                    <div name="button_box" class="oe_button_box">
+                        <button
+                            name="action_open_personal_equipment"
+                            type="object"
+                            icon="fa-briefcase"
+                            invisible="allocations_count == 0"
+                        >
+                            <!-- TODO: IN version 14.0 (with fa 5.0) we could put a t-shirt as icon -->
+                            <field
```
alexey-pelykh, line 40: This TODO comment about Font Awesome 5.0 in version 14.0 is outdated -- we're at 18.0 now. Either remove the comment or update the icon.

alexey-pelykh (commented):
Thanks for the migration! The `attrs` to inline attributes and `tree` to `list` conversions look correct. A few things worth addressing: the compute methods in `hr_employee.py` don't iterate over `self` (works only for singletons), there's a dead `_compute_equipment_request_count` method in `hr_personal_equipment_request.py` referencing fields that don't exist on that model, and the `_onchange_uom_id` returning a domain dict is deprecated in 18.0 -- consider using the `domain` attribute on the field in the view instead.

PR #1486, opened 2025-07-05
[MOD][16.0] hr_employee_language: languages selection

mymage:
Changes the source of language list moving it from the res.lang.csv file to res.lang model.
In this way if you add a language at runtime it will be immediately available.

@dreispt In the last weeks I did 3 PRs for porting the module to 16, 17 and 18 that were kindly merged.
Only after I understood that if you add a language at runtime it isn't available to the module.
I propose this solution but maybe there's a better one.
If ok, then I'll do PR for other versions.

I don't know if this can be set as FIX PR

PR #1512, opened 2025-10-12
[MOD][17.0] hr_employee_language: languages selection

mymage:
As in #1486 
Changes the source of language list moving it from the res.lang.csv file to res.lang model.
In this way if you add a language at runtime it will be immediately available.

Added _compute_display_name (ex name_get) to show language name instead language code on selection or m2o widget


mymage:
@dreispt @primes2h @TheMule71 plase review.
I'm quite sure that the _compute_display_name can be better.
I inserted the "self_data = self if self else all_lang" because on tests the for "record in self" doesn't activate.

hr_employee_language/models/hr_employee_language.py
```diff
@@ -1,17 +1,42 @@
...
+        all_lang = self.env["res.lang"].search([("active", "in", (True, False))])
+        self_data = self if self else all_lang
+        for record in self_data:
+            lang = self.env["res.lang"].search(
+                [("code", "=", record.name), ("active", "in", (True, False))], limit=1
+            )
+            language = lang.name if lang else record.name
+            name = "%s" % language
+            result.append((record.id, name))
+        return result
```
alexey-pelykh: This should follow the Odoo 17.0 compute pattern:
```python
def _compute_display_name(self):
    for record in self:
        lang = self.env['res.lang'].search(
            [('code', '=', record.name), ('active', 'in', (True, False))], limit=1
        )
        record.display_name = lang.name if lang else record.name
```
Returning a list of tuples won't work — `_compute_display_name` is a compute method on `display_name` field.

alexey-pelykh (commented):
Thanks for the contribution! The approach of using `res.lang` instead of `tools.scan_languages()` is a better direction.

However, `_compute_display_name` in Odoo 17.0 is a compute method — it shouldn't return a list of tuples. It must iterate over `self` and set `record.display_name` for each record. The current implementation will silently fail. Same issue in the test: `_compute_display_name()` won't return anything in the new API.

Also noticed the badge in the RST was changed from `license` to `licence` — looks like a typo introduced by the readme regeneration.

mymage:
@alexey-pelykh another alligned, thanks

alexey-pelykh:
Thanks @mymage — the model is now correct: `_compute_display_name` iterates `self` and assigns `record.display_name`, exactly right for 17.0.

The remaining red is `codecov/patch`, and it comes from the test not actually exercising the compute:

```python
def test_compute_display_name(self):
    lang_ref = self.env["res.lang"].search([("code", "=", "en_GB"), ...], limit=1)
    self.env["hr.employee.language"]._compute_display_name()   # empty recordset — loop never runs
    langs_name = self.env["res.lang"].search([("code", "=", "en_GB"), ...], limit=1)
    self.assertEqual(lang_ref.name, langs_name.name)           # compares res.lang to itself
```

`_compute_display_name()` is called on an empty recordset, so the loop body never runs (that's why patch coverage fails), and the assertion compares a `res.lang` record's name to itself — it passes without touching `hr.employee.language` at all.

Same fix as the one that landed on #1517: create a record and read `display_name`, which triggers the compute automatically:

```python
record = self.env["hr.employee.language"].create({
    "name": "en_GB",
    "employee_id": self.env["hr.employee"].create({"name": "Test"}).id,
})
self.assertEqual(record.display_name, "English (UK)")
```

mymage:
@alexey-pelykh  @dreispt @primes2h @TheMule71 @sergiocorato thanks

PR #1517, opened 2025-10-16
[MOD][18.0] hr_employee_language: languages selection

mymage:
As in #1486
Changes the source of language list moving it from the res.lang.csv file to res.lang model.
In this way if you add a language at runtime it will be immediately available.

Added _compute_display_name (ex name_get) to show language name instead language code on selection or m2o widget


mymage:
@dreispt @primes2h @TheMule71 plase review.
Little differences from #1512

hr_employee_language/models/hr_employee_language.py
```diff
@@ -1,19 +1,44 @@
...
     can_listen = fields.Boolean(string="Listen", default=True)
+
+    def _compute_display_name(self):
+        """
+        redefine record name to show languages name instead languages code
+        """
+        result = []
+        all_lang = self.env["res.lang"].search([("active", "in", (True, False))])
+        self_data = self if self else all_lang
+        for record in self_data:
```
dreispt: I don't understand the point for this code.
What can't you just loop computed records as usual?
```suggestion
        for record in self:
```

hr_employee_language/models/hr_employee_language.py
```diff
@@ -1,19 +1,44 @@
...
+            lang = self.env["res.lang"].search(
+                [("code", "=", record.name), ("active", "in", (True, False))], limit=1
+            )
```
dreispt, line 35: Search inside a loop is a performance red flag...

hr_employee_language/models/hr_employee_language.py
```diff
@@ -1,19 +1,44 @@
...
+            language = lang.name if lang else record.name
+            name = _("%(lang)s", lang=language)
+            result.append((record.id, name))
```
dreispt: isn't the display_name calculation assignment?
```suggestion
            record.display_name = lang.name if lang else record.name
```

mymage:
@dreispt Thanks for your notes; I'm not so skilled on coding so your suggestion is very appreciated. I'll give a look asap.

hr_employee_language/models/hr_employee_language.py
mymage: I did so because if I use your suggestion, that was my first option, test fails with:
 ```
File "/__w/hr/hr/hr_employee_language/tests/test_add_lang.py", line 33, in test_compute_display_name
    langs_name[0][1],
IndexError: list index out of range
```
Maybe the test is not well done.

File: hr_employee_language/models/hr_employee_language.py (hunk 2 above)
Comment (mymage) [reply hr_employee_language/models/hr_employee_language.py:35]: I changed the code, hope this is better

**Comment by mymage** [comment]
@dreispt Thanks to your comments I refactored the code. The only "strange" thing that I made is this:
`self_data = self.env["hr.employee.language"].search([])
        for record in self_data:`
 to have CodeCov at 100%: this because despite I create an element in hr.employee.language for test, here self is empty.

**Comment by mymage** [comment]
@dreispt @primes2h @TheMule71 ping. Thanks

**Comment by mymage** [comment]
@dreispt @primes2h @TheMule71 ping. Thanks

File: hr_employee_language/models/hr_employee_language.py [hunk 4]
```diff
@@ -1,19 +1,45 @@
...
     can_listen = fields.Boolean(string="Listen", default=True)
+
+    def _compute_display_name(self):
+        """
+        redefine record display_name to show languages name instead languages code
+        """
+        lang_tot_db = self.env["res.lang"].search([("active", "in", (True, False))])
+        dati = {}
+        res = []
+        for lan in lang_tot_db:
```
alexey-pelykh: This searches ALL `hr.employee.language` records instead of iterating over `self`. Should be `for record in self:` and then look up the display name from `dati`. The current code also won't set `display_name` on records not in `self`.

hr_employee_language/models/hr_employee_language.py
```diff
@@ -1,19 +1,45 @@
...
 class HrEmployeeLanguage(models.Model):
     _name = "hr.employee.language"
     _description = "HR Employee Language"
 
-    name = fields.Selection(
-        tools.misc.scan_languages(), string="Language", required=True
-    )
+    def _get_selection(self):
+        """
+        generate language selection from available on db
```
alexey-pelykh, line 13: Selection keys shouldn't be translated — `_("%(code)s", code=lang.code)` will attempt to translate the language code string. Use the raw `lang.code` as the key and `lang.name` as the label directly:
```python
selection.append((lang.code, lang.name))
```

alexey-pelykh (changes requested):
Thanks for improving the language selection to use `res.lang` records. A couple of things to address:

1. `_compute_display_name()` searches ALL `hr.employee.language` records (`self.env["hr.employee.language"].search([])`) instead of iterating over `self`. This is both a performance issue and semantically incorrect — compute methods should only set values on `self` records.

2. The compute method returns a list of tuples, but Odoo compute methods don't use return values — they assign to `self.display_name` (which you do, but the return is dead code).

3. Using `_()` on selection keys (`_("%(code)s", code=lang.code)`) translates the code itself, which shouldn't be translated — codes are stable identifiers.

Happy to re-review once updated.

mymage:
Thanks @alexey-pelykh for your suggestions. I updated the code, I hope in the right way.
If this is ok I will fix the other two.

alexey-pelykh (changes requested):
Thanks for the update — the model code now correctly addresses all three issues from my previous review:

- `_compute_display_name` iterates `self` instead of searching all records
- No more dead return value
- No more `_()` on language codes

However, the test for `_compute_display_name` doesn't actually exercise the compute method:

```python
self.env["hr.employee.language"].create({...})  # result not stored
self.env["hr.employee.language"]._compute_display_name()  # called on empty recordset
```

Since the create result isn't assigned, `_compute_display_name()` runs on an empty recordset — the loop body never executes. This is likely why codecov is failing. The fix:

```python
record = self.env["hr.employee.language"].create({
    "name": "en_GB",
    "employee_id": self.env["hr.employee"].create({"name": "Test"}).id,
})
self.assertEqual(record.display_name, "English (UK)")
```

Accessing `record.display_name` triggers the compute automatically — no need to call it explicitly.

Co-Reviewed-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

mymage:
Ok @alexey-pelykh another thanks for your suggestions. Now I think it is ok but I don't understand the last red codecov

alexey-pelykh (approved):
Thanks for the update @mymage — the test now correctly exercises the compute method. All feedback from both reviews is addressed.

Regarding the `codecov/project` failure: this check measures **overall repository coverage**, not your PR's coverage. Your `codecov/patch` check passes, meaning your new code is adequately covered. The `codecov/project` failure is a pre-existing coverage gap in the repo — not something you need to fix here.

Co-Reviewed-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

mymage:
@dreispt @primes2h @TheMule71 @sergiocorato please review, thanks

PR #1523, opened 2025-11-19
[18.0][ADD] hr_work_entry_timesheet: display corresponding timesheet on work entries

remi-filament:


alexey-pelykh (commented):
Interesting module — bridging work entries and timesheets is useful. I see codecov checks are failing though. Might be worth looking into the coverage on the new code paths, especially around `_get_work_entry()`. The Alpha maturity flag is fair given the scope.

PR #1524, opened 2025-11-19
[18.0][ADD] hr_work_entry_validate: simple validation of work entries by employees

remi-filament:

PR #1525, opened 2025-11-19
[18.0][ADD] hr_work_entry_timesheet_validate

remi-filament:
Glue module between hr_work_entry_validate ( #1524 ) and hr_work_entry_timesheet ( #1523 ) to show timesheet on work entry validation interface for employees.


Hugo-Trentesaux (approved):

alexey-pelykh (commented):
Thanks for this. The glue approach between `hr_work_entry_timesheet` and `hr_work_entry_validate` makes sense.

Tests pass, but the "Detect unreleased dependencies" check is failing — this depends on modules that haven't been merged into the 18.0 branch yet. Once the dependencies land, this should be good to go.

PR #1527, opened 2025-11-24
[18.0][MIG] hr_appraisal_oca: Migration to version 18.0

christian-ramos-tecnativa:
Standard migration

@Tecnativa TT59252
@ebauza @pedrobaeza


pedrobaeza (approved):
@ebauza please check

hr_appraisal_oca/i18n/hr_appraisal_oca.pot
```diff
@@ -0,0 +1,1086 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* hr_appraisal_oca
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0\n"
```
ebauza, line 7: Should we change it to version 18 here @pedrobaeza ?

hr_appraisal_oca/i18n/es.po
```diff
@@ -0,0 +1,1201 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* hr_appraisal_oca
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 16.0\n"
```
ebauza, line 7: Should we change it to version 18 here @pedrobaeza ?

ebauza:
The rest is correct.

pedrobaeza:
@ebauza you have put "The rest is correct.", so that means that there is something that is not?

ebauza:
Sorry @pedrobaeza , I mentioned it because I added a comment about the version in the translations, but I don't know if it's relevant.

pedrobaeza:
@ebauza but the comment you have introduced is not "posted" or something, but it's not here...

ebauza (commented):
Sorry @pedrobaeza, I hadn't sent the review

hr_appraisal_oca/i18n/hr_appraisal_oca.pot
pedrobaeza: Well, that's only a comment, and although it would be great to have it "synchronized", it's not conditioning anything.

pedrobaeza:
Merging then

/ocabot merge nobump

PR #1528, opened 2025-11-28
[ADD] hr_job_no_of_employee: Backport from 19.0 core to 18.0

hitrosol:
This feature is taken from the core in odoo core 19.0


hitrosol:
not sure there is error like this:

errors that caused failure (1):
2025-11-28 04:22:07,666 298 WARNING odoo odoo.addons.attachment_indexation.models.ir_attachment: Attachment indexation of PDF documents is unavailable because the 'pdfminer' Python library cannot be found on the system. You may install it from https://pypi.org/project/pdfminer.six/ (e.g. `pip3 install pdfminer.six`) 
Error: Errors detected in log.

Error: Process completed with exit code 1.


I think my module doesn't related to pdminer ?

alexey-pelykh (commented):
Interesting backport — the description correctly notes this is native in 19.0, which is good.

CI tests are failing though, and the module doesn't include any tests of its own. Even for a backport, adding a basic test for the computed field and action would be worth it. Also, `CONTRIBUTORS.md` is empty.

PR #1536, opened 2026-01-24
[16.0][IMP] hr_employee_ppe: Using related fields

anusriNPS:
Using related fields for "is_ppe", "expire_ppe"
and "indications" instead of computing these fields based on product change from UI.

### DIFF
--- a/hr_employee_ppe/models/hr_personal_equipment.py
+++ b/hr_employee_ppe/models/hr_personal_equipment.py
@@ -14,11 +14,15 @@ class HrPersonalEquipment(models.Model):
     _name = "hr.personal.equipment"
     _inherit = ["hr.personal.equipment"]
 
-    is_ppe = fields.Boolean()
+    is_ppe = fields.Boolean(related="product_id.is_ppe", store=True)
     indications = fields.Text(
         help="Situations in which the employee should use this equipment.",
+        related="product_id.indications",
+        store=True,
     )
-    expire_ppe = fields.Boolean(help="True if the PPE expires")
+    expire_ppe = fields.Boolean(
+        help="True if the PPE expires", related="product_id.expirable_ppe", store=True
+    )
     certification = fields.Char(
         string="Certification Number", help="Certification Number"
     )
@@ -28,16 +32,6 @@ class HrPersonalEquipment(models.Model):
         res = super()._accept_request_vals()
         res["issued_by"] = self.env.user.id
         return res
-
-    @api.onchange("product_id")
-    def _compute_fields(self):
-        for rec in self:
-            if rec.product_id.is_ppe:
-                rec.is_ppe = rec.product_id.is_ppe
-                if rec.product_id.expirable_ppe:
-                    rec.expire_ppe = rec.product_id.expirable_ppe
-                if rec.product_id.indications:
-                    rec.indications = rec.product_id.indications
 
     def _validate_allocation_vals(self):
         res = super()._validate_allocation_vals()
--- a/hr_employee_ppe/tests/test_hr_employee_ppe.py
+++ b/hr_employee_ppe/tests/test_hr_employee_ppe.py
@@ -81,15 +81,6 @@ class TestHREmployeePPE(TransactionCase):
         cls.hr_employee_ppe_expirable = cls.personal_equipment_request.line_ids[0]
         cls.hr_employee_ppe_no_expirable = cls.personal_equipment_request.line_ids[1]
 
-    def test_compute_fields(self):
-        self.hr_employee_ppe_expirable._compute_fields()
-        self.assertTrue(self.hr_employee_ppe_expirable.is_ppe)
-        self.assertTrue(self.hr_employee_ppe_expirable.expire_ppe)
-        self.assertEqual(
-            self.hr_employee_ppe_expirable.indications,
-            self.product_employee_ppe_expirable.indications,
-        )
-
     def test_accept_allocation(self):
         self.assertFalse(self.hr_employee_ppe_expirable.issued_by)
         self.personal_equipment_request.with_user(self.user).accept_request()
@@ -139,10 +130,15 @@ class TestHREmployeePPE(TransactionCase):
         self.assertNotEqual(self.hr_employee_ppe_no_expirable.state, "expired")
 
     def test_check_dates(self):
+        self.hr_employee_ppe_expirable.start_date = "2020-01-01"
+        self.hr_employee_ppe_expirable.expiry_date = "2019-12-31"
+        self.assertTrue(self.hr_employee_ppe_expirable.is_ppe)
+        self.assertTrue(self.hr_employee_ppe_expirable.expire_ppe)
+        self.assertEqual(
+            self.hr_employee_ppe_expirable.indications,
+            self.product_employee_ppe_expirable.indications,
+        )
         with self.assertRaises(ValidationError):
-            self.hr_employee_ppe_expirable.start_date = "2020-01-01"
-            self.hr_employee_ppe_expirable.expiry_date = "2019-12-31"
-            self.hr_employee_ppe_expirable._compute_fields()
             self.hr_employee_ppe_expirable.validate_allocation()
 
     def test_compute_contains_ppe(self):


HekkiMelody (approved):
Code review, LGTM

quirino95 (approved):
LGTM!

marcelsavegnago (approved):
LGTM

hr_employee_ppe/models/hr_personal_equipment.py
```diff
@@ -14,11 +14,14 @@ class HrPersonalEquipment(models.Model):
     _name = "hr.personal.equipment"
     _inherit = ["hr.personal.equipment"]
 
-    is_ppe = fields.Boolean()
+    is_ppe = fields.Boolean(related="product_id.is_ppe")
```
alexey-pelykh: Consider adding `store=True` here. The old field was stored, and the tree view displays this field, so filtering/grouping on `is_ppe` in list views would require a stored field to work efficiently.

hr_employee_ppe/tests/test_hr_employee_ppe.py
```diff
@@ -142,7 +133,12 @@ def test_check_dates(self):
         with self.assertRaises(ValidationError):
             self.hr_employee_ppe_expirable.start_date = "2020-01-01"
             self.hr_employee_ppe_expirable.expiry_date = "2019-12-31"
-            self.hr_employee_ppe_expirable._compute_fields()
+            self.assertTrue(self.hr_employee_ppe_expirable.is_ppe)
```
alexey-pelykh: These assertions are inside the `assertRaises(ValidationError)` block. If any assertion fails, it raises `AssertionError` (not `ValidationError`), which would cause `assertRaises` to re-raise it with a confusing traceback. Consider moving these assertions before the `with` block:
```python
self.hr_employee_ppe_expirable.start_date = "2020-01-01"
self.hr_employee_ppe_expirable.expiry_date = "2019-12-31"
self.assertTrue(self.hr_employee_ppe_expirable.is_ppe)
self.assertTrue(self.hr_employee_ppe_expirable.expire_ppe)
...
with self.assertRaises(ValidationError):
    self.hr_employee_ppe_expirable.validate_allocation()
```

alexey-pelykh (commented):
Thanks for the improvement! Using related fields instead of `@api.onchange` is definitely the right direction -- related fields work consistently regardless of whether the change comes from the UI, ORM, or import, while `onchange` only fires from the UI.

A couple of things I'd like to discuss:

**1. `store` attribute on related fields**

The new related fields default to `store=False`. The old fields (`is_ppe`, `expire_ppe`, `indications`) were regular stored fields with database columns. With this change:

- Existing stored data in those columns becomes orphaned (not a functional problem, Odoo handles this, but worth being aware of).
- If any external module or report filters/groups by `is_ppe` or `expire_ppe` in a domain, it will become significantly slower or fail since unstored fields can't be efficiently searched. The tree view already displays `expire_ppe`, which means a list view with a filter on that field would hit this.

Have you considered adding `store=True` to the related fields? That would keep the DB column, allow efficient search/filter, and Odoo would auto-maintain the values through the related mechanism. The trade-off is that related stored fields add DB triggers, but for a Boolean and a Text field this is negligible.

**2. Behavioral change on `expire_ppe` editability**

In the current code, `expire_ppe` is a regular field with the view allowing edits when `state in ['draft', 'accepted']`. The form even has `readonly="state not in ['draft','accepted']"` on `expire_ppe`, implying it was intended to be editable per-record. With a related field (which is `readonly=True` by default at ORM level), users can no longer override `expire_ppe` on a specific equipment record independently of the product.

If overriding per-record was not actually needed (which seems to be the case from the original onchange logic), then this is fine -- just wanted to confirm the intent.

**3. Minor: Test assertions inside `assertRaises` block**

In `test_check_dates`, the new assertions (`assertTrue`, `assertEqual`) are placed inside the `with self.assertRaises(ValidationError)` block before the call to `validate_allocation()`. If any of those assertions fail, they'll raise `AssertionError` (not `ValidationError`), which would cause the test to fail with a confusing error message. Consider moving those assertions before the `assertRaises` context manager, or into a separate test.

hr_employee_ppe/models/hr_personal_equipment.py
anusriNPS: Restored `store=True` for fields which are modified as `related`

hr_employee_ppe/tests/test_hr_employee_ppe.py
anusriNPS: Moved assert checks outside `assertRaises` block as suggested

anusriNPS:
Depends on #1550 which resolves error seen in pre-commit job

HekkiMelody:
#1550 has been merged, so we should be able to merge this as well. The PR has enough approvals.

Thank you!

marcos-mendez (changes requested):
**Automated Review -- Tests Failed**

### 1. **Root Cause**

The test failure occurs because the `related` fields (`is_ppe`, `indications`, `expire_ppe`) in `hr_employee_ppe/models/hr_personal_equipment.py` are not properly initialized when the record is created, due to missing `store=True` and lack of proper initialization in the test setup. This causes a database access error when the system tries to read these computed fields during test execution.

---

### 2. **Suggested Fix**

In `hr_employee_ppe/models/hr_personal_equipment.py`, ensure that the related fields are correctly initialized by verifying that the `product_id` is set before the record is saved, and that `store=True` is used correctly. Specifically:

- Ensure that `product_id` is properly set in test data.
- Add explicit initialization of related fields in `setUpClass` if needed.

**File:** `hr_employee_ppe/models/hr_personal_equipment.py`  
**Lines:** 17–20  
**Change:** No code change required here, but make sure that in tests, `product_id` is set correctly before any related field access.

**File:** `hr_employee_ppe/tests/test_hr_employee_ppe.py`  
**Lines:** 84–87 (remove `test_compute_fields`)  
**Addition:** Add `product_id` assignment in `setUpClass` to ensure related fields are initialized.

```python
# In setUpClass, ensure:
self.hr_employee_ppe_expirable.product_id = self.product_employee_ppe_expirable
self.hr_employee_ppe_no_expirable.product_id = self.product_employee_ppe_no_expirable
```

---

### 3. **Additional Code Issues**

- **No issue detected** in the current diff. The use of `related` fields with `store=True` is valid and follows Odoo conventions.
- The removal of `_compute_fields()` is correct since it's replaced by `related` fields.
- No issues with modifiers, hooks, or extensibility patterns.

---

### 4. **Test Improvements**

To improve test coverage and ensure robustness, add the following test cases:

#### Suggested Test Cases:

1. **Test related field initialization:**
   - Ensure that `is_ppe`, `expire_ppe`, and `indications` are correctly set from `product_id` when a record is created.
   - **Pattern:** Use `TransactionCase` or `SavepointCase` to test data integrity.

2. **Test that `related` fields update when `product_id` changes:**
   - Change `product_id` on a record and verify that related fields update accordingly.
   - **Pattern:** Use `self.assertEqual(...)` to assert field values after change.

3. **Test edge case: `product_id` is `False`:**
   - Ensure that related fields default to `False` or `None` when `product_id` is not set.
   - **Pattern:** Test with empty product.

#### Example Test Snippet:

```python
def test_related_fields_update_on_product_change(self):
    self.hr_employee_ppe_expirable.product_id = self.product_employee_ppe_no_expirable
    self.assertFalse(self.hr_employee_ppe_expirable.is_ppe)
    self.assertFalse(self.hr_employee_ppe_expirable.expire_ppe)
    self.assertEqual(self.hr_employee_ppe_expirable.indications, False)
```

#### OCA Testing Patterns:
- Use `SavepointCase` for tests involving data changes or complex validation logic.
- Use `TransactionCase` for tests that require transactional integrity.

---

✅ **Conclusion:**  
The test failure is due to improper test data setup (missing `product_id` assignment). Fix by ensuring `product_id` is set during test initialization. No functional issues in the code changes.

---
⏰ This PR has been open for **50 days**.

> _Every ignored PR is a contributor who might not come back. Review time matters. ([OCA Aging Report](https://git.pop.coop/pop/oca-reviewer))_

---
### Reciprocal Review Request



Hi everyone! I found some test failures on this PR and left detailed feedback above. I am happy to discuss or help debug. In the meantime, if any of you get a chance, I would appreciate a look at my open PR(s):

**My open PRs across OCA:**
- [server-tools#3554 [MIG] datetime_formatter: Migration to 18.0](https://github.com/OCA/server-tools/pull/3554)
- [server-tools#3548 [18.0][MIG] base_kanban_stage: Migration to 18.0](https://github.com/OCA/server-tools/pull/3548)
- [hr-attendance#262 [16.0][ADD] Hr_attendance_idsecure: iDSecure (ControliD) attendance integration](https://github.com/OCA/hr-attendance/pull/262)
- [stock-logistics-workflow#2276 [16.0][ADD] stock_move_line_devaluation](https://github.com/OCA/stock-logistics-workflow/pull/2276)
- [stock-logistics-workflow#2275 [16.0][ADD] Stock move line analytic account](https://github.com/OCA/stock-logistics-workflow/pull/2275)
- [stock-logistics-workflow#2268 [16.0][ADD] stock_move_line_picking_partner](https://github.com/OCA/stock-logistics-workflow/pull/2268)
- [purchase-workflow#2694 [16.0][IMP]Purchase workflow added to review state & exception fix](https://github.com/OCA/purchase-workflow/pull/2694)

Reviewing each other's work helps the whole community move forward. Thank you!

---
_Environment via [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer): Minikube + K8s Job + oca-ci/py3.10-odoo16.0 | Odoo 16.0_
_Automated review by [OCA Neural Reviewer](https://git.pop.coop/pop/oca-reviewer) + qwen3-coder:30b_

HekkiMelody:
@OCA/human-resources-maintainers Could you take a look? The PR has several reviews already. We will FW if applicable. Thank you

marcos-mendez:
@marcelsavegnago 

Agreed for is_ppe and indications, but I'd ask that this pattern not be extended to certification later on: an allocation records what was handed to a named worker on a given date, so a related field would silently rewrite past deliveries every time the product's certificate is renewed. A safety inspection asks which certificate was valid at the moment of delivery — that is the premise of NR-6 in Brazil, and I'd expect any jurisdiction that inspects PPE certificates to work the same way — so that value has to be frozen on the allocation rather than read live from the product. Would it make sense to state this explicitly, keeping certification a stored field defaulted from the product?

marcos-mendez:
is_ppe makes sense as a related field, but indications is part of the delivery record rather than product data: with store=True it will be recomputed over past allocations whenever the product text is edited, and it also drops the previous behaviour where a value typed on the allocation survived, since the old onchange only copied non-empty values

marcelsavegnago:
ping @etobella

hr_employee_ppe/models/hr_personal_equipment.py
```diff
@@ -14,11 +14,15 @@ class HrPersonalEquipment(models.Model):
     _name = "hr.personal.equipment"
     _inherit = ["hr.personal.equipment"]
 
-    is_ppe = fields.Boolean()
+    is_ppe = fields.Boolean(related="product_id.is_ppe", store=True)
```
etobella, line 17: The problem with this implementation is that if we change the value in the product, all equipments will be changed. For me we should find another way (a compute that depends only in product_id, not in product_id.is_ppe for example

anusriNPS: Thank you for your views. Will verify compute of fields using @api.depends works for this case and handle accordingly.

marcos-mendez: Tks man!

anusriNPS: As this PR is already use in production, I have created a seperate PR #1624 to address review comments. I have updated fields as computed fields which depends on `product_id`. Once reviews are received, will update this PR accordingly and close the redundant PR.

PR #1537, opened 2026-01-25
[18.0][FIX] hr_contract_employee_calendar_planning: avoid errors with…

FernandoRomera:
… hr_leave


alexey-pelykh (approved):
Good catch. Using `with_context(no_leave_resource_calendar_update=True)` to prevent the unwanted side effect makes sense. Clean fix, LGTM

ChristophAbenthungCibex (approved):
LGTM

pedrobaeza (changes requested):
I'm afraid I need a bit more context. Can you add in the extended commit message and in the PR description steps to reproduce the problem and why this fixes it? It would be also very interesting to add a regression test for this.

PR #1540, opened 2026-01-27
[19.0]MIG hr_employee_calendar_planning

Vicent-C6:
This pr has significant changes, refering some changes from odoo [#217895](https://github.com/odoo/odoo/pull/217895/changes#diff-c795ec590c418cf1f8775b1e782811dbac61063693e0e5a44574058996ca84b2)

That leads to override _get_work_days_data_batch to change the calendar and the hr.leave

Adjust the tests


Reyes4711-S73 (approved):
LGTM
Functional review ok

sbiosca-s73 (approved):

Reyes4711-S73:
@OCA/human-resources-maintainers Please, can you review/merge this PR?

Reyes4711-S73 (approved):
LGTM

alexey-pelykh (approved):
Clean migration. Manifest, scaffolding and translations are in order.

Code review LGTM.

Vicent-S73:
@pedrobaeza could this pr be merged? Has a few approves and it's blocking #1548 due to problems with the test-requierements

hr_employee_calendar_planning/__manifest__.py
```diff
@@ -2,18 +2,17 @@
...
     "name": "Employee Calendar Planning",
-    "version": "18.0.1.0.1",
+    "version": "19.0.1.0.0",
     "category": "Human Resources",
     "website": "https://github.com/OCA/hr",
     "author": "Tecnativa,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "installable": True,
-    "depends": ["hr"],
+    "depends": ["hr", "hr_holidays"],
```
victoralmau, line 11: Why has the `hr_holidays` dependency been added now?

In my opinion, if there's something to be covered in hr_holidays, an additional module should be created (`hr_holidays_calendar_planning`, for example).

pedrobaeza: Yes, I didn't check that, sorry. The dependency should be dropped. Please do it.

hr_employee_calendar_planning/hooks.py
carlos-lopez-tecnativa, line 1: Why did you remove this hook?

Note that a migration script is not the same as a hook. If you want to replace this ORM logic with SQL in the migration script, keep in mind that the migration script is not executed when the module is installed. It is only executed when migrating from a minor version to a major version.

Also, in a fresh database installed on V19, I noticed that the employee calendars are not populated as they were in V18.

In V19, the automatic calendars are not created for existing employees, whereas in V18 they are. Are these calendars no longer necessary in V19?

This is confusing to me because when I create a new employee, the automatic calendar is created, but it is not created for existing employees.

Could you please review this and clarify the reason for this behavior?

Image

carlos-lopez-tecnativa (commented):
ping @Vicent-C6 
CC: @pedrobaeza @victoralmau

hr_employee_calendar_planning/hooks.py
victoralmau: In my opinion, the hook is still necessary, although since the `date_from` and `date_to` fields (in `resource.calendar.attendance`) no longer exist, it’s likely been assumed that the hook isn’t needed.

I’m preparing the necessary change to restore it.

hr_employee_calendar_planning/migrations/19.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2026 Tecnativa - Vicent Castells
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(cr, version):
```
victoralmau, line 8: I don't understand the need for this migration script.

Currently, the `date_from` and `date_to` columns do not exist, but they did in previous versions; if this module was being used in previous versions, the `calendar_ids` field in `hr.employee` will contain the correct data; therefore, I don't think any additional action is necessary.

Ping @pedrobaeza and @carlos-lopez-tecnativa

pedrobaeza: Yes, it seems so, so it can be removed.

PR #1541, opened 2026-01-27
[18.0][ADD] hr_work_entry_contract_attendance_oca

huguesdk:
add new module `hr_work_entry_contract_attendance_oca` that allows to generate work entries from attendances.


mihien (approved):
Lgtm, only got one question/remark but nothing blocking approval.

hr_work_entry_contract_attendance_oca/models/hr_contract.py
```diff
@@ -0,0 +1,113 @@
...
+            # difference with the attendances first return the attendance
+            # intervals minus the lunch break intervals from the contracts, but
+            # still linked to the hr.attendance record.
+            resource_id = contract.employee_id.resource_id.id
+            result[resource_id] = (
+                hr_attendance_intervals - lunch_contract_intervals[resource_id]
+            )
+        return result
+
+    def _get_interval_work_entry_type(self, interval):
```
mihien, line 103: Can typing be used in those kind of case? I haven't ever seen typing being used in odoo modules, but in my opinion it would benefit from it, clarity-wise.

```suggestion
    def _get_interval_work_entry_type(self, interval: WorkIntervals):
```

huguesdk: yes, indeed, typing could be used. this is far from being the norm in odoo development (i didn’t encounter yet any type hints in the odoo codebase itself), but i’ve already seen some in oca(-like) modules, like [here](https://github.com/acsone/odoo-peppol-backport/blob/16.0/account_peppol_send_format_odoo/wizard/account_invoice_send.py). there is nothing about it in [oca’s guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst), though.

alexey-pelykh (changes requested):
Thanks for this -- generating work entries from attendances fills a real gap. The _get_attendance_intervals override with lunch break subtraction logic is well thought out, and the WorkIntervals usage is correct.

The test file only contains TODO comments with no actual test methods though. At minimum, a test verifying that work entries get generated from an attendance with check_in/check_out would be needed. The attendance domain logic and lunch break handling are the critical paths to cover.

Happy to re-review once there's at least basic test coverage.

huguesdk:
@alexey-pelykh hey, thanks for the review! i agree, there are still some important parts missing (as explained in the roadmap), but it is already partly usable now, using the “regenerate work entries” wizard. do you think i should convert the pr to draft? moreover, if you need these missing functions, feel free to take over in another pr.

alexey-pelykh:
@huguesdk Thanks for the context! I would say keep it as a non-draft — it is already partially usable with the regenerate wizard, and having it visible encourages others to build on it. The roadmap in the description makes the scope clear.

If I get a chance to contribute the missing pieces I will open a follow-up PR. Cheers!

PR #1544, opened 2026-02-04
[18.0][ADD] hr_collective_agreement

Anxo82:
@HaraldPanten @Jaimermaccione 

- [T-8786]


hr_collective_agreement/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2025 Ángel Rivas <angel.rivas@sygel.es>
```
NL66278, line 1: There should be no copyright in init files

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class CollectiveAgreement(models.Model):
+    _name = "collective.agreement"
+    _description = "Collective Agreement"
+    _check_company_auto = True
+
+    code = fields.Char(required=True)
+    name = fields.Text(required=True)
+    scope_id = fields.Many2one(
```
NL66278, line 15: What is this scope model and field for? A definition and/or explanation in the readme DESCRIPTION, and maybe in the help for this field might be useful.

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    active = fields.Boolean(default=True)
+    partner_ids = fields.Many2many(
+        comodel_name="res.partner",
+        relation="collective_agreement_partner_rel",
+        column1="agreement_id",
+        column2="partner_id",
+        string="Partners",
+        check_company=True,
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
```
NL66278, line 46: Or you might set `_check_company_auto = True` on the class level, to have this check automatic for all model references that might contain a company_id field.

NL66278 (changes requested):
👍 Thank you for your contribution. Just purge copyright lines from the init files and add some help or explanation and it will be good to go.

NL66278 (approved):
👍 Actually I am not quite sure whether the domain on the partner_id field is still needed with _check_company_auto, but it will not hurt anyway. Thanks again!

hr_collective_agreement/models/collective_agreement.py
Anxo82: Added a help text to the scope_id field and a short explanation of the Scope and Official Publication fields in the README. Thanks

Jaimermaccione (changes requested):
@Anxo82 


---
The module has been reviewed functionally in a local environment. It works as expected and no issues were found in normal usage. Below are a few improvement ideas mainly related to **usability** and **basic access control**.
#### 1. Unified search
**Current situation**  
Users need to choose whether they want to search by *code* or *name*.

**Proposal**  
Use a single search that checks both fields at the same time.

**Value**  
Faster and more comfortable to use, especially when there are many agreements.

---

#### 2. State management
**Current situation**  
The *state* field is shown as a regular selection field inside the form.

**Proposal**  
Move it to the header and show it as a status bar.

**Extra note**  
In the current Spanish and Catalan translations, the *state* field is translated as **“province”**. It should simply be **“state”**.

**Value**  
Makes it easier to understand the agreement status at a glance and follows the usual Odoo patterns.

---

#### 3. Access rights
**Current situation**  
Any internal user can create or edit collective agreements.

**Proposal**  
Limit creation and editing to **HR admin users**, keeping **read access for all internal users**.

**Optional**  
Adding separate configuration menus (scope, official publication) would be nice, but if that’s too much, the access rights change alone already helps a lot.

**Value**  
Prevents accidental changes and ensures only users with the right level of access can modify this data.

---

#### Conclusion
Functionally, the module is fine. These are small improvements that would make it clearer and safer to use.

Anxo82:
@Jaimermaccione All requested changes have been applied. Could you please review again?
Thanks

Jaimermaccione (approved):
@Anxo82 The module has been functionally reviewed in a local environment and tested according to the requested changes.

Verified points:
- Unified search: search by code and name works correctly.
- State management: the state is shown as a status bar and behaves as expected.
- Access rights: creation and editing are correctly restricted by permissions.
- Automatic state updates: state changes based on publication date and expiration date work correctly.
- Translations: the “state” translation issue has been fixed.

All tested scenarios behave as expected.

**Conclusion**  
LGTM. Good job — the module is in a very solid shape.

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+    publication_date = fields.Date(required=True)
+    end_date = fields.Date()
+    official_publication_id = fields.Many2one(
+        "collective.agreement.official.publication",
+        string="Official Publication",
+        required=True,
+    )
+    observations = fields.Text()
+    state = fields.Selection(
+        [
```
ValentinVinagre: add the parameter name

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+
+    code = fields.Char(required=True)
+    name = fields.Text(required=True)
+    scope_id = fields.Many2one(
+        "collective.agreement.scope", string="Scope", required=True
+    )
+    publication_date = fields.Date(required=True)
+    end_date = fields.Date()
+    official_publication_id = fields.Many2one(
+        "collective.agreement.official.publication",
```
ValentinVinagre: add the parameter name

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+
+class CollectiveAgreement(models.Model):
+    _name = "collective.agreement"
+    _description = "Collective Agreement"
+    _check_company_auto = True
+
+    code = fields.Char(required=True)
+    name = fields.Text(required=True)
+    scope_id = fields.Many2one(
+        "collective.agreement.scope", string="Scope", required=True
```
ValentinVinagre: add the parameter name

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+        comodel_name="res.partner",
+        relation="collective_agreement_partner_rel",
+        column1="agreement_id",
+        column2="partner_id",
+        string="Partners",
+        check_company=True,
+        domain="['|', ('company_id', '=', False), ('company_id', '=', company_id)]",
+    )
+    company_id = fields.Many2one(
+        "res.company",
```
ValentinVinagre: add the parameter name

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+                    _("The end date cannot be earlier than the publication date.")
+                )
+
+    def action_draft(self):
+        self.write({"state": "draft"})
+
+    def action_activate_agreement(self):
+        today = fields.Date.today()
+        for record in self:
+            if record.end_date and record.end_date < today:
```
ValentinVinagre: First, filter the records that have end_date in the for loop, and then in the if statement, compare it with today or add it to the filtered statement as well.

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def action_finish(self):
+        self.write({"state": "finished"})
+
+    def action_cancel(self):
+        self.write({"state": "cancelled"})
+
+    def action_update_states(self):
+        today = fields.Date.today()
+
```
ValentinVinagre: delete empty lines.

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,107 @@
...
+                raise ValidationError(
+                    _(
+                        "The collective agreement cannot be activated because its "
+                        "end date has already passed."
+                    )
+                )
+            record.state = "active"
+
+    def action_finish(self):
+        self.write({"state": "finished"})
```
ValentinVinagre, line 106: Is it not necessary to check if it's in the corresponding states for the change to be made?

hr_collective_agreement/views/collective_agreement_views.xml
```diff
@@ -0,0 +1,153 @@
...
+                        groups="hr.group_hr_manager"
+                    />
+                </header>
+                <sheet>
+                    <group>
+                        <field name="code" />
+                        <field name="name" />
+                        <field name="scope_id" />
+                        <field name="publication_date" />
+                        <field name="end_date" />
```
ValentinVinagre: Wouldn't it look better with the date widget?

ValentinVinagre (changes requested):

hr_collective_agreement/models/collective_agreement.py
Anxo82: Added

hr_collective_agreement/models/collective_agreement.py
Anxo82: Added

hr_collective_agreement/models/collective_agreement.py
Anxo82: Added

hr_collective_agreement/models/collective_agreement.py
Anxo82: Added

hr_collective_agreement/models/collective_agreement.py
Anxo82: I didn’t add the state check because the Finish button is only visible when the agreement is in the Active state. But I’ve now added the validation at model level as suggested, just to make it safer.

hr_collective_agreement/models/collective_agreement.py
Anxo82: Done

hr_collective_agreement/views/collective_agreement_views.xml
Anxo82: Done

Image

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,111 @@
...
+    def action_draft(self):
+        self.write({"state": "draft"})
+
+    def action_activate_agreement(self):
+        today = fields.Date.today()
+        expired = self.filtered(lambda r: r.end_date and r.end_date < today)
+        if expired:
+            raise ValidationError(
+                _(
+                    "The collective agreement cannot be activated because its "
```
ValentinVinagre: If more than one register fails, how do you know which one is failing?

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,111 @@
...
+                    "end date has already passed."
+                )
+            )
+        self.write({"state": "active"})
+
+    def action_finish(self):
+        invalid = self.filtered(lambda r: r.state != "active")
+        if invalid:
+            raise ValidationError(_("Only active agreements can be finished."))
```
ValentinVinagre: If more than one register fails, how do you know which one is failing?

hr_collective_agreement/models/collective_agreement.py
```diff
@@ -0,0 +1,111 @@
...
+        self.write({"state": "finished"})
+
+    def action_cancel(self):
+        self.write({"state": "cancelled"})
```
ValentinVinagre, line 118: Here, states are not checked, or can anyone cancel from anywhere?

ValentinVinagre (changes requested):

hr_collective_agreement/models/collective_agreement.py
Anxo82: State validation has been added. Only draft and active agreements can be cancelled.

Jaimermaccione (changes requested):
@Anxo82 Hi Anxo,

I have functionally reviewed the new mass actions for cancelling and/or finishing collective agreements.

From a functional perspective, both actions work correctly and behave as expected depending on the agreement state. Good job on that.

#### 1. README clarification (mass actions behavior)

I would suggest refining the README to better specify how the mass actions behave depending on the agreement state.

Instead of:

> Additionally, multiple agreements can be selected from the list view and  
> processed simultaneously using the action menu.

I would propose something more explicit, such as:

> Additionally, multiple agreements can be selected from the list view and  
> processed simultaneously using the action menu.  
>  
> From this action menu, users can:  
>  
> - Cancel agreements that are in *Draft* or *Active* state.  
> - Finish agreements that are in *Active* state only.

Please review this wording in case anything does not exactly match the implemented behavior.

As a small UX improvement suggestion:  
currently, when selecting agreements in mixed states, the action is shown and a warning is displayed if some records are not eligible, which is perfectly fine.

As a possible enhancement, it might be worth considering enabling or displaying mass actions only when all selected agreements are in compatible states (or showing them disabled with a tooltip). This would make the behavior more predictable and prevent invalid selections upfront.

That said, since the current behavior is already correct, this would only make sense if it can be implemented as a very lightweight change; otherwise, the existing approach is perfectly acceptable.

#### 2. Duplication behavior (small usability improvement)

Apologies if this goes slightly beyond the current scope.

While testing the feature, I needed to duplicate some agreements and noticed that duplication is currently blocked because the `code` field cannot be repeated (which is correct).

However, it might improve usability if, when duplicating a record, the `code` is not copied to the new record. This would allow users to reuse most of the configuration values while still being forced to define a new unique code.

Could you please review whether it would be feasible to override the duplication behavior so that `code` is cleared on copy?

Let me know your thoughts.

Thanks!

Anxo82:
Hi @Jaimermaccione ,

Thank you for the suggestions.

- The README has been updated accordingly. 
- The possibility to duplicate agreements has been included. Since the fields ``code`` and ``name`` are required and cannot be empty, duplicating a record while leaving them blank is not possible. Therefore, in order to allow duplication, the suffix “(copy)” is added to both fields. This follows the usual Odoo approach, as required fields cannot remain empty during record creation.

- Regarding the mass actions to cancel and finish agreements from the tree view: while this could slightly improve the user experience (UX), implementing dynamic availability for mass actions would require additional complexity (e.g., a wizard or custom logic). Since the current behavior already prevents invalid state transitions safely at the backend level, I would prefer to keep the implementation lightweight for now.

Thank you

Jaimermaccione (approved):
@Anxo82 Thank you for the update.

I have reviewed the latest changes and everything looks good to me

- I have reviewed the README modification — it looks perfect as it is now.
- Regarding the duplication behavior: appending “(copy)” to both code and name is perfectly reasonable and consistent with standard Odoo patterns for required fields. I agree with this solution.
- As for the mass actions, I also agree that keeping the current lightweight implementation makes sense, given that invalid transitions are already safely handled at the backend level.

From my side, **LGTM.** Thank you!

hr_collective_agreement/models/collective_agreement_publication.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class CollectiveAgreementOfficialPublication(models.Model):
+    _name = "collective.agreement.official.publication"
+    _description = "Collective Agreement Official Publication"
+
+    name = fields.Char(required=True)
+
```
ValentinVinagre, line 12: Is the copying not done here?

hr_collective_agreement/models/collective_agreement_scope.py
```diff
@@ -0,0 +1,13 @@
...
+
+from odoo import fields, models
+
+
+class CollectiveAgreementScope(models.Model):
+    _name = "collective.agreement.scope"
+    _description = "Collective Agreement Scope"
+
+    name = fields.Char(required=True)
+
```
ValentinVinagre, line 12: Is the copying not done here?

hr_collective_agreement/models/collective_agreement_publication.py
Anxo82: In this model there is no unique constraint on the ``name``  field, so duplication works without any issue and no custom copy logic is required. Moreover, these values are meant to be reused across multiple agreements, so enforcing uniqueness would not be appropriate in this case.

hr_collective_agreement/models/collective_agreement_scope.py
Anxo82: In this model there is no unique constraint on the ``name``  field, so duplication works without any issue and no custom copy logic is required. Moreover, these values are meant to be reused across multiple agreements, so enforcing uniqueness would not be appropriate in this case.

ValentinVinagre (approved):
LGTM 👍🏻

Anxo82:
@HaraldPanten Could you review this PR  when you have a moment?
Thanks!

HaraldPanten (approved):
/ocabot merge nobump

Jaimermaccione:
@Anxo82  Great job on this module!

@NL66278  @ValentinVinagre  @HaraldPanten , your input and support have been very valuable in moving this forward and getting it to a solid state.

Thanks to everyone involved for the collaboration and effort!

PR #1545, opened 2026-02-05
[18.0][ADD] hr_collective_agreement_partner

Anxo82:
@HaraldPanten @Jaimermaccione 

- This module depends on the hr_collective_agreement module (https://github.com/OCA/hr/pull/1544)

- [T-8786]


hr_collective_agreement_partner/views/res_partner_views.xml
```diff
@@ -0,0 +1,48 @@
...
+                    invisible="collective_agreement_count == 0"
+                    string="Collective Agreements"
+                >
+                    <field name="collective_agreement_count" widget="statinfo" />
+                </button>
+            </xpath>
+        </field>
+    </record>
+
+    <record id="action_partner_collective_agreements" model="ir.actions.act_window">
```
NL66278: This should go into its own xml file, either `collective_agreement_views.xml` or you might op to have all actions together in `ir_actions_act_window.xml`.

hr_collective_agreement_partner/views/res_partner_views.xml
```diff
@@ -0,0 +1,48 @@
+<odoo>
+    <record id="view_res_partner_form_inherit_collective_agreement" model="ir.ui.view">
```
NL66278: id's should remain the same as the inherit_id. However you can add the module name to the name of the view, to easily see where a view comes from in the UI.

So:
```
    <record id="view_partner_form" model="ir.ui.view">
        <field name="name">res.partner.form - hr_collective_agreement_partner</field>
```

Same for all other inherited views. Check https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#id29

The "rule" for id is in the guidelines, the way to use the name field is my own practice, but found it helpful to find original and inheriting views together.

NL66278 (changes requested):
👍 Thank you also for this module. Again a few changes would help adoption.

Anxo82:
@NL66278 the ir.actions.act_window has been moved to its own XML file, and inherited views now keep the original XML-IDs, with the view name updated to include the module name as suggested.
Thanks.

NL66278 (approved):
👍 LGTM

Jaimermaccione (changes requested):
@Anxo82 
The module has been reviewed functionally in a local environment.

The link between **company-type partners** and **collective agreements** has been tested and works as expected.

#### Contact smart button
As a possible improvement, assuming the current access flow is kept (only **HR / HR Admin users** can create or edit collective agreements), it may be worth adjusting the behavior of the smart button on contacts:

- The smart button should also be visible when the company has **0 linked agreements**, allowing any user (or users with contact permissions) to **assign an already existing agreement**.
- From this list, users **should not be able to create new agreements**, only link an existing one.

If this approach is technically feasible, it would provide a more consistent user experience and reduce confusion between creating agreements and assigning them.

#### README – usage section
As an additional improvement, it would be good to review and enhance the **README usage section**.  
Based on the changes that may be introduced following these requests, the README could include a clearer and more explicit “usage” section for the module.

Thank you!

Anxo82:
@Jaimermaccione Changes applied.
The smart button is now always visible on the partner form. From the partner’s collective agreements list, existing agreements can be assigned through a dedicated wizard.
Could you review it again? 
Thanks

Jaimermaccione:
@Anxo82 I have tested the new changes and the behavior matches my expectations.

I have a couple of comments:

#### 1. README – "Authorized users"

In the README you mention:

> From the collective agreements list of a contact, authorized users can assign additional collective agreements using the **“Select Agreement”** button.

Is this restricted to a specific user group? I have not detected any particular access restriction.

I think it is fine if any internal user can assign a collective agreement. If there is no actual group limitation, it might be better to remove the reference to **“authorized users”** from the README to avoid confusion.

---

#### 2. Existing assignments are preserved

When it states:

> Existing assignments are preserved.

I understand that collective agreements cannot be unlinked from this view, correct?

Would it be possible to add a column allowing users to unlink them? In other standard Odoo models, there is usually a column with an **“X”** or a trash icon to remove the relation.

I believe it would be appropriate for a user to be able to manage the unlinking of a collective agreement from a contact as well.

Please review this and assess whether the implementation would be reasonably simple.

---

Thank you.

Anxo82:
@Jaimermaccione changes done. Could you review it again?
Thanks

Jaimermaccione (approved):
@Anxo82 I have functionally reviewed the previously discussed changes in a local environment.

The updates have been correctly implemented. It is now possible to remove collective agreements from contacts that already have them assigned.

Additionally, the README wording has been improved to avoid confusion regarding user permissions.

Everything has been verified and works as expected.

LGTM 👍

Thank you!

hr_collective_agreement_partner/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    collective_agreement_ids = fields.Many2many(
+        comodel_name="collective.agreement",
+        relation="collective_agreement_partner_rel",
+        column1="partner_id",
+        column2="agreement_id",
+        string="Collective Agreements",
+        domain=[("active", "=", True)],
```
ValentinVinagre: Is this domain not the default?

hr_collective_agreement_partner/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+            "type": "ir.actions.act_window",
+            "name": "Collective Agreements",
+            "res_model": "collective.agreement",
+            "view_mode": "list",
+            "view_id": self.env.ref(
+                "hr_collective_agreement_partner.view_collective_agreement_tree_partner"
+            ).id,
+            "domain": [
+                ("partner_ids", "in", [self.id]),
+                ("active", "=", True),
```
ValentinVinagre: Is this domain not the default?

hr_collective_agreement_partner/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def action_view_collective_agreements(self):
+        self.ensure_one()
+        return {
+            "type": "ir.actions.act_window",
+            "name": "Collective Agreements",
+            "res_model": "collective.agreement",
+            "view_mode": "list",
+            "view_id": self.env.ref(
+                "hr_collective_agreement_partner.view_collective_agreement_tree_partner"
```
ValentinVinagre, line 35: Wouldn't it be better to instantiate it first in case it doesn't exist?

hr_collective_agreement_partner/tests/test_hr_collective_agreement_partner.py
```diff
@@ -0,0 +1,127 @@
...
+            action["domain"],
+        )
+
+        self.assertIn(("active", "=", True), action["domain"])
+
+        self.assertEqual(
+            action["context"]["partner_id"],
+            self.partner.id,
+        )
+
```
ValentinVinagre: delete empty

hr_collective_agreement_partner/wizard/collective_agreement_assign_wizard.py
```diff
@@ -0,0 +1,22 @@
...
+    _name = "collective.agreement.assign.wizard"
+    _description = "Assign Collective Agreements to Partner"
+
+    partner_id = fields.Many2one(
+        "res.partner",
+        required=True,
+        readonly=True,
+    )
+
+    agreement_ids = fields.Many2many(
```
ValentinVinagre: Why is it called 'agreements_ids' here and 'collective_agreement_ids' everywhere else?

ValentinVinagre (changes requested):

hr_collective_agreement_partner/models/res_partner.py
Anxo82: Correct, the explicit domain was redundant and has been removed.

hr_collective_agreement_partner/models/res_partner.py
Anxo82: Correct, the explicit domain was redundant and has been removed.

hr_collective_agreement_partner/models/res_partner.py
Anxo82: I believe the view belongs to this module (or to a required dependency), so it is guaranteed to exist. Therefore, using env.ref() directly is appropriate here.

hr_collective_agreement_partner/tests/test_hr_collective_agreement_partner.py
Anxo82: Done.

hr_collective_agreement_partner/wizard/collective_agreement_assign_wizard.py
Anxo82: Done. The field has been renamed to collective_agreement_ids for consistency with the rest of the module.

ValentinVinagre (approved):
LGTM 👍🏻

Anxo82:
@HaraldPanten  Could you review this PR when you have a moment?
Thanks!

HaraldPanten (approved):
/ocabot merge nobump

PR #1546, opened 2026-02-09
[19.0][FIX] hr_employee_fistname error with permisions

Vicent-S73:
If the user are not in hr.group_hr_user but is in other group like hr_expense.group_hr_expense_manager if you try to go in,  it raise this error because the fields defined in hr.employee are not defined in the public model and are not reachable

image


flotho (changes requested):
Hi, 
thanks for this PR.
I don't thinks you need to bump teh release.
I think that's the CI/CD command that will bump the release.

alexey-pelykh (approved):
Good catch. hr.employee.public needs these related fields to avoid access errors when non-HR users view employee data. Clean fix, CI is green. LGTM.

Minor: copyright header references Studio73 but the URL points to onestein.eu.

flotho (approved):
easy code review
LGTM

PR #1547, opened 2026-02-14
[17.0][MIG] hr_employee_document_from_applicant: Migration to 17.0

pilarvargas-tecnativa:
Move attachment transfer logic from hr.employee.create to hr.applicant.create_employee_from_applicant
to avoid unnecessary execution when creating employees outside recruitment workflow.
No behavior change, only performance optimization.

@Tecnativa TT60569

@victoralmau @pedrobaeza please review


victoralmau (approved):
Code and functional review OK

hr_employee_document_from_applicant/models/hr_applicant.py
```diff
@@ -0,0 +1,43 @@
...
+
+from odoo import _, models
+
+
+class HrApplicant(models.Model):
+    _inherit = "hr.applicant"
+
+    def create_employee_from_applicant(self):
+        action = super().create_employee_from_applicant()
+        if self.emp_id:
```
pedrobaeza: Please use instead returned action (`action["res_id"]`)

hr_employee_document_from_applicant/models/hr_applicant.py
```diff
@@ -0,0 +1,43 @@
...
+                .search(
+                    [
+                        ("res_model", "=", "hr.applicant"),
+                        ("res_id", "=", self.id),
+                    ]
+                )
+            )
+            if applicant_attachments:
+                for attachment in applicant_attachments:
+                    attachment.write(
```
pedrobaeza: You have to make a copy, not to move them.

hr_employee_document_from_applicant/models/hr_applicant.py
```diff
@@ -0,0 +1,43 @@
...
+                        {
+                            "res_model": "hr.employee",
+                            "res_id": self.emp_id.id,
+                        }
+                    )
+                self.emp_id.message_post(
```
pedrobaeza: Then, this message is not needed.

hr_employee_document_from_applicant/models/hr_applicant.py
pilarvargas-tecnativa: This is a change I have added as an improvement, not as part of the migration.

hr_employee_document_from_applicant/models/hr_applicant.py
```diff
@@ -0,0 +1,33 @@
...
+                self.env["ir.attachment"]
+                .sudo()
+                .search(
+                    [
+                        ("res_model", "=", "hr.applicant"),
+                        ("res_id", "=", self.id),
+                    ]
+                )
+            )
+            if applicant_attachments:
```
pedrobaeza: No need of this `if`

pilarvargas-tecnativa: done

pedrobaeza (approved):
/ocabot merge nobump

PR #1549, opened 2026-02-25
18.0 mig hr announcement

Odoodev2:


hr_announcement/models/announcement.py
```diff
@@ -0,0 +1,123 @@
...
+                and ann
+                not in (
+                    x.user_id.read_announcement_ids + x.user_id.unread_announcement_ids
+                )
+            ):
+                employee.user_id.unread_announcement_ids |= announcement
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        announcement = super().create(vals_list)
```
alexey-pelykh, line 49: `vals_list` is a list (from `@api.model_create_multi`), not a dict — `list.get()` doesn't exist. This will raise `AttributeError`. Same pattern in `write()` works because `write()` receives a dict. Consider iterating: `any(v.get("employee_ids") for v in vals_list)`.

alexey-pelykh (changes requested):
Thanks for the migration! Found a bug in `create()`:

`vals_list.get("employee_ids")` will crash — with `@api.model_create_multi`, `vals_list` is a list of dicts, not a single dict. `list` doesn't have `.get()`. The `write()` method is fine since `write()` receives a dict.

Should be something like `any(v.get("employee_ids") for v in vals_list)`.

Also, the README badges still point to 17.0 for Weblate and Runboat — should be updated to 18.0.

pedrobaeza:
Superseded by #1582

PR #1562, opened 2026-04-18
[FIX] hr_employee_second_lastname: fix access error when reading employees public info

isaako34:
In v19, hr.employee.base was removed [1] and lastname2 was moved back to `hr.employee` but did not replicate it on `hr.employee.public`.

Since lastname2 is part of the employee display name and not sensitive data, declare it as a related field on hr.employee.public so non-HR users can read employees without triggering an AccessError.

[1]: https://github.com/OCA/hr/commit/d2e5fe75

### DIFF
--- a/hr_employee_second_lastname/models/__init__.py
+++ b/hr_employee_second_lastname/models/__init__.py
@@ -1,1 +1,2 @@
 from . import hr_employee
+from . import hr_employee_public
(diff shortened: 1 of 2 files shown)


luisg123v (changes requested):
Regarding commit message:
- Title could focuse on the actual issue, which is an access error when reading employees public info
- The reference `[1]` (ceb9584e) is for the migration of `hr_employee_firstname`, not `hr_employee_second_lastname`.

luisg123v (approved):
LGTM 👍

luistorresm (approved):

PR #1566, opened 2026-04-27
[FIX] hr_employee_second_lastname: expose lastname2 on public

luistorresm:
The field lastname2 was not available on hr.employee.public, causing the error 'fields are not available for employee public profiles' when non-HR users accessed employee records.

This follows the same approach used in hr_employee_firstname commit 51c92d0 which exposed firstname and lastname on the public model but missed lastname2 from this module.


luisg123v:
@luistorresm duplicate of https://github.com/OCA/hr/pull/1562

PR #1553, opened 2026-03-23
[FIX] hr_employee_second_lastname: replace mapped address_home_id

xavierbol:
Replace address_home_id mapped by work_contact_id. The relational field no longer exists on the address in V17 and replace by non relational field(Char) private_street.

Forward port fix (#1370) made in version 17

close #1551

PR #1571, opened 2026-05-12
[18.0][FIX] hr_employee_second_lastname: No more private partner

ypapouin:
Since Odoo 17.0, there's no more private `res.partner` records, everything is stored in private_* fields


Saran440:
@ypapouin Thanks for your contribute, Closed because I think this PR similar https://github.com/OCA/hr/pull/1553
please review instead.

PR #1583, opened 2026-06-02
[18.0][IMP] hr_appraisal_oca: add configurable email notifications

ebauza:
## Description

This PR adds a configurable option to disable email notifications
in the appraisal workflow.

A new boolean field **"Send email notifications"** allows managers
to control whether automatic emails are sent during appraisal
confirmation and completion.

## Functional behavior

- By default, email notifications are enabled.
- When disabled, no emails are sent during confirmation or completion.
- All other workflow behavior remains unchanged (state transitions,
  activities, etc.).
- This configuration is only visible to managers.

## Technical details

- Added field `send_email_notifications` on `hr.appraisal`
- Introduced helper method `_can_send_appraisal_email`
- Updated `_send_email` to respect the configuration and context override

## Testing

- Added tests to validate:
  - notifications enabled → emails are sent
  - notifications disabled → no emails are sent
  - context override behavior
- Existing tests remain unchanged and passing

## Rationale

This improvement provides flexibility to avoid sending automatic
emails in specific appraisal cases without altering the standard workflow.


hr_appraisal_oca/models/hr_appraisal.py
```diff
@@ -142,6 +147,14 @@ def default_get(self, fields_list):
                 )
         return res
 
+    def _can_send_appraisal_email(self):
+        self.ensure_one()
+        # Permite sobreescribir por contexto en acciones puntuales:
```
pedrobaeza, line 152: No comments in Spanish, please.

pedrobaeza (commented):
Thanks for the new feature for disabling email notifications. However, this is not using the `mail.thread` native possibilities, using followers, auto-suscriptions and mail subtypes for being able to have granularity over which ones want to receive the mails. It would be good to standardize the mails mechanism to the core features for having less code, hidden options, and have more granularity.

hr_appraisal_oca/models/hr_appraisal.py
ebauza: Thanks for the review.

I need to keep the per-appraisal setting for enabling/disabling notifications, as this is a business requirement.

After reviewing the mail framework in the core, I will refactor the implementation to rely on the standard "mail.thread" mechanism using "message_post_with_source" with mail templates.

For cases where notifications are disabled, the system will still log the event in the chatter using "message_post", without triggering email delivery.

This allows preserving the existing template-based notifications while aligning the implementation with the standard Odoo messaging framework.

I will also update the remaining Spanish comments to English.

pedrobaeza: That's the good thing, with the standard system, you can decide the granularity, and let the decision to other level. For example, at employee or department level. Check how it's done in the core the sales team subscriptions, and how it affects on sales orders and invoices according the subscribed subtypes.

ebauza: Thanks for the clarification, that makes sense.

Given that the current requirement involves per-appraisal control of notifications, we need to review this with the business team to see how it can align with the standard notification approach (followers / subtypes).

We’ll discuss this internally and revisit the implementation once we have more clarity.

For now, I’ll leave this on standby.

ebauza:
After further discussion and clarification of the requirements, we realized that the original proposal (controlling email notifications via a boolean) does not address the actual need.

The real requirement is to manage appraisal visibility (including confidential/internal appraisals), which differs significantly from the initial approach.

Based on reviewer feedback, we will close this PR and open a new one with a redesigned solution focused on visibility and access control.

Thanks a lot for the feedback and guidance!

PR #1601, opened 2026-08-03
[18.0][IMP] hr_employee_calendar_planning: Add flexible_hours compatibility

victoralmau:
Add flexible_hours compatibility

@Tecnativa TT62028


victoralmau:
@pedrobaeza What do you think of this approach?

This approach would require several additional modules to ensure compatibility with the `hr_attendance` and `hr_holidays` methods.

victoralmau:
It's ready for review. It has been tested with a client. Ping @pedrobaeza and @carlos-lopez-tecnativa

hr_employee_calendar_planning/models/hr_employee.py
```diff
@@ -48,6 +48,16 @@ class HrEmployee(models.Model):
         copy=True,
     )
 
+    def _get_available_calendards(self, from_date, to_date):
```
pedrobaeza: This method has a typo (extra `d`), but I think it should have a more specific name:


```suggestion
    def _get_planning_calendars(self, from_date, to_date):
```

hr_employee_calendar_planning/migrations/18.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,37 @@
...
+        """,
+    )
+    openupgrade.logged_query(
+        env.cr,
+        """
+        ALTER TABLE resource_calendar
+        ADD COLUMN IF NOT EXISTS stored_hours_per_day double precision
+        """,
+    )
+    openupgrade.logged_query(
```
pedrobaeza: This operation should be done at pre-install hook for this module to work on fresh installs, and call the hook from this migration script.

pedrobaeza (approved):

victoralmau:
Ping @carlos-lopez-tecnativa

hr_employee_calendar_planning/tests/test_hr_employee_calendar_planning.py
```diff
@@ -348,6 +353,90 @@ def test_calendar_planning_two_weeks_multi(self):
...
+            {
+                "stored_flexible_hours": True,
+                "stored_full_time_required_hours": 20,
+                "stored_hours_per_day": 4,
+            }
+        )
+        self.employee.calendar_ids = [
+            (0, 0, {"date_end": "2019-12-31", "calendar_id": self.calendar1.id}),
+            (0, 0, {"date_start": "2020-01-01", "calendar_id": self.calendar2.id}),
+        ]
```
carlos-lopez-tecnativa, line 379:
```suggestion
        self.employee.calendar_ids = [
            Command.create({"date_end": "2019-12-31", "calendar_id": self.calendar1.id}),
            Command.create({"date_start": "2020-01-01", "calendar_id": self.calendar2.id}),
        ]
```

carlos-lopez-tecnativa (changes requested):
Sorry if I don't understand the changes or if I'm missing some configuration, but I tested this locally by creating an employee, assigning this calendar, and installing the hr_attendance_report_theoretical_time module.

When reviewing the attendance report, the worked hours are not computed correctly. I have attached the images for reference.

Image

Image

Image

victoralmau:
> Sorry if I don't understand the changes or if I'm missing some configuration, but I tested this locally by creating an employee, assigning this calendar, and installing the hr_attendance_report_theoretical_time module.
> 

The screenshot of the theoretical hours report shows that it is working correctly; that is, the 25th has 4 theoretical hours (as you configured in the calendar), but the 24th does not.

The reason for this is how `hr_attendance_report_theoretical_time` works; most likely, the following happened:
- An attendance record already existed for the 24th > It had 0 hours defined as theoretical
- You changed the calendar to a flexible schedule with 4 hours
- You created an attendance record > At that moment, it calculated 4 hours of theoretical time

To have the “previous” data (attendance records prior to the current day) recalculated, you must use the corresponding wizard for `hr_attendance_report_theoretical_time`.

carlos-lopez-tecnativa:
I simplified the test by installing only this module and `hr_attendance`, without `hr_attendance_report_theoretical_time`.

I created a calendar with flexible hours, 20 hours per week, and 4 hours per day.
I have an attendance from the 6th to the 12th, but the worked hours are 2, and the extra worked hours are -0:43. Is this correct?
Without this module, the worked hours should be approximately 6.
I have attached the screenshots for reference.

image

image

image

victoralmau:
> I simplified the test by installing only this module and `hr_attendance`, without `hr_attendance_report_theoretical_time`.
> 
> I created a calendar with flexible hours, 20 hours per week, and 4 hours per day. I have an attendance from the 6th to the 12th, but the worked hours are 2, and the extra worked hours are -0:43. Is this correct? Without this module, the worked hours should be approximately 6. I have attached the screenshots for reference.
> image image image

Now I'll review the information regarding `hr_attendance` and extra hours.

Another example where you can see that this pull request fix an existing incompatibility issue:

Install `hr_employee_calendar_planning` + `hr_holidays`

**Before**
antes

**After**
despues

You can see that **BEFORE**, in a flexible calendar, the displayed calendar still used the old calendar (Monday to Friday), and now it doesn't.

victoralmau:
I see that an extension for `hr_attendance` is needed, so I'm going to create (in another PR in OCA/hr-attendance) a new module (`hr_attendance_employee_calendar_planning`) to address this specific case.

carlos-lopez-tecnativa (approved):
LGTM.

The issue related to attendance is solved with https://github.com/OCA/hr-attendance/pull/300. Thanks for the improvement!

PR #1418, opened 2024-12-13
[18.0][MIG] hr_employee_document: Migration to 18.0 

BhaveshHeliconia:

PR #1607, opened 2026-08-24
[18.0][FIX] hr_employee_document: Ensure the module works correctly for users who do not have Employee permissions

victoralmau:
Ensure the module works correctly for users who do not have Employee permissions

Related to https://github.com/OCA/hr/pull/1418

@Tecnativa TT63990

### DIFF
--- a/hr_employee_document/models/hr_employee.py
+++ b/hr_employee_document/models/hr_employee.py
@@ -2,7 +2,7 @@
 # Copyright 2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import api, fields, models
+from odoo import fields, models
 
 
 class HrEmployeeBase(models.AbstractModel):
@@ -13,7 +13,6 @@ class HrEmployeeBase(models.AbstractModel):
     )
 
     def _compute_document_count(self):
-        self.document_count = 0
         results = self.env["ir.attachment"]._read_group(
             [("res_model", "=", "hr.employee"), ("res_id", "in", self.ids)],
             groupby=["res_id"],
@@ -23,12 +22,11 @@ class HrEmployeeBase(models.AbstractModel):
         for record in self:
             record.document_count = count_dict.get(record.id, 0)
 
-    @api.model
-    def check_access(self, operation):
+    def _check_access(self, operation):
         """Return access to the hr.employee model if we pass a specific context,
         is a trick to list the attachments related to an employee."""
         if (
-            not self.env.is_superuser()
+            not self.env.su
             and not self.env.user.has_group("hr.group_hr_user")
             and operation == "read"
             and self._name == "hr.employee"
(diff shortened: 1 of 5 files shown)


hr_employee_document/models/hr_employee_public.py
```diff
@@ -8,6 +8,7 @@ class HrEmployeePublic(models.Model):
     _inherit = "hr.employee.public"
 
     is_logged = fields.Boolean(compute="_compute_is_logged", store=False)
+    document_count = fields.Integer(compute_sudo=True)
```
victoralmau:
```suggestion
```

hr_employee_document/models/hr_employee.py
```diff
@@ -23,8 +22,7 @@ def _compute_document_count(self):
         for record in self:
             record.document_count = count_dict.get(record.id, 0)
 
-    @api.model
-    def check_access(self, operation):
+    def _check_access(self, operation):
```
carlos-lopez-tecnativa, line 25: To avoid inverting the logic, you can use the `has_access` function.

carlos-lopez-tecnativa (approved):
LGTM, thanks for the fix. Indeed, this does not work as expected in V18 without this commit.

hr_employee_document/models/hr_employee.py
victoralmau: It is not possible to use the `has_access()` method you suggest because otherwise the module would not work correctly; it is important to use the `_check_access()` method because it is used indirectly in many places.

hr_employee_document/models/ir_attachment.py
```diff
@@ -0,0 +1,23 @@
...
+        """
+        res = super().validate_access(access_token)
+        record_sudo = self.sudo()
+        if (
+            res
+            and record_sudo.res_model == "hr.employee"
+            and record_sudo.res_id
+            and record_sudo.res_id in self.env.user.employee_ids.ids
+        ):
+            res = record_sudo
```
pedrobaeza: Shouldn't you sudoed `res`, not `self`?

```suggestion
        record_sudo = res.sudo()
        if (
            res
            and record_sudo.res_model == "hr.employee"
            and record_sudo.res_id
            and record_sudo.res_id in self.env.user.employee_ids.ids
        ):
            res = record_sudo
```

victoralmau: No, the record will be “res” without “sudo,” and that's incorrect.: https://github.com/odoo/odoo/blob/102924c49947c396229caf0ec77dc58e0b662bd1/odoo/addons/base/models/ir_attachment.py#L743

pedrobaeza: But I mean the mix between `res` and `self`, which may not be the same. Then I talk to do `res.sudo()` in that case.

victoralmau: `res` (https://github.com/odoo/odoo/blob/102924c49947c396229caf0ec77dc58e0b662bd1/odoo/addons/base/models/ir_attachment.py#L743) will be `self`, which is why I use `self. sudo()` (just like in https://github.com/odoo/odoo/blob/102924c49947c396229caf0ec77dc58e0b662bd1/odoo/addons/base/models/ir_attachment.py#L725) but if that’s a problem, I’ll use `res.sudo()` even though it’s exactly the same.

pedrobaeza: But maybe someone in other module override `validate_access` and return a different record by any kind of needed thing, so you can't assume that `self` will be always the same as `res`. You should always follow the inheritance chain. That's why returned values are returned... Not doing that is when bogus things happen.

victoralmau: I don't agree with that reasoning (if we're validating a record x, we should validate that record; we shouldn't return y, and doing that—regardless of which module does it—would be completely wrong). 
We want to validate a specific record, and that's exactly what we're doing in this method, but I don't want to argue about it anymore—I'll change it.

pedrobaeza: We have already done overrides where in a `create` call, we return another existing record instead. Don't focus in this specific case, where you won't probably have the problem because as you say, you won't probably return other thing, but focus on the **good practice**, that you have to manipulate variables according inheritance, and if the record is returned through the super, you manipulate that returned record, not the original `self` one. Doing this always, you will save your future you from hours of debugging on weird errors.

pedrobaeza (approved):
/ocabot merge patch

PR #1618, opened 2026-09-02
[19.0][IMP] hr_employee_calendar_planning: Add hours_per_week compatibility

victoralmau:
Change done:
- [x] Add `hours_per_week` compatibility
- [x] Define the appropriate condition in `_test_module_hr_attendance_employee_calendar_planning()`
- [x] FWP from 18.0: https://github.com/OCA/hr/pull/1621
- [x] Define the appropriate labels in the calendar form view
- [x] Define the correct behavior of `_compute_work_time_rate()`

@Tecnativa

### DIFF
--- a/hr_employee_calendar_planning/models/resource_calendar.py
+++ b/hr_employee_calendar_planning/models/resource_calendar.py
@@ -25,6 +25,7 @@ class ResourceCalendar(models.Model):
     # - stored_full_time_required_hours: This field will behave the same as the
     # full_time_required_hours field
     # - stored_hours_per_day: This will behave the same as the hours_per_day field
+    # - stored_hours_per_week: This will behave the same as the hours_per_week field
     # The flexible_hour and full_time_required_hours fields are converted to compute
     # fields with store=False to ensure the correct stored_* field is used, except
     # when the calendar is auto_generated and dates are passed via context; in that
@@ -48,11 +49,17 @@ class ResourceCalendar(models.Model):
         readonly=False,
         help="Average hours per day a resource is supposed to work with this calendar.",
     )
+    stored_hours_per_week = fields.Float(
+        compute="_compute_stored_hours_per_week",
+        store=True,
+        readonly=False,
+    )
     # We set the field to store=False; we want it to be computed whenever that data
     # needs to be accessed.
     flexible_hours = fields.Boolean(store=False)
     full_time_required_hours = fields.Float(store=False)
     hours_per_day = fields.Float(store=False)
+    hours_per_week = fields.Float(store=False)
 
     @api.depends("schedule_type")
     def _compute_stored_flexible_hours(self):
@@ -152,6 +159,22 @@ class ResourceCalendar(models.Model):
             )
 
     @api.depends(
+        "attendance_ids",
+        "attendance_ids.hour_from",
+        "attendance_ids.hour_to",
+        "two_weeks_calendar",
+        "flexible_hours",
+    )
+    def _compute_stored_hours_per_week(self):
+        """This method is the same as _compute_hours_per_week() in the resource
+        module.
+        """
+        for calendar in self.filtered(lambda c: not c.flexible_hours):
+            calendar.stored_hours_per_week = float_round(
+                calendar._get_hours_per_week(), precision_digits=2
+            )
+
+    @api.depends(
         "auto_generate",
         "flexible_hours",
         "employee_ids.calendar_ids",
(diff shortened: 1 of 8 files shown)


pedrobaeza (approved):

carlos-lopez-tecnativa (changes requested):
I'm not sure if the error is related to this change, but in the other Runboat instance, this is working. Here, however, I get the error, but it is hidden and I cannot expand it to see the details. You can check the error in the log, though. Please review.

http://oca-hr-19-0-pr1618-114795d1b022.runboat.odoo-community.org/odoo/employees

https://github.com/user-attachments/assets/0f8fe0c3-bec2-46e6-a5b8-009255d0d8b3

https://runboat.odoo-community.org/api/v1/builds/b2b9ec677-70eb-4c84-bd19-1361b2a5c2d7/log

victoralmau:
> I'm not sure if the error is related to this change, but in the other Runboat instance, this is working. Here, however, I get the error, but it is hidden and I cannot expand it to see the details. You can check the error in the log, though. Please review.
> 
> http://oca-hr-19-0-pr1618-114795d1b022.runboat.odoo-community.org/odoo/employees
> 2026-09-09.08-08-00.mp4
> 
> https://runboat.odoo-community.org/api/v1/builds/b2b9ec677-70eb-4c84-bd19-1361b2a5c2d7/log

Fixed.

hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -283,6 +308,10 @@ def _attendance_intervals_batch(
             if resources
             else None
         )
+        if resources:
+            resources.mapped("calendar_id").filtered(
+                lambda x: x.auto_generate
+            )._compute_flexible_hours()
```
carlos-lopez-tecnativa: I'm not sure I understand this part of the code. Why is `_compute_flexible_hours` called every time `_attendance_intervals_batch` is called? Could you explain this and add some comments, please?

I think this could be more expensive and reduce performance if `_attendance_intervals_batch` is called from multiple places across different modules.

carlos-lopez-tecnativa (changes requested):
I can't test these changes because the archive calendar is not created, as I mentioned in https://github.com/OCA/hr/pull/1620#pullrequestreview-5156789477.

hr_employee_calendar_planning/models/resource_calendar.py
victoralmau: Actually, this isn't necessary anymore. Could you check it again?

hr_employee_calendar_planning/models/resource_calendar.py
```diff
@@ -185,6 +191,35 @@ def _compute_hours_per_day(self):
...
+                to_date = self.env.context.get("flexible_hours_to_date")
+                employee = item.employee_ids[:1]
+                if (from_date or to_date) and employee:
+                    calendars = employee._get_planning_calendars(from_date, to_date)
+                    hours = (
+                        (sum(c.stored_hours_per_week for c in calendars.calendar_id))
+                        if calendars
+                        else False
+                    )
+            item.hours_per_week = hours
```
carlos-lopez-tecnativa, line 263: Is it correct to take `stored_hours_per_week` out of the if?

When you create a new calendar, `hours_per_week` and `stored_hours_per_week` are set to zero. If the user doesn't take this into account and doesn't change the value, the calendar is created with 8 hours per day **but 0 hours per week**. If I try to create a leave, the number of days or hours is zero, which is incorrect from my point of view.

Please review this and clarify whether this is the expected behavior.

Image

Image

Before this change, or in Odoo core, the calendar was created with 8 hours per day and 40 hours per week, with either flexible or fully fixed time.

Image

carlos-lopez-tecnativa (changes requested):

hr_employee_calendar_planning/models/resource_calendar.py
victoralmau: Yes, `stored_hours_per_week` will have the correct value.

Can you check it again with the changes you made?

hr_employee_calendar_planning/views/resource_calendar_views.xml
```diff
@@ -89,6 +89,62 @@
...
+            </xpath>
+            <xpath expr="(//field[@name='hours_per_week'])[4]" position="after">
+                <field
+                    name="stored_hours_per_week"
+                    nolabel="1"
+                    style="width: 6ch;"
+                    widget="float_time"
+                    readonly="1"
+                />
+            </xpath>
```
carlos-lopez-tecnativa, line 201: I noticed a visual issue with these fields.

The original view added a label with `label for="hours_per_week"`. You hide the field, but the label is still visible when you hover over it. This displays the field name for an invisible field, while the editable field is actually `stored_hours_per_week`. If I trust the field name displayed in the label, I could assume that the field I'm editing is `hours_per_week` instead of `stored_hours_per_week`.

Image

Additionally, `Work Time Rate` is not recomputed when I change stored_hours_per_week. I noticed that it is only updated when the record is saved. This is non-blocking, but it caught my attention.

https://github.com/user-attachments/assets/ff03aa7d-4b37-4e6a-9c87-fa930ea46f98

carlos-lopez-tecnativa (approved):
LGTM, just a non-blocking comment, but if you can address this, that would be great.

carlos-lopez-tecnativa (approved):
LGTM

PR #1619, opened 2026-09-07
[19.0][IMP] hr_employee_calendar_planning: Fix test compatibility

victoralmau:
FWP from 18.0: https://github.com/OCA/hr/pull/1611

Fix test compatibility

Install hr_employee_calendar_planning + test hr_holidays_leave_repeated

When hr_employee_calendar_planning is installed and tests for other modules that use resource_calendar_id are run, the tests will fail

@Tecnativa

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -2,10 +2,9 @@
 # Copyright 2022-2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
We have discovered a new way to detect the current module being tested that can be more convenient, as you don't depend on the context propagation:

https://github.com/OCA/server-auth/pull/999/changes#diff-6717eb9e1d50756860fe6d8e0f068c6a75dbedb4d75e6d468e9dd44b2a4d7c6aR74

You can consider changing it.

carlos-lopez-tecnativa (approved):

PR #1620, opened 2026-09-08
[19.0][IMP] hr_employee_calendar_planning: Test compatibility with dependent modules

victoralmau:
Test compatibility with dependent modules

Related to https://github.com/OCA/hr/pull/1619

It is important to include a method to verify whether the tests for module x are running, because there will be cases (such as `hr_attendance_employee_calendar_planning`) where the tests for that module —which depends on this one (`hr_employee_calendar_planning`)— also need to satisfy the condition.

**Note**: Something similar will happen with the `hr_holidays_calendar_planning` module

@Tecnativa

### DIFF
--- a/hr_employee_calendar_planning/models/hr_employee.py
+++ b/hr_employee_calendar_planning/models/hr_employee.py
@@ -228,14 +228,31 @@ class HrEmployee(models.Model):
         new.filtered("calendar_ids").regenerate_calendar()
         return new
 
+    def _test_module_hr_attendance_employee_calendar_planning(self):
+        """This method indicates whether the module currently being executed is the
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):
/ocabot merge patch

carlos-lopez-tecnativa (commented):
Now, when you create a new employee, the archive calendar with `Auto generated calendar for employee` is not created. Please review this.

victoralmau:
> Now, when you create a new employee, the archive calendar with `Auto generated calendar for employee` is not created. Please review this.

Fixed in an additional commit in https://github.com/OCA/hr/pull/1618 Thanks!

